Issue #52, opened 2025-07-06
Add microbenchmarks for sub-component performance evaluation


liran-funaro:
To optimize the system's overall latency and throughput, it's useful to evaluate the performance of each sub-component in isolation.
Introducing microbenchmarks will enable us to measure and analyze the efficiency of individual components, identify potential bottlenecks, and guide targeted optimizations.
It will also allow testing with a profiler for fine-grain optimizations.

This will provide greater visibility into system behavior and help ensure that performance regressions are detected early during development.

- [x] loadgen - generation
- [x] dependency graph - processing
- [x] preparer - processing
- [x] sidecar - block mapping
- [x] verifier - signature and digest

PR #64, opened 2025-07-06
[BREAKING] Fix TX ASN1 digest

### LINKED ISSUE #53: [tx digest] DigestTxNamespace does not encode the namespace ID and version
- DigestTxNamespace does not encode the namespace ID and version
- DigestTxNamespace has poor performance. Using reflection reduce performance dramatically and introduces redundant memory allocations.

liran-funaro:
#### Type of change

- Bug fix
- Test update
- Breaking change
 
#### Description

- Add namespace ID and version to the ASN.1 schema
- Remove redundant marshaling when using NONE schema
  - Allows better benchmarking of non-sig related components
- Add a test that validates the marshalling schema includes all the fields of the TX by reconstructing the TX from the marshalled data
- Add fuzz test for the ASN.1 marshalling
- Add a benchmark for the digest method

The tests discovered the following:
1. The ASN.1 library may encode a string in two ways, depending on the input string.
If it is ASCII, it encodes it as PrintableString; otherwise, it uses UTF-8.
This may cause incompatibility with the schema, which requires a specific type. Thus, we force the type to be UTF-8.
2. The ASN.1 marshalling cannot differentiate between `nil` and empty slice (`[]byte{}`).
We must ensure our implementation does not make this differentiation to avoid inconsistencies.
The current implementation parses version `[]byte{}` as zero. If the endorser signed on `nil` version, an attacker can change the TX to have version zero by replacing the `nil` with `[]byte{}` and go unnoticed.

To fix these issues, we:
- Modify the ASN.1 schema to use UTF-8 (UTF8String) instead of ASCII (PrintableString)
- Change version to be uint64. 
   - Since ASN.1 doesn't support unsigned int, we use int64 and encode nil as `-1`.
   - Since SQL doesn't support unsigned number, we use bigint (int64) and add check for NULL and version>=0.

#### Related issues

- resolves #53  
- address #52


service/vc/preparer.go
```diff
@@ -263,22 +268,18 @@ func (p *preparedTransactions) addReadWrites(id TxID, ns *protoblocktx.TxNamespa
...
+		p.readToTxIDs[cr] = append(rtt, id)
 
 		if rw.Version != nil {
 			if !present {
 				nsReads.append(rw.Key, rw.Version)
 			}
-			ver := types.VersionNumberFromBytes(rw.Version) + 1
-			nsWrites.append(rw.Key, rw.Value, ver.Bytes())
+			ver := *rw.Version + 1
+			nsWrites.append(rw.Key, rw.Value, ver)
```
cendhu, line 280: nit. 
```suggestion
			nsWrites.append(rw.Key, rw.Value, *rw.Version + 1)
```

service/vc/test_exports.go
```diff
@@ -276,11 +276,24 @@ func (env *DatabaseTestEnv) populateData( //nolint:revive
...
-			INSERT INTO %s (key, value, version)
+		if writes.empty() {
+			continue
+		}
+
+		require.NotNil(t, writes.keys)
+		require.NotNil(t, writes.values)
+		require.NotNil(t, writes.versions)
+		require.Len(t, writes.values, len(writes.keys))
+		require.Len(t, writes.versions, len(writes.keys))
```
cendhu, line 287: I see this code at multiple places. What was the observation or need to introduce this check given that these helpers are used by the test code only? Was this introduced for debugging?

service/vc/validator_test.go
```diff
@@ -89,33 +85,33 @@ func TestValidate(t *testing.T) { //nolint:maintidx // cannot improve.
...
-			versions: [][]byte{v2},
+			versions: []uint64{2},
 		},
 	}
 	tx3BlindWrites := namespaceToWrites{
 		"1": {
 			keys:     [][]byte{k1_6},
 			values:   [][]byte{[]byte("value1.6")},
-			versions: [][]byte{nil},
+			versions: []uint64{0},
```
cendhu, line 114: For future reference, we can add a comment that the version would not be included in the transaction for the blind-write and the vcservice would fetch the version from disk irrespective of what we pass here. Anyway, for this test, it does not matter what the version is. I believe this will help us in the future.

utils/signature/block_tx.go
```diff
@@ -17,77 +17,113 @@ import (
...
-		Value []byte
+		Version int64 `asn1:"optional,default:-1"`
 	}
 	// ReadWrite is a stab for [protoblocktx.ReadWrite].
+	// Any change to this protobuf requires a change to these structures.
 	ReadWrite struct {
 		Key     []byte
-		Version []byte
 		Value   []byte
+		Version int64 `asn1:"optional,default:-1"`
```
cendhu, line 121: Why do we have to make it optional as we always use `ProtoToAsnVersion`?

cendhu (commented):
LGTM. A few minor comments.

api/types/types.go
```diff
@@ -6,17 +6,6 @@ SPDX-License-Identifier: Apache-2.0
...
-)
-
-type (
-	// VersionNumber represents a row's version.
-	VersionNumber uint64
-)
-
 const (
 	// MetaNamespaceID is the system's namespace ID that holds information about application's namespaces.
 	MetaNamespaceID = "_meta"
```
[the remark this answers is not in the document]
liran-funaro: Yes. This will be done in steps.

service/vc/committer.go
```diff
@@ -219,10 +219,10 @@ func (c *transactionCommitter) populateVersionsAndCategorizeBlindWrites(
 			nsState := state[ns]
 			for i, key := range nsWrites.keys {
 				if ver, present := nsState[string(key)]; present {
-					nextVer := (types.VersionNumberFromBytes(ver) + 1).Bytes()
+					nextVer := ver + 1
 					vTx.validTxNonBlindWrites.getOrCreate(curTxID, ns).append(key, nsWrites.values[i], nextVer)
 				} else {
-					vTx.newWrites.getOrCreate(curTxID, ns).append(key, nsWrites.values[i], nil)
+					vTx.newWrites.getOrCreate(curTxID, ns).append(key, nsWrites.values[i], 0)
```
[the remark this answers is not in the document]
liran-funaro: Yes. We ignore this value for new-writes since it is assigned using the DB default value, but we must append some value.
Having this be nullable (`*uint64`) is not useful, as writes must have a version.
It makes sense to write `0` as it shows the true intent.

service/vc/create_namespace.sql
```diff
@@ -4,14 +4,24 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-CREATE TABLE IF NOT EXISTS ns_table
+/*
+This SQL file is a template for creating a new namespace.
```
cendhu, line 8: add `tmpl` suffix to the file name `create_namespace_tmpl.sql`.

cendhu (approved):
LGTM.

PR #67, opened 2025-07-06
[verifier] Benchmark verifyTxForm

liran-funaro:
#### Type of change

- Test update
 
#### Description

- Add a test and benchmark for `verifyTxForm()`

#### Related issues

- address #52


service/verifier/verify_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger/fabric-x-committer/api/protoblocktx"
+	"github.com/hyperledger/fabric-x-committer/loadgen/workload"
+	"github.com/hyperledger/fabric-x-committer/utils/logging"
+)
+
+func TestVerifyBadTxForm(t *testing.T) {
```
cendhu, line 19: There is already a test that checks badly formed txs -- [TestBadTxFormat](https://github.com/hyperledger/fabric-x-committer/blob/fd133090bdb4365cd1de4ff97ae33cc04e128aec/service/verifier/verifier_server_test.go#L124)

service/verifier/verify_test.go
```diff
@@ -0,0 +1,47 @@
...
+		t.Run(tt.Tx.Id, func(t *testing.T) {
+			t.Parallel()
+			actualStatus := verifyTxForm(tx)
+			require.Equal(t, expectedStatus, actualStatus)
+		})
+	}
+}
+
+func BenchmarkVerifyForm(b *testing.B) {
+	logging.SetupWithConfig(&logging.Config{Enabled: false})
```
cendhu, line 37: Is this not disabled by default?

service/verifier/verify_test.go
```diff
@@ -0,0 +1,47 @@
...
+	txs := workload.GenerateTransactions(b, workload.DefaultProfile(8), b.N)
+
+	var status protoblocktx.Status
+	b.ResetTimer()
+	for _, tx := range txs {
+		status = verifyTxForm(tx)
```
cendhu, line 43: If this is for benchmark, we can ignore the return value.

cendhu (commented):
LGTM. A few minor comments.

service/verifier/verify_test.go
liran-funaro: This test only verifies the verify method directly. Without streams and all that.

cendhu: Okay. Looks redundant. Anyway, when you move this to sidecar, the redundancy would be removed.

cendhu (approved):

service/verifier/verify_test.go
liran-funaro: Yes. It would be converted to a block mapping test, which is missing.

PR #87, opened 2025-07-17
[coordinator] restrict batch size

### LINKED ISSUE #77: [coordinator] Restrict the Batch Size
When all signature verifiers or VC services fail, the queue in the signature verifier manager or validator-committer manager builds up. Although the orderer manages size restrictions when creating a block, the problem arises when a service comes back online. The coordinator can club multiple transactions together into a single batch, whose combined size can exceed the gRPC send limit of around 100 MB. As a result, this causes progress to stall with a gRPC send error. Therefore, we must restrict the batch size, as this issue can also occur when the application stores a large amount of data in the write-set.

cendhu:
#### Type of change

- Bug fix
 
#### Description

This commit restrict the batch size sent to verifier and vcservice when it receives resource exhausted status code on grpc.

#### Related issues

  - Resolves #77


utils/connection/client_util.go
```diff
@@ -34,8 +34,10 @@ const (
 
 	// defaultGrpcMaxAttempts is set to a high number to allow the timeout to dictate the retry end condition.
 	defaultGrpcMaxAttempts = 1024
-	maxMsgSize             = 100 * 1024 * 1024
-	scResolverSchema       = "sc.connection"
+	// TODO: All services including the orderer must use the same default maximum message size.
+	//       Hence, we need to move this constant to fabrix-x-common.
```
liran-funaro, line 38: nit
```suggestion
	//       Hence, we need to move this constant to fabric-x-common.
```

utils/connection/server_util.go
```diff
@@ -67,6 +67,8 @@ func (c *ServerConfig) GrpcServer() *grpc.Server {
 			PermitWithoutStream: c.KeepAlive.EnforcementPolicy.PermitWithoutStream,
 		}))
 	}
+	opts = append(opts, grpc.MaxRecvMsgSize(maxMsgSize))
+	opts = append(opts, grpc.MaxSendMsgSize(maxMsgSize))
```
liran-funaro, line 71: nit: these can be added at the beginning of the function.

service/coordinator/validator_committer_manager.go
```diff
@@ -260,17 +261,54 @@ func (vc *validatorCommitter) sendTransactionsToVCService(
...
+			}
+			blkToBatch[tx.BlockNumber] = rBatch
+		}
+
+		rBatch.Transactions = append(rBatch.Transactions, tx)
+	}
+
+	for _, rBatch := range blkToBatch {
+		if err := stream.Send(rBatch); err != nil {
+			return err
```
liran-funaro, line 305: minor: We should wrap the error here, not at the caller.

service/coordinator/signature_verifier_manager.go
```diff
@@ -231,13 +232,72 @@ func (sv *signatureVerifier) sendTransactionsToSVService(
...
+	for _, rBatch := range blkToBatch {
+		if !updateSent {
+			rBatch.Update = r.Update
+			updateSent = true
+		}
+
+		if err := stream.Send(rBatch); err != nil {
+			// ResourceExhausted should not occur here, as we have split a block's transactions
+			// into two batches, assuming the block size is less than the maximum gRPC message size.
+			return err
```
liran-funaro, line 294: minor: We should wrap the error here, not at the caller.

liran-funaro (approved):

utils/connection/retry.go
```diff
@@ -130,7 +129,6 @@ func (p *RetryProfile) MakeGrpcRetryPolicyJSON() string {
 				"retryableStatusCodes": []string{
 					"UNAVAILABLE",
 					"DEADLINE_EXCEEDED",
-					"RESOURCE_EXHAUSTED",
```
liran-funaro, line 133: Can `RESOURCE_EXHAUSTED` be a temporary problem? If so, shouldn't we retry.
I understand that in the case this PR is solving, the error is permanent. But I'm not sure it is always the case for this error.

utils/connection/server_util.go
```diff
@@ -42,7 +42,8 @@ func NewLocalHostServer() *ServerConfig {
 
 // GrpcServer instantiate a [grpc.Server].
 func (c *ServerConfig) GrpcServer() *grpc.Server {
-	var opts []grpc.ServerOption
+	opts := append([]grpc.ServerOption{}, grpc.MaxRecvMsgSize(maxMsgSize))
+	opts = append(opts, grpc.MaxSendMsgSize(maxMsgSize))
```
liran-funaro, line 46: nit
```suggestion
	opts := []grpc.ServerOption{grpc.MaxRecvMsgSize(maxMsgSize), grpc.MaxSendMsgSize(maxMsgSize)}
```

liran-funaro (approved):
Please see comments

utils/connection/retry.go
cendhu: I will revert this change for now. We can investigate more and make changes to this code later.

Issue #111, opened 2025-08-13
[proto] Unnecessery protobuf messages


liran-funaro:
Some of the protobuf messages can be removed to reduce the number of protobuf messages.
This can improve the API's readability.

For example:

#### VC
```proto
message Tx {
  protocoordinatorservice.TxRef ref = 1;
  repeated protoblocktx.TxNamespace namespaces = 2;
  optional InvalidTxStatus prelim_invalid_tx_status = 3;
}

message InvalidTxStatus {
    protoblocktx.Status code = 1;
}
```

Can be reduced to:
```proto
message Tx {
  protocoordinatorservice.TxRef ref = 1;
  repeated protoblocktx.TxNamespace namespaces = 2;
  optional protoblocktx.Status prelim_invalid_tx_status = 3;
}
```

#### Notify
```proto
message NotificationResponse {
    repeated TxStatusEvent tx_status_events = 1;
    repeated string timeout_tx_ids = 2;
}

message TxStatusEvent {
    string tx_id = 1;
    protoblocktx.StatusWithHeight status_with_height = 2;
}
```

Can be reduced to:
```proto
message NotificationResponse {
    repeated protoblocktx.TransactionsStatus tx_status_events = 1;
    repeated string timeout_tx_ids = 2;
}
```

Moreover, `protoblocktx.TransactionsStatus` and `protoblocktx.StatusWithHeight` can be removed completely by using `protoblocktx.TxStatusInfo`.

liran-funaro:
The following:
```proto
// blocktx
message TransactionsStatus {
    map<string, StatusWithHeight> status = 1;
}

message StatusWithHeight {
    Status code = 1;
    uint64 block_number = 2;
    uint32 tx_number = 3;
}

// notify
message TxStatusEvent {
    string tx_id = 1;
    protoblocktx.StatusWithHeight status_with_height = 2;
}

// verifier
message Response {
    protocoordinatorservice.TxRef ref = 1;
    protoblocktx.Status status = 2;
}
```

Can be replaced with:
```proto
message TxStatusBatch {
    repeated TxStatus status = 1;
}

message TxStatus {
    protocoordinatorservice.TxRef ref = 1;
    protoblocktx.Status status = 2;
}
```

Issue #44, opened 2025-07-06
[API] separating external and internal API


liran-funaro:
Currently, all of our protobuf API are under `/api`. But most of the protos there are for internal use.

The external APIs are:
- TX (Block is internal)
- NamespacePolicy
- query service (+policies)
- delivery (from Fabric)

liran-funaro:
### Comments from original issue:

---

**Comment from @cendhu:**

I agree. It would be better to delay this change since we plan to move the common protos to a separate repository, similar to Fabric. Originally, this separation was planned, but to expedite development, we decided to keep them in the committer repository. Once we move the common protos, we can simply rename the `api/` directory to `protos/`. Otherwise, we have to make a two huge changes. 

---

**Comment from @cendhu:**

Further, if we make any changes in the package path, it would break the backward compatibility of other services such as endorserX and fcs. Hence, it would be better to involve all teams and create a repo for the common protos used across components. 

---

PR #132, opened 2025-09-10
[proposal] API split and deduplication

liran-funaro:
#### Type of change

- Breaking change
 
#### Description

- Conform to the Google protobuf naming conventions. E.g. `emptypb.Empy`, `durationpb.Duration`
- Split the API into two main components:
   1. `committerpb`: all the API that are used to access the committer publicly. It will eventually be moved to the common repository.
   2. `servicepb`: all the internal API
- Deduplicate messages

After this split, the `committerpb` package will be moved to `fabric-x-common`.

#### Alternatives
- `committerpb` can be split into `applicationpb` and `committerpb`, where the former will include the `block_tx.proto` and `config.proto`.
- Also move the load-gen to `fabric-x-common`, without the committer adapters. 

#### Related issues

- address #111 
- address #44


cendhu:
@liran-funaro The proposed changes looks fine to me. The applicationpb and committerpb in the fabric-x-common make sense. Further, server.Coordinator reads well too.

Issue #124, opened 2025-09-02
[committer] containerized TLS test


dean-amar:
To verify TLS communication between the committer components, we need to create an integration test that runs each service in a separate Docker container and establishes connections between them.

Issue #123, opened 2025-09-02
[integration-test] add integration test for release images


dean-amar:
Currently, we have an integration test that uses only the test-node image. We need to add another integration test that runs with the release images.

PR #143, opened 2025-09-21
[docker-test] Add release-image–based containerized integration tests with TLS modes

dean-amar:
#### Type of change
- Test update
#### Description
 * Add release-image–based containerized integration tests with different TLS modes.
 * Add support for multi-SAN server TLS credential creation.
 * Align configuration names.
 * Move common Docker utils to utils/test.
#### Additional details
The integration test runs each service in a separate Docker container and establishes connections between them with different TLS modes.
For now, we load TLS certificates into the mock-orderer container, even though it doesn’t support them yet (a future PR will add support).
We also need to align the config/template names, which will be addressed in a separate PR.
#### Related issues
* solves: #123 
* solves: #124

### DIFF
--- a/cmd/config/app_config_test.go
+++ b/cmd/config/app_config_test.go
@@ -34,18 +34,18 @@ import (
 var (
 	defaultServerTLSConfig = connection.TLSConfig{
 		Mode:     connection.MutualTLSMode,
-		CertPath: "/server_certs/public-key",
-		KeyPath:  "/server_certs/private-key",
+		CertPath: "/server-certs/public-key",
+		KeyPath:  "/server-certs/private-key",
 		CACertPaths: []string{
-			"/server_certs/ca-certificate",
+			"/server-certs/ca-certificate",
 		},
 	}
 	defaultClientTLSConfig = connection.TLSConfig{
 		Mode:     connection.MutualTLSMode,
-		CertPath: "/client_certs/public-key",
-		KeyPath:  "/client_certs/private-key",
+		CertPath: "/client-certs/public-key",
+		KeyPath:  "/client-certs/private-key",
 		CACertPaths: []string{
-			"/client_certs/ca-certificate",
+			"/client-certs/ca-certificate",
 		},
 	}
 )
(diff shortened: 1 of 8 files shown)


cmd/config/app_config_test.go
```diff
@@ -458,6 +482,14 @@ func newMonitoringConfig(host string, port int) monitoring.Config {
 	}
 }
 
+//nolint:unparam // currently, we use this function only with host:""
+func newServerConfigWithDefaultTLS(host string, port int) *connection.ServerConfig {
```
liran-funaro: nit: If we always use it with host="", then please remove the parameter.
This linter is here for a reason :)

docker/test/common.go
```diff
@@ -0,0 +1,150 @@
...
+	err = dockerClient.ContainerRemove(ctx, id, container.RemoveOptions{
+		RemoveVolumes: true,
+		Force:         true,
+	})
+	if err != nil {
+		t.Logf("unable to remove container: %s", err)
+	}
+}
+
+func containerMappedHostPort(
```
liran-funaro: nit: The method name was confusing. I suggest the following:
```suggestion
func getContainerMappedHostPort(
```

cmd/config/app_config_test.go
dean-amar: True, fixed.

docker/test/container_release_image_test.go
```diff
@@ -0,0 +1,254 @@
...
+func assembleBinds(t *testing.T, serverCredsPath, clientCredsPath string, additionalBinds ...string) []string {
+	t.Helper()
+	return append([]string{
+		fmt.Sprintf("%s:/certs", serverCredsPath),
+		fmt.Sprintf("%s:/client_certs", clientCredsPath),
+	}, additionalBinds...)
+}
+
+// mustGetWD returns the current working directory.
+func mustGetWD(t *testing.T) string {
```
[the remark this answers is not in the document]
dean-amar: Reverted and explained in the comment above.

PR #155, opened 2025-10-22
[proto] Remove redundant RPC GetLastCommittedBlockNumber()

### LINKED ISSUE #48: [proto] Unused RPC in the Coordinator: GetLastCommittedBlockNumber
The method `rpc GetLastCommittedBlockNumber` in the coordinator is not used outside of testing.
It can be removed.

liran-funaro:
#### Type of change

#### Description

- Remove redundant RPC in the coordinator: `GetLastCommittedBlockNumber()` - It was already covered by `GetNextExpectedBlockNumber()`
- Remove unnecessary stream check in `GetNextExpectedBlockNumber()` - The sidecar verifies this anyway before calling this method
- Align API with VC - Replaced  `GetLastCommittedBlockNumber()`  with `GetNextExpectedBlockNumber()`
- Adjust relevant tests

#### Related issues

- resolves #48


cendhu (commented):
My suggestion is to use the name `GetLastCommittedBlockNumber` instead of `GetNextExpectedBlockNumber`. This also aligns with the existing `SetLastCommittedBlockNumber`. The caller (i.e., the sidecar) can then derive the next expected block number from the last committed block number.

PR #180, opened 2025-11-19
Adding support for production grade image in Dockerfile

anushka159-a:
I have modified the current Dockerfile to be able to generate a Production grade Docker Image based on UBI, the new Dockerfile follows the below good practices for Production.

Multi-stage build: keeps the final image small and clean.
Non-root execution: creates appuser (uid 10001) and runs the service as non-root.
Minimal runtime dependencies: no extra tools/packages installed (keeps attack surface small).
Metadata labels: provides traceability (name, version, maintainer, vendor, etc.).
Deterministic build: dependencies are pulled in the build stage, final image only has the binary.
Absolute path: in ENTRYPOINT: safer for production.

### DIFF
--- a/docker/images/release/Dockerfile
+++ b/docker/images/release/Dockerfile
@@ -14,7 +14,11 @@ ARG ARCHBIN_PATH
 ARG TARGETOS
 ARG TARGETARCH
 ARG PORTS
+
+# Label args
 ARG VERSION=1.0
+ARG CREATED
+ARG REVISION=1.0
 
 # Add non-root user (10001) using BIN argument
 RUN /usr/sbin/useradd -u 10001 -r -g root -s /sbin/nologin -c "Hyperledger Fabric-X ${BIN} user" ${BIN} && \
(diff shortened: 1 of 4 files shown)


pasquale95 (dismissed):
LGMT

pasquale95 (approved):
LGTM

docker/images/release/Dockerfile
```diff
@@ -1,13 +1,48 @@
...
+# Expose ports
 EXPOSE ${PORTS}
+
+# OCI metadata labels
+LABEL name="${BIN}" \
+    maintainer="IBM Research ZRL Decentralized Trust Group" \
+    version="${VERSION}" \
+    description="Hyperledger Fabric-X ${BIN} packaged in a UBI image" \
+    license="Apache-2.0" \
+    vendor="IBM"
```
cendhu: I am not sure about these metadata especially maintainer and vendor. What labels do you see for the fabric images? We should not move away from that.

PR #192, opened 2025-12-08
Move notify and query to committerpb

liran-funaro:
#### Type of change

- Breaking change
 
#### Description

- Move notify and query to committerpb

#### Additional details

This is part of a series of refactors

#### Related issues

- address #44


Makefile
```diff
@@ -56,7 +56,7 @@ go_build       ?= $(env) $(go_cmd) build $(build_flags)
 go_test        ?= $(go_cmd) test -json -v -timeout 30m
 proto_path     ?=
 
-ifneq ($(wildcard "/usr/include"),"")
+ifneq ("$(wildcard /usr/include)","")
```
cendhu: That's why `make proto` printed warning for non-existence of this directory.

cendhu (commented):
Needs to be rebased.

Makefile
liran-funaro: In Linux, it resides in `/usr/include`. In Mac, I don't know. But we don't have to manually set this in Mac.

cendhu (approved):
LGTM.

PR #203, opened 2025-12-18
Introduce support for MSP load generation

### LINKED ISSUE #188: [loadgen] Use MSP to sign transactions

liran-funaro:
#### Type of change

- New feature
 
#### Description

- Introduce support for MSP load generation

### Additional Info

Following is the micro benchmark for endorsing and verifying with the MSP using 3 peer organizations.

```
goos: linux
goarch: amd64
pkg: github.com/hyperledger/fabric-x-committer/utils/signature/sigtest
cpu: Intel(R) Core(TM) i9-14900K

BenchmarkDigest-32    			  466413	      2487 ns/op	402,091 TX/sec

BenchmarkSign/ECDSA-32         	   50209	     22191 ns/op	 45,063 TX/sec
BenchmarkSign/BLS-32           	   17622	     67974 ns/op	 14,712 TX/sec
BenchmarkSign/EDDSA-32         	   73783	     14747 ns/op	 67,810 TX/sec
BenchmarkSign/MSP-32           	   17944	     63762 ns/op	 15,683 TX/sec

BenchmarkVerify/ECDSA-32       	   25924	     44132 ns/op	 22,659 TX/sec
BenchmarkVerify/BLS-32         	    2367	    474782 ns/op	  2,106 TX/sec
BenchmarkVerify/EDDSA-32       	   40964	     28859 ns/op	 34,651 TX/sec
BenchmarkVerify/MSP-32         	    8385	    138661 ns/op	  7,212 TX/sec
```

#### Related issues

- resolves #188


cmd/config/templates/loadgen_common.yaml
```diff
@@ -38,15 +38,15 @@ load-profile:
...
-          scheme: {{ $element.Scheme }}
-          seed: {{ $element.Seed }}
-        {{ end }}
-      config-block-path: {{ .ConfigBlockPath }}
+  policy:
+    channel-id: {{ .Policy.ChannelID }}
+    namespace-policies:
+      {{ range $nsID, $element := .Policy.NamespacePolicies }}
+      {{ $nsID }}:
+        scheme: {{ $element.Scheme }}
```
cendhu: Please add a comment listing all schemes including the MSP scheme and the default policy for the MSP scheme.

loadgen/workload/config.go
```diff
@@ -113,17 +113,31 @@ type PolicyProfile struct {
 	// If this field is empty, a default config block will be generated.
 	ConfigBlockPath string `mapstructure:"config-block-path" yaml:"config-block-path"`
 
+	// CryptoMaterialPath may specify the path to the material generated by CreateDefaultConfigBlockWithCrypto().
+	// If this field is empty, the material will be generated into a temporary folder.
+	// If this path does not exist, or it is empty, the material will be generated into it.
+	// If ConfigBlockPath is empty, the config block will be fetched from this folder.
```
cendhu: This is confusing with the comment in line 113. It would be better to explain both ConfigBlockPath and CryptoMaterialPath in combination. What if one is empty or both empty or neither non-empty.

loadgen/workload/config.go
```diff
@@ -113,17 +113,31 @@ type PolicyProfile struct {
...
+	CryptoMaterialPath string `mapstructure:"crypto-material-path" yaml:"crypto-material-path"`
+
 	// ChannelID and Identity are used to create the TX envelop.
 	ChannelID string                      `mapstructure:"channel-id"`
 	Identity  *ordererconn.IdentityConfig `mapstructure:"identity"`
+
+	// PeerOrganizationCount may specify the number of peer organizations to generate if the CryptoMaterialPath
+	// is not provided.
+	PeerOrganizationCount uint32 `mapstructure:"peer-organization-count"`
```
cendhu: I just realized that `CreateConfigBlockWithCrypto` always uses TwoOrgsSampleFabricX profile. What does 4 peer organization count means in two org profile?

cmd/config/templates/loadgen_common.yaml
liran-funaro: I've added a comment to the sample file.

PR #206, opened 2025-12-25
Use common TX status

liran-funaro:
#### Type of change

- Breaking change
 
#### Description

- Use `committerpb.TxStatus` instead of `StatusWithHeight`

#### Related issues

- address #44


service/coordinator/validator_committer_manager.go
```diff
@@ -310,10 +310,12 @@ func (vc *validatorCommitter) receiveStatusAndForwardToOutput(
...
 
-		txsNode, untrackedTxIDs := vc.getTxsAndUpdatePolicies(txsStatus)
-		for _, txID := range untrackedTxIDs {
-			// untrackedTxIDs can be non-empty only when the coordinator restarts.
-			delete(txsStatus.Status, txID)
+		txsNode, untrackedTxIdx := vc.getTxsAndUpdatePolicies(txsStatus)
+		if len(untrackedTxIdx) > 0 {
+			// untrackedTxIdx can be non-empty only when the coordinator restarts.
+			for _, i := range slices.Backward(untrackedTxIdx) {
+				txsStatus.Status = append(txsStatus.Status[:i], txsStatus.Status[i+1:]...)
```
cendhu: Negligible performance impact is fine as this is a rare case.

service/sidecar/sidecar.go
```diff
@@ -406,19 +406,23 @@ func waitForIdleCoordinator(ctx context.Context, client servicepb.CoordinatorCli
...
-	expectedHeight map[string]*servicepb.Height,
+	statuses []*committerpb.TxStatus,
+	expectedHeight []*committerpb.TxRef,
 ) error {
-	for txID, height := range expectedHeight {
-		s, ok := statuses[txID]
+	statusMap := make(map[string]*committerpb.TxStatus, len(statuses))
+	for _, s := range statuses {
+		statusMap[s.Ref.TxId] = s
+	}
```
cendhu: This copy is fine as it is a rare-case. It is better to add a comment. Otherwise, in the future, someone will try to optimize this unnecessarily.

service/vc/committer.go
```diff
@@ -172,28 +171,24 @@ func (c *transactionCommitter) commitTransactions(
...
 
-	for _, lst := range []transactionToWrites{
-		vTx.validTxNonBlindWrites, vTx.validTxBlindWrites, vTx.newWrites,
-	} {
-		for txID := range lst {
-			setStatus(txID, committerpb.Status_COMMITTED)
-		}
+	statuses := make([]*committerpb.TxStatus, 0, len(txIDs))
+	for txID, status := range txIDs {
+		statuses = append(statuses, vTx.txIDToHeight[txID].WithStatus(string(txID), status))
```
cendhu: This is in a hot path. I don't know its real impact as it could leads to a lot of allocation and garbage collection. The reason for using the map in TransactionStatus was to enable quicker lookup now and in the future (for new features). When such a need arises, we can revert this to map.

PR #215, opened 2026-01-06
Use a single mock instance for each mock type

### LINKED ISSUE #212: [mock] Use a single mock instance for multiple servers
The Orderer and VC mocks support running a single instance with multiple server endpoints with different configurations (e.g., TLS).
This is useful for testing.
This issue requests to adopt this behavior to the rest of the mocks.

### LINKED ISSUE #227: [mock] Mock-Verifier breaks down if the test accidentally opens two streams to the same verifier
The mock-verifier have a single channel for all streams. If we open two streams, the requests/responses will get mangled between them.

liran-funaro:
#### Type of change

- Test update
 
#### Description

Modify the mock verifier:
- Use a single mock instance for each mock type
- Remove redundant methods
- Rename to `Verifier`

#### Related issues

- resolves #212 
- resolves #227

### DIFF
--- a/cmd/config/cobra_test_exports.go
+++ b/cmd/config/cobra_test_exports.go
@@ -43,7 +43,7 @@ type CommandTest struct {
 // StartDefaultSystem starts a system with mocks for CMD testing.
 func StartDefaultSystem(t *testing.T) SystemConfig {
 	t.Helper()
-	_, verifier := mock.StartMockSVService(t, 1)
+	_, verifier := mock.StartMockVerifierService(t, 1)
 	_, vc := mock.StartMockVCService(t, 1)
 	_, orderer := mock.StartMockOrderingServices(t, &mock.OrdererConfig{NumService: 1})
 	_, coordinator := mock.StartMockCoordinatorService(t)
--- a/loadgen/client_test.go
+++ b/loadgen/client_test.go
@@ -160,7 +160,7 @@ func TestLoadGenForCoordinator(t *testing.T) {
 					t.Parallel()
 					clientConf := DefaultClientConf(t)
 					clientConf.Limit = limit
-					_, sigVerServer := mock.StartMockSVService(t, 1)
+					_, sigVerServer := mock.StartMockVerifierService(t, 1)
 					_, vcServer := mock.StartMockVCService(t, 1)
 
 					cConf := &coordinator.Config{
--- a/mock/coordinator.go
+++ b/mock/coordinator.go
@@ -28,7 +28,7 @@ import (
 	"github.com/hyperledger/fabric-x-committer/utils/grpcerror"
 )
 
-// Coordinator is a mock coordinator.
+// Coordinator is a mock implementation of servicepb.CoordinatorServer.
 type Coordinator struct {
 	servicepb.CoordinatorServer
 	lastCommittedBlock atomic.Pointer[servicepb.BlockRef]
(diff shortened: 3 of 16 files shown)


cendhu:
@liran-funaro I am not in favor of making the mock too restrictive and do not see much benefit in sharing a single instance for all gRPC servers.

1. Complexity & Safety: Enforcing thread safety in a mock adds unnecessary cognitive load and maintenance overhead, as every future modification requires careful locking to prevent race conditions. Simple, isolated instances are naturally thread-safe by design and far less prone to introducing flaky test behavior.
2. Observability: Sharing a single instance blurs the test topology, making it impossible to assert that a specific server node received a particular request. We lose the ability to verify specific routing logic or load distribution since all traffic merges into one sink.
3. Granular Fault Injection: We lose the capability to configure asymmetric fault scenarios, such as having only one specific node drop packets via MockFaultyNodeDropSize while others remain healthy. This prevents us from testing partial outages or quorum recovery mechanisms effectively.
4. Test Flexibility: Coupling all gRPC servers to a single stateful object removes the flexibility to simulate independent node behaviors, such as resetting one server while keeping others alive. It forces a shared-state architecture that fundamentally contradicts the distributed, shared-nothing nature of the system we are testing.

liran-funaro:
@cendhu Thanks for the comment. I addressed some of the issues in the PR, and I hope the explanation below clarifies why I believe this change is beneficial.

First, using a single instance is required to correctly mock the **Orderer** and the **VC**, as both operate on a globally shared state. The **sidecar** and **coordinator** are already singletons. The **verifier** is the only stateless component: all its state is stream-scoped, and no state is preserved once a stream is closed. As a result, the verifier behaves identically whether streams are served by distinct instances or by a shared one.

The goal of this PR is to align all mocks to a consistent pattern for readability and maintainability. Since all other mock services already use a single instance, applying the same pattern to the verifier improves consistency. It also reduces code clutter by removing the `index` parameter from many methods where it was effectively ignored.

Addressing your concerns:

**Complexity & Safety:**
The mock verifier already required synchronization due to parallel streams within a single service. Using a shared instance does not introduce new thread-safety requirements. In fact, the change simplifies the logic by making all state explicitly stream-local rather than partially relying on shared global state. An additional benefit is that if the unit under test behaves incorrectly (e.g., opens multiple streams to the same verifier), the previous implementation would not handle this well, whereas the new implementation handles it correctly and fails the test as expected.

**Observability:**
This is addressed by recording the server address for each stream, making it possible to assert that a stream was opened by the expected endpoint. This mechanism is used in tests where such validation is required.

**Granular Fault Injection:**
The PR supports per-server logic. Specifically, the `MockFaultyNodeDropSize` parameter is stream-scoped, allowing error injection within the scope of a specific stream tied to a specific server endpoint, as demonstrated in the tests.

**Test Flexibility:**
The PR allows simulating independent node behaviors. Individual gRPC endpoints can be stopped independently of the underlying mock instance, which is also demonstrated in the tests.

This PR does not force a shared state. As shown above, it preserves the ability to simulate a distributed system while providing a single control point, which simplifies test orchestration and maintenance.

cendhu:
@liran-funaro Okay, I will review this code. In general, the mock has to be as dumb as possible. This is what most mock generator provides. Anyway, we can live with it. In future, if our current approach causes more complexity to the mock, we should be open to revert to a simple mocks.

cendhu (approved):
LGTM.

PR #228, opened 2026-01-13
[loadgen] Rate limiter with maximal wait and minimal batch size

### LINKED ISSUE #220: [loadgen] Limiter not correcly pasing it self
The limiter submits many TXs in the beginning, and then it blocks because it has passed the limit.
Also, since the generator generates "blockSize" transactions in one batch, if the limit is lower than this, then we wait for a long time.

### LINKED ISSUE #227: [mock] Mock-Verifier breaks down if the test accidentally opens two streams to the same verifier

liran-funaro:
#### Type of change

- Bug fix
- New feature
- Breaking change
 
#### Description

- Introduce a new rate controller. Unlike standard rate limiters, it enforces both a latency deadline and a minimum useful batch, making it suitable for block- and batch-driven systems rather than per-request throttling.
- Use protobuf to generate the HTTP server for the rate limiter.
- Use `connection.Server` instead of `connection.Endpoint` to allow configuring TLS if needed.
- Refactor `connection`'s TLS to allow creating `tls.Config`.

#### Related issues

- resolves #220 
- resolves #227


loadgen/workload/config.go
```diff
@@ -159,10 +159,17 @@ type StreamOptions struct {
 }
 
 // LimiterConfig is used to create a limiter.
+// TXs are released at Rate (default: unlimited).
+// Blocks/batches wait up to MaxBatchWait (default: 1 second) before submission.
+// If a full batch is not ready by then, a partial batch is
```
cendhu: Where do we define the full batch size?

loadgen/workload/config.go
```diff
@@ -159,10 +159,17 @@ type StreamOptions struct {
...
+// submitted if it meets MinBatchSize (default: 1);
+// otherwise, the system waits until MinBatchSize is available.
 type LimiterConfig struct {
-	InitialLimit rate.Limit `mapstructure:"initial-limit"`
-	// Endpoint for a simple http server to set the limiter.
-	Endpoint connection.Endpoint `mapstructure:"endpoint"`
+	Rate         uint64        `mapstructure:"rate"`
+	MaxBatchWait time.Duration `mapstructure:"max-batch-wait"`
+	MinBatchSize uint64        `mapstructure:"min-batch-size"`
```
cendhu: `MaxBatchWait` and `MinBatchSize` are contradicting. The `MaxBatchWait` should be the hard flush and only the timer can be reset if the batch is empty. Further, how are these used when we limit the total number of transactions sent? A deadlock can occur if the remaining transaction quota is smaller than `MinBatchSize`. 

I would suggest we follow something similar to the orderer config but we can use the timeout and max batch size alone. The use-case was to make the loadgen work with smaller rates on low end machine. I think a hard timeout suffice. 

```yaml
BatchTimeout: 1s // hard flush
BatchSize:
  MaxMessageCount: 10
  AbsoluteMaxBytes: 10 MB
```

The max message count and absolute max bytes may not be applicable to loadgen.

loadgen/workload/limiter.go
```diff
@@ -7,15 +7,164 @@ SPDX-License-Identifier: Apache-2.0
...
+	}
+	logger.Debugf("Setting limit to %d requests per second.", rate)
+	l := &RateLimiter{maxWait: maxWait, minBatch: minBatch}
+	l.SetRate(rate)
+	return l
+}
+
+// Take requests up to requested tokens from the limiter.
+// The call may block for some time, and may return fewer tokens
+// than requested depending on timing and configuration.
```
cendhu: I think this is not supported in `x/time/rate` package. Please add the reason for not reusing the existing package which would help us in the future.

loadgen/workload/limiter.go
```diff
@@ -7,15 +7,164 @@ SPDX-License-Identifier: Apache-2.0
...
+// of tokens available at the specified rate.
+//
+// availableTokens = (duration/second) * rate
+// We want requestedTokens < availableTokens
+// ==> requestedTokens < (duration/second) * rate
+// ==> (duration/second) > requestedTokens / rate
+// ==> duration > (requestedTokens / rate) * second
+// ==> duration > requestedTokens * second / rate.
+func expectedDurationForTokens(rate uint64, requestedTokens int64) time.Duration {
+	return time.Duration(math.Floor(float64(requestedTokens) * float64(time.Second) / float64(rate)))
```
cendhu: why does the following work? 

```go 
return time.Duration(requestedTokens) * time.Second / time.Duration(rate)
```

loadgen/workload/limiter.go
```diff
@@ -7,15 +7,164 @@ SPDX-License-Identifier: Apache-2.0
...
+	l := &RateLimiter{maxWait: maxWait, minBatch: minBatch}
+	l.SetRate(rate)
+	return l
+}
+
+// Take requests up to requested tokens from the limiter.
+// The call may block for some time, and may return fewer tokens
+// than requested depending on timing and configuration.
+//
+// The limiter chooses one of three outcomes:
```
cendhu: Can't we use x/time/rate? Is the Partial fill not possilbe with Reserve() and Delay()?

loadgen/workload/limiter.go
```diff
@@ -7,15 +7,164 @@ SPDX-License-Identifier: Apache-2.0
...
+		rate:      rate,
+		maxWait:   max(0, l.maxWait),  // ensure maxWait is non-negative.
+		minBatch:  max(1, l.minBatch), // ensure minBatch is at least 1.
+	})
+}
+
+// take calculates how much time should have passes for the total number of tokens to be taken.
+// If not enough time has passed, it waits for the remaining time or until maxWait is reached.
+// It then returns the number of tokens that can be taken within the allowed time.
+func (l *limiterState) take(ctx context.Context, requestedTokens int64) int64 {
```
cendhu: `take` can be renamed to `acquire`

loadgen/workload/limiter.go
```diff
@@ -7,15 +7,164 @@ SPDX-License-Identifier: Apache-2.0
...
+	if l.rate == 0 {
+		return requestedTokens
+	}
+
+	// This is the linearization point which determines the place in the timeline for this request.
+	// We might return some of the tokens after this point, but we do not read this value again.
+	takenTokens := l.taken.Add(requestedTokens)
+
+	// Calculate the expected duration to have taken 'takenTokens' tokens.
+	expectedDuration := expectedDurationForTokens(l.rate, takenTokens)
```
cendhu:
```suggestion
	readyAt := expectedDurationForTokens(l.rate, takenTokens)
```

loadgen/workload/limiter.go
```diff
@@ -7,15 +7,164 @@ SPDX-License-Identifier: Apache-2.0
...
+	elapsedDuration := time.Since(l.startTime)
+
+	// If enough time has passed, return immediately.
+	if elapsedDuration > expectedDuration {
```
cendhu:
```suggestion
	if age > readyAt {
```

loadgen/workload/limiter.go
```diff
@@ -7,15 +7,164 @@ SPDX-License-Identifier: Apache-2.0
...
+	}
+
+	// We want to wait only up to maxWait.
+	// However, if the requestedTokens is already less or equal to minBatch, we have to wait the entire duration.
+	remainingDuration := expectedDuration - elapsedDuration
+	if l.maxWait > 0 && remainingDuration > l.maxWait && uint64(requestedTokens) > l.minBatch {
+		// Calculate how many tokens we overtook by limiting the wait to maxWait.
+		over := expectedTokensForDuration(l.rate, remainingDuration-l.maxWait)
+		// We deliver the difference, or minBatch (the higher between them).
+		deliveringTokens := max(requestedTokens-over, int64(l.minBatch)) //nolint:gosec // uint64 -> int64.
```
cendhu:
```suggestion
		granted := max(requestedTokens-over, int64(l.minBatch)) //nolint:gosec // uint64 -> int64.
```

loadgen/workload/limiter.go
```diff
@@ -7,15 +7,164 @@ SPDX-License-Identifier: Apache-2.0
...
+		if deliveringTokens < requestedTokens {
```
cendhu:
```suggestion
		if granted < requestedTokens {
```

loadgen/workload/limiter.go
```diff
@@ -7,15 +7,164 @@ SPDX-License-Identifier: Apache-2.0
...
+			diff := requestedTokens - deliveringTokens
```
cendhu:
```suggestion
			refund := requestedTokens - deliveringTokens
```

api/servicepb/loadgen.proto
```diff
@@ -11,13 +11,23 @@ option go_package = "github.com/hyperledger/fabric-x-committer/api/servicepb";
...
-    rpc SetLimit(Limit) returns (google.protobuf.Empty);
+    rpc GetRateLimit(google.protobuf.Empty) returns (RateLimit) {
+        option (google.api.http) = {
+            get: "/getRateLimit"
+        };
+    };
+    rpc SetRateLimit(RateLimit) returns (google.protobuf.Empty) {
+        option (google.api.http) = {
+            post: "/setRateLimit"
+            body: "*"
```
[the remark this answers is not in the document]
liran-funaro: This is what was previously implemented. Anyway, it is uncommon to use `/setRateLimit/{rate}` for updates. We should use the body for that. But we can skip the JSON and just post the value to simplify things.
The main user here is @pasquale95 . Whatever he prefers works for me.

loadgen/workload/config.go
liran-funaro: The total limit does not affect the sending behaviour. We stop sending once we passed the limit, but if the block size is larger than what was remained so we will surpass the limit by some amount.
Anyway, the limiter supports fetching smaller batches. If we request smaller number of tokens, we don't wait for min-batch-size. This was omitted by mistake from the docs. I'll add it back.

api/servicepb/loadgen.proto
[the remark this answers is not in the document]
cendhu: I am fine with JSON body as we can easily add additional parameters to them without breaking the compatibility.

loadgen/workload/config.go
liran-funaro: Added the following comment:
If the block size is less than or equal to MinBatchSize, MaxBatchWait is ignored.

loadgen/workload/limiter.go
```diff
@@ -7,15 +7,164 @@ SPDX-License-Identifier: Apache-2.0
...
+		return requestedTokens
+	}
+
+	// This is the linearization point which determines the place in the timeline for this request.
+	// We might return some of the tokens after this point, but we do not read this value again.
+	takenTokens := l.taken.Add(requestedTokens)
+
+	// Calculate the expected duration to have taken 'takenTokens' tokens.
+	expectedDuration := expectedDurationForTokens(l.rate, takenTokens)
+	elapsedDuration := time.Since(l.startTime)
```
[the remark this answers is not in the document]
liran-funaro: done

loadgen/client_test.go
```diff
@@ -459,6 +464,86 @@ func testLoadGenerator(t *testing.T, c *ClientConfig) {
...
+	t.Parallel()
+	clientConf := DefaultClientConf(t)
+	clientConf.Adapter.VerifierClient = startVerifiers(t, test.InsecureTLSConfig, test.InsecureTLSConfig)
+	curRate := uint64(100)
+	clientConf.Stream.RateLimit = curRate
+	// We use small wait to ensure the rate limiter serves in low granularity.
+	clientConf.LoadProfile.Block.PreferredRate = 10 * time.Millisecond
+	clientConf.LoadProfile.Block.MaxSize = 100
+	clientConf.LoadProfile.Block.MinSize = 1
+	clientConf.HTTPServer = connection.NewLocalHostServerWithTLS(test.InsecureTLSConfig)
```
cendhu: nit. unrelated to this test. As we do not have `NewLocalHostServer`, we can remove the `WithTLS`.

liran-funaro:
@cendhu I think you have a good point that this can be solved via better adjusting the configuration.
Maybe allowing changing the block size limits dynamically, along with the rate, can help as well.
This PR allows automatic block size adjustment, instead of relying on the user.
But I see how this can be an overkill.

>If the configured batchSize is higher than the configured rate, throw an error during config validation.

This is good only for the initial rate limit, not when adjusting the limit dynamically.

>If the configured rate is much higher than what the server can handle and if the configured batchSize is higher than the actual rate, log a warning. If the user has configured incorrectly for the available hardware resources, the impact would be seen and the loadgen does not need to be adaptive. When we benchmarked, it was always trial and error.

I agree. Slow producer is not a big issue.

>The rate at which the loadgen submits transactions to the orderer would be adapted anyway based on the router performance and committer throughput. Hence, it is necessary for the user to configure the rate and batch size appropriately for performance benchmarking.

GRPC should be able to pace the submitter anyway if the receiver cannot handle the load.
So adjusting the rate is only useful to reduce the latency.

PR #229, opened 2026-01-13
Use Identity in SignedData to avoid serialization

cendhu:
#### Type of change

#### Description

Instead of using the serialized bytes of identity in the signed data, this commit makes the committer to pass the Identity directly to avoid unnecessary serialization.

#### Related issues

  - https://github.com/hyperledger/fabric-x-common/issues/44


utils/signature/sigtest/digest.go
```diff
@@ -1,58 +0,0 @@
-/*
```
liran-funaro: Did I forget to remove this file?

PR #248, opened 2026-01-19
[loadgen] Create orderer node for each orderer endpoint

### LINKED ISSUE #247: [loadgen] Create Orderer node for each given Orderer endpoint
This is required to generate the TLS credentials for that endpoint.

liran-funaro:
#### Type of change

#### Description

- Create an orderer node for each orderer endpoint using the endpoint's host as the SANS for this node.

#### Related issues

- resolves #247

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ docker/runner/out
 test/**/*.log
 
 .DS_Store
+venv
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -37,3 +37,4 @@ docker/runner/out
 test/**/*.log
 
 .DS_Store
+venv
```
cendhu: venv for python? is it used for sqlfluff?

cendhu (approved):
LGTM.

.gitignore
liran-funaro: I use it locally in my environment because my Mac prevents me from installing packages globally.

PR #255, opened 2026-01-20
speed-up config block path

### LINKED ISSUE #249: Speedup single TX per block path for quick config-block processing
Currently, the timeouts defined in various parts of the system are too long and slow down the processing of the config block.
We should add a fast pass for the config block to avoid waiting for these timeouts.

cendhu:
#### Type of change

- Bug fix

#### Description

Currently, the timeouts defined in various parts of the system are too long and slow down the processing of the config block. This commit add a fast pass for the config block to avoid waiting for these timeouts.

#### Related issues

  - resolves #249

### DIFF
--- a/service/vc/test_exports.go
+++ b/service/vc/test_exports.go
@@ -52,27 +52,12 @@ type (
 	}
 )
 
-// DefaultVCTestEnvOpts returns default options for VC test environment.
-func DefaultVCTestEnvOpts() *TestEnvOpts {
-	return &TestEnvOpts{
-		NumServices: 1,
-		ServerCreds: test.InsecureTLSConfig,
-		ResourceLimits: &ResourceLimitsConfig{
-			MaxWorkersForPreparer:             2,
-			MaxWorkersForValidator:            2,
-			MaxWorkersForCommitter:            2,
-			MinTransactionBatchSize:           1,
-			TimeoutForMinTransactionBatchSize: 20 * time.Second,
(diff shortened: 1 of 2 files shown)


service/vc/test_exports.go
```diff
@@ -42,47 +42,63 @@ type (
...
-// NewValidatorAndCommitServiceTestEnvWithTLS creates a new test environment with a vcservice and a database.
-// It allows TLS with the acceptance of server creds.
-func NewValidatorAndCommitServiceTestEnvWithTLS(
-	t *testing.T,
-	numServices int,
-	serverCreds connection.TLSConfig, // one credentials set for all the vc-services.
-	db ...*DatabaseTestEnv,
-) *ValidatorAndCommitterServiceTestEnv {
+// DefaultVCTestEnvOpts returns default options for VC test environment.
+func DefaultVCTestEnvOpts() *TestEnvOpts {
```
liran-funaro: minor: `DefaultVCTestEnvOpts()` is only used inside `NewValidatorAndCommitServiceTestEnv()`.
Can we inline it? Or make it a private method.
Also, it should be after positioned `NewValidatorAndCommitServiceTestEnv()`

liran-funaro (approved):
LGTM

service/vc/test_exports.go
cendhu: Made it private.

PR #309, opened 2026-02-10
[sidecar] sync interval for block commit

### LINKED ISSUE #305: [sidecar] support async block commit to block store
To improve the performance, we should allow async write of blocks to the block store and then do sync periodically.

cendhu:
#### Type of change

#### Description

Allows skipping fsync on intermediate blocks to improve write throughput. Un-synced blocks are recoverable from the orderer on crash.

#### Related issues

  - resolves #305


liran-funaro (approved):
LGTM.
I suggest adding an exception for a config block. That is, syncing after a config block.
This will ensure easier recovery with negligible performance cost.

Issue #61, opened 2025-07-06
[tablet] Tablet split for evaluation


liran-funaro:
A tablet split into 120 tablets is required to get a stable performance in the evaluation.
Reintroducing this has the following issues:
- Requires non-compliant SQL (`WITH TABLETS=120`)
- Flaky test issues

Alternative solutions:
1. Use build tags to ensure this option is enabled only in production or evaluation
2. Use different Yugabyte flags in the production/evaluation environment.
`--ysql_num_shards_per_tserver` flag can be set to force the number of tablets per table.

For evaluation, we would like a static environment. For that, we can use the following flags:
```
--enable_automatic_tablet_splitting=false --ysql_num_shards_per_tserver=10
``` 

Here is the relevant information taken from [`--ysql_num_shards_per_tserver`](https://docs.yugabyte.com/preview/reference/configuration/yb-master/#ysql-num-shards-per-tserver)
>The number of shards (tablets) per YB-TServer for each YSQL table when a user table is created.
>
>Default: -1, where the number of shards is determined at runtime, as follows:
>
>If [enable_automatic_tablet_splitting](https://docs.yugabyte.com/preview/reference/configuration/yb-master/#enable-automatic-tablet-splitting) is true
>
>The default value is considered as 1.
>For servers with 4 CPU cores or less, the number of tablets for each table doesn't depend on the number of YB-TServers. >Instead, for 2 CPU cores or less, 1 tablet per cluster is created; for 4 CPU cores or less, 2 tablets per cluster are created.
>If enable_automatic_tablet_splitting is false
>
>For servers with up to two CPU cores, the default value is considered as 2.
>For servers with three or four CPU cores, the default value is considered as 4.
>Beyond four cores, the default value is considered as 8.

PR #311, opened 2026-02-10
[vc] split tables into predefined number of tablets

### LINKED ISSUE #308: [vc] enable pre-defined tablet splitting for tables.
We should split tables into predefined tablets count for YugabyteDB if such a value is configured or set via yaml file. If this value is not set or using PostgreSQL, the tables will not be split into predefined tablets.

cendhu:
#### Type of change

#### Description

When a new table is created, YugabyteDB starts with a single tablet that handles all initial writes. Auto-splitting only kicks in after the tablet reaches a size threshold (~1 GB by default), meaning the first N transactions all funnel through one tablet on one tserver — creating a severe bottleneck during the critical warm-up phase.

By pre-splitting tables at CREATE TABLE time (e.g., SPLIT INTO 3 TABLETS for a 3-node cluster), data is hash-distributed across all tservers from the very first write. This converts cold-start throughput from single-tserver-bound to cluster-wide from block 1.

This applies to all three table types: metadata, tx_status, and per-namespace state tables (ns_*). The setting auto-detects the DB backend and is silently ignored on PostgreSQL, which has no tablet concept.

By default, we are not enabling the table pre-split in config so that the test can run without stress. 

#### Related issues

  - resolves #308 #61

### DIFF
--- a/.sqlfluff
+++ b/.sqlfluff
@@ -6,7 +6,7 @@
 
 # Supported dialects https://docs.sqlfluff.com/en/stable/perma/dialects.html
 # Or run 'sqlfluff dialects'
-dialect = snowflake
+dialect = postgres
 
 # One of [raw|jinja|python|placeholder]
 templater = placeholder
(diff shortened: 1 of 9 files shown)


cendhu:
> Issue #61 suggests using `--ysql_num_shards_per_tserver`. Wasn't it useful?

Let me do a run with this setting.

cendhu:
The default behavior without the 'split into' or num shard flags
Screenshot 2026-02-11 at 3 33 21 PM

When using --ysql_num_shards_per_tserver=2 with 11 tservers (resulting in 22 tablets), I suspect the tablets are dynamic or that rebalancing is taking place. We still have to wait for stabilization during experiments, although it is faster than the default.
Screenshot 2026-02-11 at 3 20 45 PM

In contrast, with a predefined split into N tablets (where N=24), stabilization is immediate.
Screenshot 2026-02-11 at 3 23 40 PM

I understand this adds about 140 lines to the production code, but it saves time by eliminating the need to wait 20 minutes for the system to settle or having to discard the first 20 minutes of data.

cendhu:
Let me also try with --ysql_num_shards_per_tserver=10

liran-funaro (approved):
LGTM. This is fine because it is an opt-in configuration.

PR #314, opened 2026-02-11
Refactor DB container lifecycle to use shared TestMain container

### LINKED ISSUE #312: [macOS] DB_DEPLOYMENT=container go test -v on service/vc blocks indefinitely

cendhu:
#### Type of change

#### Description

  - Refactor container methods to return errors instead of calling t.Fatal()
  - Add SetupSharedContainer()/CleanupSharedContainer() for TestMain
  - Use UUID-suffixed container names and auto-assigned host ports
  - Remove findContainer() — always create fresh containers
  - Move real DB setup out of StartDefaultSystem() into tests that need it

#### Related issues

  - resolves #312

### DIFF
--- a/integration/test/main_test.go
+++ b/integration/test/main_test.go
@@ -18,6 +18,7 @@ import (
 	"google.golang.org/grpc/grpclog"
 
 	"github.com/hyperledger/fabric-x-committer/integration/runner"
+	"github.com/hyperledger/fabric-x-committer/service/vc/dbtest"
 )
 
 func TestMain(m *testing.M) {
@@ -43,5 +44,6 @@ func TestMain(m *testing.M) {
 
 	// Waiting a seconds solves a bug where the binaries are not yet accessible by the filesystem.
 	time.Sleep(time.Second)
-	os.Exit(m.Run())
+
+	dbtest.RunTestMain(m)
 }
--- a/service/coordinator/main_test.go
+++ b/service/coordinator/main_test.go
@@ -1,0 +1,17 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package coordinator
+
+import (
+	"testing"
+
+	"github.com/hyperledger/fabric-x-committer/service/vc/dbtest"
+)
+
+func TestMain(m *testing.M) {
+	dbtest.RunTestMain(m)
+}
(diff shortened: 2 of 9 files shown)


cendhu:
@liran-funaro ready now.

integration/test/main_test.go
```diff
@@ -43,5 +44,6 @@ func TestMain(m *testing.M) {
 
 	// Waiting a seconds solves a bug where the binaries are not yet accessible by the filesystem.
 	time.Sleep(time.Second)
-	os.Exit(m.Run())
+
+	dbtest.RunTestMain(m)
```
liran-funaro: We must call `os.Exit()` with the error code returned from `m.Run()`.
Otherwise, our test will always show "pass" even if tests fail.

service/vc/main_test.go
```diff
@@ -0,0 +1,17 @@
...
+package vc
+
+import (
+	"testing"
+
+	"github.com/hyperledger/fabric-x-committer/service/vc/dbtest"
+)
+
+func TestMain(m *testing.M) {
+	dbtest.RunTestMain(m)
```
liran-funaro: We must call `os.Exit()` with the error code returned from `m.Run()`.
Otherwise, our test will always show "pass" even if tests fail.

service/query/main_test.go
```diff
@@ -0,0 +1,17 @@
...
+package query
+
+import (
+	"testing"
+
+	"github.com/hyperledger/fabric-x-committer/service/vc/dbtest"
+)
+
+func TestMain(m *testing.M) {
+	dbtest.RunTestMain(m)
```
liran-funaro: We must call `os.Exit()` with the error code returned from `m.Run()`.
Otherwise, our test will always show "pass" even if tests fail.

service/coordinator/main_test.go
```diff
@@ -0,0 +1,17 @@
...
+package coordinator
+
+import (
+	"testing"
+
+	"github.com/hyperledger/fabric-x-committer/service/vc/dbtest"
+)
+
+func TestMain(m *testing.M) {
+	dbtest.RunTestMain(m)
```
liran-funaro: We must call `os.Exit()` with the error code returned from `m.Run()`.
Otherwise, our test will always show "pass" even if tests fail.

dean-amar (approved):
LGTM.
Just a minor comment.

Issue #306, opened 2026-02-10
[sidecar] optimize mapBlock method


cendhu:
The mapBlock methods runs UnmarshalEnvelope and UnmarshalTx for every transaction. At high scale (~200k tps), these unmarshaling operations are become very expensive due to high frequency of memory allocation. For a block of 500 transactions, the mapBlock takes around 2 msecs which limits the overall system throughput. Hence, we need to optimize this to improve the performance of mapBlock.

PR #316, opened 2026-02-11
Introduce UnmarshalEnvelopeLite to speed up mapBlock

cendhu:
#### Type of change

#### Description

  Replace full proto.Unmarshal of Envelope→Payload→ChannelHeader (3 chained
  deserializations) with direct protobuf wire-format scanning that extracts
  only the needed fields (HeaderType, TxID, Data). This reduces envelope
  parsing from 287 ns/op to 72 ns/op — a 4x speedup with 7 fewer allocations.
```
  BenchmarkUnwrapEnvelope-14        4079934    286.7 ns/op   616 B/op   9 allocs/op
  BenchmarkUnwrapEnvelopeLite-14   16929306     71.5 ns/op    72 B/op   2 allocs/op

  [New mapBlock]
  BenchmarkMapBlock/txs=500-14                   2850            400160 ns/op          663662 B/op      13579 allocs/op
  BenchmarkMapBlock/txs=500-14                   2599            405622 ns/op          664974 B/op      13594 allocs/op
  BenchmarkMapBlock/txs=500-14                   2767            408145 ns/op          657391 B/op      13505 allocs/op

  [Old mapBlock]
  BenchmarkMapBlock/txs=500-14                   1497            677487 ns/op         1165977 B/op      17648 allocs/op
  BenchmarkMapBlock/txs=500-14                   1584            659822 ns/op         1168212 B/op      17674 allocs/op
  BenchmarkMapBlock/txs=500-14                   1808            635695 ns/op         1158503 B/op      17561 allocs/op
```
  With UnmarshalEnvelopeLite, the new mapBlock is able to process 1.7x more blocks.

#### Related issues

  - partially addresses #306

### DIFF
--- a/service/sidecar/mapping_test.go
+++ b/service/sidecar/mapping_test.go
@@ -20,7 +20,7 @@ import (
 	"github.com/hyperledger/fabric-x-committer/utils"
 )
 
-func BenchmarkMapBlock(b *testing.B) {
+func BenchmarkMapOneBlock(b *testing.B) {
 	flogging.ActivateSpec("fatal")
 	txs := workload.GenerateTransactions(b, nil, b.N)
 	block := workload.MapToOrdererBlock(1, txs)
@@ -31,9 +31,12 @@ func BenchmarkMapBlock(b *testing.B) {
 	b.StopTimer()
 	require.NoError(b, err, "This can never occur unless there is a bug in the relay.")
 	require.NotNil(b, mappedBlock)
-	flogging.Init(flogging.Config{LogSpec: "fatal"})
+}
+
+func BenchmarkMapBlockSize(b *testing.B) {
+	flogging.ActivateSpec("fatal")
 	for _, blockSize := range []int{100, 1000, 5000, 10000} {
-		b.Run(fmt.Sprintf("txs=%d", blockSize), func(b *testing.B) {
+		b.Run(fmt.Sprintf("blockSize=%d", blockSize), func(b *testing.B) {
 			// Generate b.N total TXs so each iteration processes a unique block,
 			// avoiding cache/locality effects from reusing the same block.
 			// Pad to ensure at least one full block.
@@ -62,7 +65,7 @@ func BenchmarkMapBlock(b *testing.B) {
 				}
 				blockIdx++
 			}
-			b.ReportMetric(float64(b.N*blockSize)/b.Elapsed().Seconds(), "tx/s")
+			b.ReportMetric(float64(b.N)/b.Elapsed().Seconds(), "tx/s")
 		})
 	}
 }
--- a/utils/serialization/envelope_wrapper_test.go
+++ b/utils/serialization/envelope_wrapper_test.go
@@ -163,8 +163,8 @@ func FuzzUnwrapEnvelopeLiteNoPanic(f *testing.F) {
 	f.Add([]byte("not a protobuf"))
 	f.Add([]byte{0x0a, 0x00})
 
-	f.Fuzz(func(_ *testing.T, data []byte) {
-		_, _ = serialization.UnwrapEnvelopeLite(data) // checking for panics only
+	f.Fuzz(func(t *testing.T, data []byte) {
+		assertLiteMatchesOriginal(t, data)
 	})
 }
 
@@ -198,6 +198,7 @@ func BenchmarkUnwrapEnvelope(b *testing.B) {
 			b.Fatal(err)
 		}
 	}
+	b.ReportMetric(float64(b.N)/b.Elapsed().Seconds(), "tx/s")
 }
 
 func BenchmarkUnwrapEnvelopeLite(b *testing.B) {
@@ -209,6 +210,7 @@ func BenchmarkUnwrapEnvelopeLite(b *testing.B) {
 			b.Fatal(err)
 		}
 	}
+	b.ReportMetric(float64(b.N)/b.Elapsed().Seconds(), "tx/s")
 }
 
 // loadgenEnvelopes generates realistic serialized envelopes using the load


utils/serialization/envelope.go
```diff
@@ -48,3 +50,201 @@ func ParseEnvelope(envelope *common.Envelope) (*common.Payload, *common.ChannelH
...
+}
+
+// extractField scans a protobuf wire-format message for the first occurrence
+// of targetField with the given expectedType. It returns the unconsumed slice
+// starting at the field's value, so the caller can decode it with the
+// appropriate protowire.ConsumeXxx function.
+func extractField(b []byte, targetField protowire.Number, expectedType protowire.Type) ([]byte, error) {
+	for len(b) > 0 {
+		num, wtype, n := protowire.ConsumeTag(b)
+		if n < 0 {
```
liran-funaro: major: We need to ensure progress
```suggestion
		if n <= 0 {
```

utils/serialization/envelope.go
```diff
@@ -48,3 +50,201 @@ func ParseEnvelope(envelope *common.Envelope) (*common.Payload, *common.ChannelH
...
+			return b, nil
+		}
+
+		vn := protowire.ConsumeFieldValue(num, wtype, b)
+		if vn < 0 {
+			return nil, errors.New("invalid protobuf field value")
+		}
+		b = b[vn:]
+	}
+	return nil, nil
```
liran-funaro: minor: It is confusing to follow the cases that this method returns `nil, nil`.
I suggest returning `nil, err` or `[]bye, nil` all the time.
This also applies to `extractBytesField()`.

service/sidecar/mapping.go
```diff
@@ -166,17 +174,19 @@ func (b *blockMappingResult) rejectNonDBStatusTx(
...
+	debugTx(envLite, "excluded: %s (%s)", &status, reason)
 	return nil
 }
 
-func (b *blockMappingResult) addTxIDMapping(txNum uint32, hdr *common.ChannelHeader) (idAlreadyExists bool, err error) {
-	_, idAlreadyExists = b.txIDToHeight.LoadOrStore(hdr.TxId, servicepb.Height{
+func (b *blockMappingResult) addTxIDMapping(txNum uint32, envLite *serialization.EnvelopeLite) (
+	idAlreadyExists bool, err error,
+) {
+	_, idAlreadyExists = b.txIDToHeight.LoadOrStore(envLite.TxID, servicepb.Height{
```
liran-funaro: I assume a major performance improvement can be achieved by replacing `txIDToHeight` to be a regular  map as the mapping processes does not require any parallel access.
But this requires careful redesign.

utils/serialization/envelope.go
```diff
@@ -48,3 +50,201 @@ func ParseEnvelope(envelope *common.Envelope) (*common.Payload, *common.ChannelH
...
+// appropriate protowire.ConsumeXxx function.
+func extractField(b []byte, targetField protowire.Number, expectedType protowire.Type) ([]byte, error) {
+	for len(b) > 0 {
+		num, wtype, n := protowire.ConsumeTag(b)
+		if n < 0 {
+			return nil, errors.New("invalid protobuf tag")
+		}
+		b = b[n:]
+
+		if num == targetField {
```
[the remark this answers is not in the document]
cendhu: Yes. I am changing the existing UnwrapEnvelope to use auxilary proto and the new UnwrapEnvelopeLite to use wire scan so that both return same results in fuzzing. Otherwise, if timestamp is corrupted, UnwrapEnvelope returns error while UnwrapEnvelopeLite does not. Fuzzing helps to identity a lot of issues.

cendhu:
> I understand this PR aims to match protobuf behaviour as much as possible. But what is the motivation for that?

When I ran the fuzz test against the protobuf unmarshal, a few test cases were failing. Hence, I tried to match the unmarshalEnvelopeLite to match the protobuf behaviour. 

> 
> One case is clear: We should be able to unmarshal correctly ANY message that was encoded correctly. I.e., we should fail if the vanilla unmarshaller fails.
> 
> The question is whether this condition should be "if and only if". Do we allow a successful unmarshalling for cases that the vanilla unmarshaller would fail?

Yes, if the unrelated fields have some issues. To fix this, I have introduced auxiliary proto so that vanilla unmarshaler ignores irrelevant fields. 

> 
> If the answer is NO, then we MUST parse all the fields (not just the one we want). Because the vanilla unmarshaller might have failed on them.
> 
> But I'm not sure if this is the correct approach. The only part of the system that will unmarshal TXs is the committer. We can choose the behaviour that will be the most straightforward for us.

> Anyway, if we still choose to parse all the fields for FULL compliance, I'm not sure it will have a great performance impact. To my understanding, the main benefit comes from slicing the raw byte array instead of copying it multiple times. If this is the case, unmarshaling the entire envelope should not change the results too much.

what is your suggestion here? Should we remove the auxiliary proto and try to match the behavior of full unmarshaler? I do not understand the problem with the employed approach followed. Please clarify.

cendhu:
> @cendhu I suggest only matching the behaviour on success (relaxed approach).
> 
> Relaxed approach: If protubuf succeed -> we succeed If protobuf fail -> we may succeed or fail (don't care).
> 
> This will greatly simplify code as we won't have to parse non-relevant fields and we won't have to go over the entire buffer when the required fields are found.
> 
> My point was that partial matching of the behaviour (your approach) has non benefit over the relaxed approach. Either we have full match (protobuf fail -> we fail), or we can go with the relaxed approach.

Good point. I have simplified the code and also added the following comment.

```
// The orderer already validates envelope structure (channelHeader, signatureHeader)
// before including transactions in a block, so the committer only receives valid
// envelopes. The only scenario where malformed envelopes could arrive is if BFT
// consensus itself is compromised (more than f malicious nodes), at which point
// the system has already failed. Therefore, this function only needs to handle
// well-formed input correctly and does not attempt to match proto.Unmarshal's
// error behavior on malformed input.
```

Makefile
```diff
@@ -215,6 +216,10 @@ bench-sign: FORCE
 bench-sidecar: FORCE
 	$(go_cmd) test ./service/sidecar/... -bench "Benchmark.*" -run "^$$" | awk -f scripts/bench-tx-per-sec.awk
 
+# Run serialization benchmarks.
+bench-serialization: FORCE
+	$(go_cmd) test ./utils/serialization/... -bench "Benchmark.*" -run "^$$"
```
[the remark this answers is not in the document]
cendhu: I have used `b.ReportMetrics()`.

utils/serialization/envelope_wrapper_test.go
```diff
@@ -7,88 +7,238 @@ package serialization_test
...
+}
 
-	expectedChannelHeader := &common.ChannelHeader{
-		ChannelId: "test-channel",
+func BenchmarkUnwrapEnvelope(b *testing.B) {
+	envelopes := loadgenEnvelopes(b, b.N)
+	b.ResetTimer()
+	i := 0
+	for b.Loop() {
+		_, _, err := serialization.UnwrapEnvelope(envelopes[i%len(envelopes)])
```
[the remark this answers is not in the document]
cendhu: No. Changed to for range.

service/sidecar/mapping_test.go
```diff
@@ -29,6 +31,40 @@ func BenchmarkMapBlock(b *testing.B) {
...
+			blockIdx := 0
+			for b.Loop() {
+				var txIDToHeight utils.SyncMap[string, servicepb.Height]
+				_, err := mapBlock(blocks[blockIdx%numBlocks], &txIDToHeight)
+				if err != nil {
+					b.Fatal(err)
+				}
+				blockIdx++
+			}
+			b.ReportMetric(float64(b.N*blockSize)/b.Elapsed().Seconds(), "tx/s")
```
liran-funaro: This report is incorrect

```suggestion
			b.ReportMetric(float64(b.N)/b.Elapsed().Seconds(), "tx/s")
```

PR #324, opened 2026-02-12
[loadgen] Align block creation with updated API

liran-funaro:
#### Type of change

#### Description

- The new API for creating a genesis block support updating a block using existing crypto material. This will simplify TLS testing when updating the endpoints.


cendhu (approved):
LGTM.

PR #341, opened 2026-02-17
[query] Add unit-test for GetTransactionStatus

### LINKED ISSUE #294: [query] `readTxStatusRows()` coverage
`readTxStatusRows()` always returns an empty list in all the tests.
So we never reach the actual scanning code.
We should add test that returns at least one status.

liran-funaro:
#### Type of change

- Test update

#### Description

- Add unit-test for the query-service `GetTransactionStatus()`

#### Related issues

- resolves #294

### DIFF
--- a/service/query/query_service_test.go
+++ b/service/query/query_service_test.go
@@ -72,10 +72,9 @@ func TestQuerySecureConnection(t *testing.T) {
 func TestQuery(t *testing.T) {
 	t.Parallel()
 	env := newQueryServiceTestEnv(t, nil)
-	requiredItems := env.makeItems(t)
+	requiredItems := env.insertSampleKeysValueItems(t)
 	query, _, _ := makeQuery(requiredItems)
-	txIDs := env.makeTXs(t)
-	statusQuery := &committerpb.TxStatusQuery{TxIds: txIDs}
+	txIDs := env.insertSampleTxsStatus(t)
 	expectedStatus := make([]*committerpb.TxStatus, len(txIDs))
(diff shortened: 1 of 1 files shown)


service/query/query_service_test.go
```diff
@@ -88,20 +98,47 @@ func TestQuery(t *testing.T) {
...
 		require.NoError(t, err)
 		requireResults(t, requiredItems, ret.Namespaces)
 	})
 
+	t.Run("Query GetTransactionStatus interface", func(t *testing.T) {
+		t.Parallel()
+		ret, err := env.qs.GetTransactionStatus(t.Context(), statusQuery)
+		require.NoError(t, err)
+		require.NotNil(t, ret)
+		test.RequireProtoElementsMatch(t, expectedStatus, ret.Statuses)
```
cendhu: We can remove NotNit and do ret.GetStatuses.

service/query/query_service_test.go
```diff
@@ -88,20 +98,47 @@ func TestQuery(t *testing.T) {
...
+	})
+
 	t.Run("Query GetRows client", func(t *testing.T) {
 		t.Parallel()
 		ret, err := env.clientConn.GetRows(t.Context(), query)
 		require.NoError(t, err)
 		requireResults(t, requiredItems, ret.Namespaces)
 	})
 
+	t.Run("Query GetTransactionStatus client", func(t *testing.T) {
```
cendhu: We can retain this test but remove the one in line 119.

PR #364, opened 2026-02-22
[deliver] Remove non-production parameters

### LINKED ISSUE #360: [deliver] Verify block order

### LINKED ISSUE #362: [sidecar] Remove non-production parameters from delivery
Our delivery use case does not require the end block number, and never use a negative start block number.
We always want to pull a continues stream of blocks, starting from the last committed block
We currently use the end block for recovery, but this can be eliminated by joining the recovery delivery with the mainline workflow.
This will reduce the lines of code, and thus, the testing surface to cover.

liran-funaro:
#### Type of change

#### Description

- Verify block order in delivery client
- Remove end block number
- Rename start block number to next block number and change to `uint64`
- Modify the sidecar to avoid the need for "end block number", by moving the ledger recovery to the relay 

#### Related issues

- resolves #360 
- resolves #362

### DIFF
--- a/utils/deliver/deliver_cft.go
+++ b/utils/deliver/deliver_cft.go
@@ -35,9 +35,8 @@ type (
 
 	// Parameters needed for deliver to run.
 	Parameters struct {
-		StartBlkNum int64
-		EndBlkNum   uint64
-		OutputBlock chan<- *common.Block
+		NextBlockNum uint64
+		OutputBlock  chan<- *common.Block
 	}
 
 	// Stream requires the following interface.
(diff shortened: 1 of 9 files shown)


utils/deliver/deliver_cft.go
```diff
@@ -153,9 +148,18 @@ func (c *CftClient) deliverRelay(
 			return *status, nil
 		}
 
-		//nolint:gosec // integer overflow conversion uint64 -> int64
-		p.StartBlkNum = int64(block.Header.Number) + 1
-		logger.Debugf("next expected block number is %d", p.StartBlkNum)
+		// We make minimal verifications to ensure we receive blocks in order.
+		// This allows us to restart the connection from the next expected block upon failure.
```
cendhu: I assume when the validation fails, the delivery would be attempted from other nodes in the case of bft puller.

PR #372, opened 2026-02-23
[query-service] validate inputs and limit active views

### LINKED ISSUE #367: [input-validation] query service
The query service should send codes.InvalidArgument when

1. namespace is empty.
2. key set is empty
3. txID set is empty

### LINKED ISSUE #369: [concurrencyLimit] query service
Restrict the number of active views.

cendhu:
#### Type of change

#### Description

1. Adds MaxViewLimit
2. Return invalid argument error on empty keys, empty namespaces, empty txIDs in the query. 

#### Related issues

  - resolves #367 
  - resolves #369

### DIFF
--- a/utils/concurrency_limit_test.go
+++ b/utils/concurrency_limit_test.go
@@ -19,40 +19,41 @@ func TestConcurrencyLimiter(t *testing.T) {
 	t.Run("LimitReached", func(t *testing.T) {
 		t.Parallel()
 		counter := NewConcurrencyLimiter(1)
-		require.NoError(t, counter.TryAcquire(t.Context()))
-		require.Equal(t, int64(1), counter.Load())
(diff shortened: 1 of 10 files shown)


utils/concurrency_limit_test.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func TestConcurrencyLimiter(t *testing.T) {
+	t.Parallel()
+
+	t.Run("LimitReached", func(t *testing.T) {
+		t.Parallel()
+		counter := NewConcurrencyLimiter(1)
+		require.NoError(t, counter.TryAcquire(t.Context()))
+		require.Equal(t, int64(1), counter.Load())
```
liran-funaro: nit: We can use `require.EqualValues()` to avoid casting.
```suggestion
		require.EqualValues(t, 1, counter.Load())
```

PR #392, opened 2026-03-01
[sidecar & query] dynamic root CA certificate rotation without service restarts

### LINKED ISSUE #290: [sidecar, notification & query service] Dynamic updates of server root CAs according to config-block
Currently, we start each service with fixed acceptable root CAs.
We need to allow updates via the config-block to allow new peers to join the network (party).

### LINKED ISSUE #472: [DRAFT] Dynamic Root CA Certificate Rotation - Architecture Decision
# Dynamic Root CA Certificate Rotation - Architecture Discussion

## Summary

This document describes the architecture for dynamic root CA certificate rotation in our services, enabling TLS client certificate validation to retrieve new organization CAs from configuration blocks without service restarts. We compare our implementation with Hyperledger Fabric's proven approach and explain the design decisions behind it.

## Problem Statement

In multi-organization blockchain deployments, new organizations may join the channel after services have started. Each organization brings its own CA certificates that must be trusted for client authentication. Without dynamic CA rotation, services would need to restart to pick up new certificates, causing downtime and disrupting operations.

**Requirements:**
- Add new organization CAs without service restart
- Maintain high performance (minimal overhead per TLS handshake)
- Ensure thread safety during CA updates
- Support both event-driven (sidecar) and polling (query) services

### The Fabric Pattern

Hyperledger Fabric uses a **push-based architecture** where services update the server's TLS configuration when CA certificates change.

#### Key Components

**1. TLSConfig Wrapper (Stores the Built Config)**

```go
type TLSConfig struct {
    config *tls.Config      // Holds the complete TLS config
    lock   sync.RWMutex     // Protects concurrent access
}

// Called on CA updates (rare)
func (t *TLSConfig) SetClientCAs(certPool *x509.CertPool) {
    t.lock.Lock()
    defer t.lock.Unlock()
    t.config.ClientCAs = certPool  // Update the cert pool
}

// Called on handshakes (frequent)
func (t *TLSConfig) Config() tls.Config {
    t.lock.RLock()
    defer t.lock.RUnlock()
    return *t.config.Clone()  // Return a clone
}
```

**2. Server Method to Update CAs**

```go
func (gServer *GRPCServer) SetClientRootCAs(clientRoots [][]byte) error {
    // Build CertPool ONCE when CAs change
    certPool := x509.NewCertPool()
    for _, clientRoot := range clientRoots {
        certPool.AppendCertsFromPEM(clientRoot)
    }
    
    // Push update to server
    gServer.tls.SetClientCAs(certPool)
    return nil
}
```

**3. Handshake Uses Pre-Built Config**

```go
func (sc *serverCreds) ServerHandshake(rawConn net.Conn) (net.Conn, credentials.AuthInfo, error) {
    // Clone the current config (includes current ClientCAs)
    serverConfig := sc.serverConfig.Config()  // ← Read lock, returns clone
    
    // Wrap the raw TCP connection with TLS using the cloned config
    conn := tls.Server(rawConn, &serverConfig)
    
    // Perform the TLS handshake
    if err := conn.Handshake(); err != nil {
        return nil, nil, err
    }
    return conn, credentials.TLSInfo{State: conn.ConnectionState()}, nil
}
```

### Understanding the TLS Handshake Flow

**What Actually Happens:**

1. **Client Connects**: You get a `rawConn` (plain TCP connection)
2. **Clone Config**: Take a snapshot of current TLS settings (including current CAs)
3. **Wrap Connection**: `tls.Server()` wraps the TCP pipe with a "TLS layer"
4. **Handshake**: The TLS layer negotiates encryption and validates certificates
5. **Return**: The wrapped connection is returned - all data is now encrypted

### Fabric's Architecture: Push Model

```
┌─────────────────┐
│  Service Layer  │
│  (Detects CA    │
│   changes)      │
└────────┬────────┘
         │ Push update
         │ server.SetClientCAs()
         ▼
┌─────────────────┐
│  Server Layer   │
│  (Stores config │
│   with CAs)     │
└────────┬────────┘
         │ Clone config
         │ on each handshake
         ▼
┌─────────────────┐
│  TLS Handshake  │
│  (Uses cloned   │
│   config)       │
└─────────────────┘
```

**Performance:**
- **Per CA Update**: Build CertPool once, update server config
- **Per Handshake**: Clone config (cheap), no CertPool building

---

### Current Architecture: Hybrid Approach

We use **`GetConfigForClient` callback** - a Go standard library feature designed specifically for dynamic TLS configuration.

#### Why GetConfigForClient?

`GetConfigForClient` is Go's idiomatic way to handle dynamic TLS configuration. It's called automatically during each TLS handshake and allows you to:

1. **Inspect the client's request** (via `ClientHelloInfo`)
2. **Return a custom `tls.Config`** tailored for that specific connection
3. **Access the request context** for service-specific logic

#### Our Implementation Pattern

```go
func (m *TLSMaterials) CreateDynamicServerTLSConfig(
    getDynamicFunc func(ctx context.Context) *tls.Config,
) (*tls.Config, error) {
    tlsConfig, err := m.CreateServerTLSConfig()
    if err != nil {
        return nil, err
    }

    tlsConfig.GetConfigForClient = func(chi *tls.ClientHelloInfo) (*tls.Config, error) {
        // Service returns pre-built config with current CAs
        cfg := getDynamicFunc(chi.Context())
        
        if cfg == nil {
            return tlsConfig, nil  // Fallback to base config
        }
        
        return cfg, nil  // Return service's pre-built config
    }
    return tlsConfig, nil
}
```

### Service-Level Implementation

**Services store complete `tls.Config`** (not just CertPool):

```go
type Service struct {
    tlsConfig     atomic.Pointer[*tls.Config]  // Complete pre-built config
    staticCACerts [][]byte                      // Static CAs from YAML
    // ... other fields
}

// Return pre-built config (atomic read - very fast)
func (s *Service) GetDynamicTLSConfig(ctx context.Context) *tls.Config {
    if v := s.tlsConfig.Load(); v != nil {
        return *v
    }
    return nil
}

// Update config when CAs change
func (s *Service) updateCACertificates(dynamicCAs [][]byte) {
    // Merge static + dynamic CAs
    mergedCAs := connection.MergeCACerts(s.staticCACerts, dynamicCAs)
    certPool, _ := connection.BuildCertPoolFromBytes(mergedCAs)
    
    // Clone current config and update ClientCAs
    currentConfig := s.tlsConfig.Load()
    newConfig := (*currentConfig).Clone()
    newConfig.ClientCAs = certPool
    
    // Store atomically
    s.tlsConfig.Store(&newConfig)
}
```

### Sidecar: Push Architecture (Event-Driven)

The sidecar service uses a **push model** - it receives config blocks from the orderer and immediately updates its TLS configuration:

```
┌──────────────┐
│   Orderer    │
│              │
└──────┬───────┘
       │ Config block arrives
       ▼
┌──────────────────┐
│ Sidecar Service  │
│ - Extract CAs    │
│ - Merge with     │
│   static CAs     │
│ - Build CertPool │
│ - Update config  │
└──────┬───────────┘
       │ Pre-built config stored
       │ (atomic.Pointer)
       ▼
┌──────────────────┐
│ TLS Handshake    │
│ - Callback fires │
│ - Atomic read    │
│ - Return config  │
└──────────────────┘
```

**Characteristics:**
- ✅ Event-driven (no polling)
- ✅ Immediate updates when config blocks arrive
- ✅ Zero overhead when no config changes
- ✅ True push architecture

### Query Service: Pull Architecture (Polling)

The query service uses a **pull model** with intelligent caching:

```
┌──────────────────┐
│   Database       │
│ (Config blocks)  │
└──────┬───────────┘
       │ Poll on client connection
       │ (if cache expired)
       ▼
┌──────────────────┐
│ Query Service    │
│ - Check cache    │
│ - Fetch if stale │
│ - Extract CAs    │
│ - Merge & build  │
│ - Update config  │
└──────┬───────────┘
       │ Pre-built config stored
       │ (atomic.Pointer)
       ▼
┌──────────────────┐
│ TLS Handshake    │
│ - Callback fires │
│ - Trigger refresh│
│   (if needed)    │
│ - Atomic read    │
│ - Return config  │
└──────────────────┘
```

**Characteristics:**
- Polling-based (checks DB periodically)
- Lazy loading (only when clients connect)
- Double-check locking (prevents redundant queries)

**Why Polling?**

The query service has no event mechanism for config block updates. Options:

1. **Background goroutine** (continuous polling) - Wastes resources when idle
2. **Lazy loading** (current) - Only queries DB when clients actually connect
3. **Database LISTEN/NOTIFY** - Requires YugabyteDB 2025.2+ (not yet available)

---

### 1. GetConfigForClient vs ServerHandshake

| Feature | GetConfigForClient | ServerHandshake |
|---------|-------------------|-----------------|
| **Control** | Automatic (Go calls your function) | Explicit (you control timing) |
| **Use Case** | Dynamic cert rotation, SNI routing | gRPC custom credentials |
| **Flexibility** | Access to ClientHelloInfo | Full control over handshake |
| **Concurrency** | Safe by design | Manual locking required |
| **Idiomatic** | Yes (standard Go pattern) | Only for gRPC/custom needs |

**When to use GetConfigForClient:**
- ✅ Standard HTTPS/TCP servers
- ✅ Dynamic certificate rotation
- ✅ SNI-based routing
- ✅ Clean, built-in solution

**When to use ServerHandshake:**
- ✅ Implementing gRPC `TransportCredentials`
- ✅ Custom TLS logic before gRPC takes over
- ✅ Need to inspect connection metadata

**Our Choice**: We use `GetConfigForClient` because:
1. It's the idiomatic Go way
2. Provides access to connection context
3. Handles concurrency automatically
4. Cleaner than custom ServerHandshake

### Performance Analysis

| Operation | Fabric (Push) | Our Implementation (GetConfigForClient) |
|-----------|---------------|----------------------------------------|
| **Per CA Update** | Build CertPool once | Build CertPool once, update config |
| **Per Handshake** | Clone config | Atomic read + return config |
| **Memory Allocation** | One clone per handshake | One atomic read per handshake |
| **Lock Contention** | RWMutex read lock | Lock-free atomic read |
| **Complexity** | Medium | Low |

**Performance Verdict**: Both approaches are highly efficient. Our implementation has slightly less overhead (no RWMutex) but the difference is negligible in practice.

### Architectural Differences

| Aspect               | Fabric                            | Our Implementation                    |
| ----------------------| -----------------------------------| ---------------------------------------|
| **Pattern**          | Push (service → server)           | Push (service → server)               |
| **Update Mechanism** | Service calls `SetClientCAs()`    | Service updates `atomic.Pointer`      |
| **Read Mechanism**   | `ServerHandshake()` clones config | `GetConfigForClient()` returns config |
| **Synchronization**  | RWMutex (read lock per handshake) | Lock-free atomic read                 |
| **Flexibility**      | Fixed at server creation          | Dynamic per connection (via context)  |
| **Context Access**   | No                                | Yes (via ClientHelloInfo)             |
| **Code Complexity**  | Higher (custom credentials)       | Lower (standard Go pattern)           |

**Key Insight**: Both implementations follow the same fundamental pattern - the service updates the config when CAs change, and the handshake mechanism reads it. The main differences are in the synchronization mechanism (RWMutex vs atomic operations) and the Go API used (custom `ServerHandshake` vs standard `GetConfigForClient`).

---

### Option 1: Current Implementation ✅ (Recommended)

**Description**: Use `GetConfigForClient` with service-managed `tls.Config`

**Architecture**:
- Services store complete pre-built `tls.Config`
- Callback returns service's current config
- Sidecar: Event-driven (push from orderer)
- Query: Lazy loading with caching

**Advantages**:
- ✅ Idiomatic Go pattern
- ✅ Lock-free reads (atomic operations)
- ✅ Access to connection context
- ✅ Sidecar is truly event-driven (no polling)
- ✅ Query service only queries DB when needed
- ✅ Simple, maintainable code

**Disadvantages**:
- ⚠️ Query service still polls database
- ⚠️ Slight delay in query service picking up new CAs

**Status**: **Implemented and working**

---

### Option 2: Fabric-Style Push Model

**Description**: Adopt Fabric's exact pattern with `ServerHandshake`

**Architecture**:
- Server exposes `SetClientRootCAs()` method
- Services call this when CAs change
- Query service needs background goroutine for polling

**Advantages**:
- ✅ Proven pattern (used in Fabric)
- ✅ Clear separation of concerns
- ✅ Zero per-connection overhead

**Disadvantages**:
- ❌ Query service requires continuous background polling
- ❌ More complex (custom gRPC credentials)
- ❌ No access to connection context
- ❌ Wastes resources when idle

**Verdict**: Not recommended - adds complexity without significant benefit

---

### Option 3: Database LISTEN/NOTIFY

**Description**: Use PostgreSQL/YugabyteDB native pub/sub for config updates

**Architecture**:
```sql
-- Query service
LISTEN config_updates;

-- Sidecar (after writing config block)
NOTIFY config_updates;
```

**Advantages**:
- ✅ Event-driven for all services
- ✅ No polling overhead
- ✅ Database-native solution
- ✅ Guaranteed consistency

**Disadvantages**:
- ❌ Requires YugabyteDB 2025.2+ (not yet available)
- ❌ Database-specific (vendor lock-in)
- ❌ Requires fallback for older versions

**Verdict**: Ideal future solution, but blocked on database version

---

### Option 4: Dedicated Configuration Service

**Description**: New microservice monitors configuration changes and notifies other services via gRPC streams.
**Architecture**:
```
┌─────────────────┐
│ Config Service  │
│ - Monitors DB   │
│ - Publishes     │
│   updates       │
└────────┬────────┘
         │ gRPC stream
         ├──────────────┬──────────────┐
         ▼              ▼              ▼
    ┌─────────┐   ┌─────────┐   ┌─────────┐
    │ Sidecar │   │  Query  │   │ Service │
    │ Service │   │ Service │   │    N    │
    └─────────┘   └─────────┘   └─────────┘
```

**Advantages**:
- ✅ Centralized configuration management
- ✅ Event-driven architecture for all services
- ✅ Extensible for future configuration updates
- ✅ Uniform notification mechanism

**Disadvantages**:
- ❌ Additional service to deploy and maintain
- ❌ Increased operational complexity
- ❌ Additional network hop for configuration updates
- ❌ Over-engineered for single use case
- ❌ Single point of failure requiring high availability

**Verdict**: Not recommended - adds significant operational overhead for minimal benefit

---

### Option 5: Query Service Subscribes to Notification Service

**Description**: Query service subscribes to sidecar's existing notification service for configuration updates

**Architecture**:
```
┌─────────────────┐
│ Sidecar Service │
│ - Receives      │
│   config blocks │
│ - Notifies      │
│   subscribers   │
└────────┬────────┘
         │ Notification
         ▼
┌─────────────────┐
│ Query Service   │
│ - Subscribes    │
│ - Updates CAs   │
└─────────────────┘
```

**Advantages**:
- ✅ Event-driven for query service
- ✅ Reuses existing notification infrastructure
- ✅ No new services required

**Disadvantages**:
- ❌ Tight coupling between query and sidecar services
- ❌ Notification timing may not guarantee config block commitment
- ❌ Notification service not designed for configuration distribution
- ❌ Query service depends on sidecar availability
- ❌ Violates service independence principle

**Verdict**: Not recommended - creates undesirable service coupling

---

### Option 6: Sidecar Reads from Query Service Database

**Description**: Sidecar connects to query service and reads configuration from database

**Architecture**:
```
┌─────────────────┐
│ Sidecar Service │
│ - Writes to DB  │
│ - Reads from DB │ ← Circular!
└────────┬────────┘
         │
         ▼
┌─────────────────┐
│    Database     │
│ (Config blocks) │
└─────────────────┘
```

**Disadvantages**:
- ❌ Circular dependency (sidecar writes to DB, then reads from it)
- ❌ Doesn't address query service's polling requirement
- ❌ Violates separation of concerns
- ❌ Adds latency to sidecar's configuration updates
- ❌ Sidecar already has config blocks from orderer

**Verdict**: Not recommended - creates circular dependency and adds unnecessary complexity

---

### Current State: Production Ready ✅

Our current implementation is **production-ready** and follows Go best practices:

1. **Sidecar Service**: True push architecture (event-driven)
2. **Query Service**: Intelligent lazy loading with caching
3. **TLS Layer**: Idiomatic `GetConfigForClient` pattern
4. **Performance**: Lock-free reads, minimal overhead
5. **Maintainability**: Simple, clean code

### Why This Works

**Sidecar (Event-Driven)**:
- Receives config blocks from orderer immediately
- Updates TLS config as soon as CAs change
- Zero polling overhead
- Perfect push architecture

**Query Service (Lazy Loading)**:
- Only queries database when clients actually connect
- Caches results to avoid redundant queries
- Double-check locking prevents concurrent DB hits
- Minimal resource usage when idle

**Performance**:
- Both services: Pre-build config once per CA update
- Both services: Atomic read on every handshake (nanoseconds)
- No expensive operations on hot path

### Future Optimization: Database LISTEN/NOTIFY

**When Available**: Migrate query service to database notifications

**Benefits**:
- Eliminates polling entirely
- Event-driven for all services
- Uniform architecture

**Timeline**: When YugabyteDB 2025.2+ is available

---

## Conclusion

Our implementation successfully combines:
- **Fabric's core insight**: Build CertPool once, reuse many times
- **Go's idiomatic pattern**: `GetConfigForClient` for dynamic configuration
- **Practical architecture**: Event-driven where possible, lazy loading where necessary

The result is a production-ready solution that:
- ✅ Enables CA rotation without service restart
- ✅ Maintains high performance (lock-free reads)
- ✅ Uses simple, maintainable code
- ✅ Follows Go best practices
- ✅ Provides clear path for future optimization

**Status**: Ready for production deployment

dean-amar:
#### Type of change

- New feature
- Test update
- Documentation update
 
#### Description

This PR introduces dynamic root CA certificate rotation for the sidecar and query services, enabling TLS client certificate validation to pick up new organization CAs from config blocks without a service restart.

#### Additional details

How the dynamic CA merge works:
* Static CAs (from YAML ca-cert-paths) are always trusted. Dynamic CAs (from config blocks) are appended on top. This means existing clients with YAML-configured CAs continue to work, and new org clients become trusted as soon as their CA appears in a config block.
* Mutual TLS only:
CreateDynamicServerTLSConfig only installs the GetConfigForClient callback in mtls mode. For TLS and non-TLS modes it returns the standard static config unchanged.

#### Related issues

* Resolved #290

#### Following PRs: 
The following improvements are deferred to subsequent PRs:
* Add metrics for CA refresh - track successful refreshes, failures, and the number of active dynamic CAs. This gives operators visibility into CA rotation health and allows alerting on persistent DB fetch failures.


service/query/query_service.go
```diff
@@ -114,6 +135,74 @@ func (q *Service) RegisterService(server *grpc.Server) {
...
+
+	if lastFetch == 0 || (now-lastFetch) > q.config.CAFetchInterval.Nanoseconds() {
+		q.refreshDynamicRootCAs(ctx)
+	}
+
+	if v := q.dynamicRootCAs.Load(); v != nil {
+		return *v
+	}
+	return nil
+}
```
liran-funaro: This method should use a lock, not atomic operations.
Otherwise, multiple go-routines will attempt to update the root-CAs concurrently.
Please revert to a lock based implementation, and use `time.Time` instead of `atomic.Int64` to represent time.

service/query/query_service.go
```diff
@@ -114,6 +135,74 @@ func (q *Service) RegisterService(server *grpc.Server) {
...
+		return
+	}
+
+	logger.Debug("Refreshing dynamic root CAs from config transaction")
+
+	// TODO: Add version optimization, We don't need to read the config-block entirely, only the version.
+	ctx, cancel := context.WithTimeout(ctx, caFetchTimeout)
+	defer cancel()
+
+	configTx, err := q.GetConfigTransaction(ctx, nil)
```
liran-funaro: `q.GetConfigTransaction(ctx, nil)` wraps the error with grpc error. It is meant for clients.
Please use the internal method instead.

utils/dynamictls/tls.go
```diff
@@ -0,0 +1,49 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package dynamictls
```
liran-funaro: This should be part of the `connection` package.

liran-funaro (commented):
Quick review

service/query/query_service.go
dean-amar: Done.

service/query/query_service.go
```diff
@@ -58,16 +65,35 @@ type (
...
+		// Dynamic CA management fields: dynamicRootCAs provides atomic, restart-free certificate
+		// rotation; lastCAFetch tracks the last refresh to a throttle database load; and
+		// refreshMutex implements a double-check locking pattern to prevent redundant,
+		// concurrent refresh operations during connection spikes.
+		//
+		// Design note: We use atomic operations for reads and a mutex only for writings.
+		// This lock-free read pattern ensures clients read are not serialized,
+		// while the mutex in refreshDynamicRootCAs
+		// guarantees only one goroutine performs the expensive DB fetch at a time and writes it.
+		dynamicRootCAs atomic.Pointer[[][]byte]
```
cendhu: Please remove the dynamic prefix from all files. In Fabric, the root CA is provided in the peer configuration primarily to ensure the initial peer join channel --block block1 command succeeds for the organization. Since we have already transitioned to using the config block path in the sidecar configuration, the CAs within the configuration block serve as the definitive source of truth. Terming the YAML-based CAs as 'static' and the block-based CAs as 'dynamic' feels awkward; If you would like to differentiate them, let's use the term `RootCAsInConfig` instead.

utils/connection/server_util.go
```diff
@@ -48,46 +61,71 @@ func NewLocalHostServer(creds TLSConfig) *ServerConfig {
...
-	}
-	if c.KeepAlive != nil && c.KeepAlive.EnforcementPolicy != nil {
-		opts = append(opts, grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{
-			MinTime:             c.KeepAlive.EnforcementPolicy.MinTime,
-			PermitWithoutStream: c.KeepAlive.EnforcementPolicy.PermitWithoutStream,
-		}))
+// StartDynamicService runs a service with dynamic CA certificate support which
+// enables certificate rotation without a service restart.
+// Note: Only supports single server configuration. Use StartService for multiple servers.
+func StartDynamicService(
```
cendhu: I do not understand why we have StartService and StartDynamicService which differ in only one argument.  Rest of the code look duplicate.

cendhu (commented):
The current design has the gRPC server calling back into the service (`GetDynamicRootCAs`) on every TLS handshake via `GetConfigForClient`. This inverts the natural dependency — the server now depends on the service, creating a circular relationship. It also introduces per-handshake overhead (new `tls.Config` + `x509.CertPool` allocation on every connection), and requires new abstractions (`DynamicService` interface, `DynamicGrpcServer`, `StartDynamicService`, `RunGrpcDynamicServer`) that duplicate the non-dynamic paths.

A cleaner approach is to flip the direction: have the service push CA updates to the server when it detects changes, rather than the server pulling on every handshake. This is the pattern Fabric uses in its peer — the service that learns about new CAs calls `server.SetClientRootCAs(certs)`, and the server just uses whatever it was last told to trust. The CertPool is built once per CA change, not once per connection.

This push model is better for long-term maintainability and readability because:
  - The dependency flows naturally: service → server, not server → service
  - No need for `DynamicService` interface or separate dynamic server creation paths — the server is just a server with a `SetClientRootCAs` method
  - The refresh/polling logic stays entirely in the service, and the server remains a simple TLS endpoint

In terms of implementation, you don't need to reimplement gRPC's `TransportCredentials` like Fabric does. A simpler approach would work: wrap the `tls.Config` with a mutex-protected `ClientCAs` field, and have `GetConfigForClient` just return the current config (a read lock, no allocations). The service calls `SetClientRootCAs` which builds the `CertPool` once and swaps it under a write lock. 

Fabric references: 
- https://github.com/hyperledger/fabric/blob/main/internal/pkg/comm/creds.go
- https://github.com/hyperledger/fabric/blob/main/internal/pkg/comm/server.go

cendhu (commented):
I haven't reviewed the implementation in detail yet, but I have gone through the high-level design. I have two major concerns:

1. **Interface Bloat:** We have a method in the interface that only two out of seven implementers actually use; the others are simply providing empty implementations to satisfy the compiler. Since only a small minority requires this method, it suggests the interface is not designed correctly.
2. **Function as Arguments:** We generally avoid passing functions as arguments or storing them in structs to maintain readability. This PR introduces functional arguments to four methods, which I don't believe is necessary for this feature.

Since my suggestions involve significant changes, I will submit a separate PR demonstrating the alternative design I have in mind.

liran-funaro (commented):
Services should NOT load the TLS config and should not create one.
They should only fetch a list of CAs and let the server part handle the TLS.
There should be a clear separation of concerns in the code.
Please modify the PR accordignly.

PR #404, opened 2026-03-03
upgrade go to 1.26 and apply go fix

cendhu:
#### Type of change

#### Description

go fix is a new tool that get rid of many tech debt.


service/vc/preparer_test.go
```diff
@@ -729,6 +728,8 @@ func BenchmarkPrepare(b *testing.B) {
 }
 
 // invalidStatus is a oneliner to create a pointer to the given status.
+//
+//go:fix inline
 func invalidStatus(status committerpb.Status) *committerpb.Status {
-	return &status
+	return new(status)
 }
```
liran-funaro: Can we remove this method now that `new()` works on literals?

PR #416, opened 2026-03-08
[sidecar] newBlockDelivery and newBlockQuery accept blockStore

liran-funaro:
#### Type of change

#### Description

- Align `newBlockDelivery()` and `newBlockQuery()` to accept only `blockStore`. There is no need for `channelID` as it is included in `blockStore`.


cendhu (approved):
LGTM.

PR #425, opened 2026-03-10
[sidecar] Remove channel ID from committer ledger

### LINKED ISSUE #418: Channel ID Mismatch Between Sidecar Ledger and State Database
## Problem Statement

The current implementation creates an inconsistency in how channel IDs are handled between the sidecar's ledger and the state database:

**Current Behaviour:**
- The sidecar enforces channel ID validation in its ledger layer
- The ledger is initialized using the Orderer's channel ID
- Delivery requests with mismatched channel IDs are rejected
- **However**, the state database does NOT track or enforce channel IDs

**Impact:**
This inconsistency can lead to data corruption scenarios. For example:
If the sidecar is restarted with the same underlying storage but connected to an orderer from a different channel, the ledger will correctly connect to the appropriate channel-specific ledger.
But the state DB will contain mixed data from multiple channels without any way to distinguish between them.
The ledger may appear empty while the state DB is populated with data from a different channel

## Proposed Solutions

We need to choose one of two approaches to resolve this inconsistency:

### Option 1: Remove Channel ID Enforcement (Simplify)
- Use a fixed channel ID when creating the ledger
- Remove channel ID validation for delivery requests
- Treat the system as single-channel only

**Pros:**
- Simpler implementation
- Reduces validation overhead
- Clear single-channel semantics

**Cons:**
- Cannot support multiple channels with the same storage
- Less protection against misconfiguration

### Option 2: Add Channel ID to State Database (Multi-Channel Support)
- Track channel ID in the state database schema
- Implement channel isolation (e.g., using channel ID as a prefix for table names or as a column in all tables)
- Ensure both ledger and state DB enforce channel boundaries consistently

**Pros:**
- Enables true multi-channel support
- Consistent channel enforcement across all layers
- Better isolation and data integrity

**Cons:**
- More complex implementation
- Requires database schema changes
- May impact performance (additional filtering/indexing)

## Recommendation

Consider the project's roadmap regarding multi-channel support requirements.

## Additional Context

- This issue affects data integrity and system reliability
- Should be addressed before production deployment in multi-channel environments

liran-funaro:
#### Type of change

#### Description

- Remove channel ID from committer ledger
- Move `services/sidecar/sidecarclient` to `utils/delivercommitter`
- Replace the "client" notation with `ToChannel()` method that delivers to a channel

#### Related issues

- resolves #418

### DIFF
--- a/utils/delivercommitter/committer.go
+++ b/utils/delivercommitter/committer.go
@@ -22,23 +22,23 @@ import (
 
 // Parameters needed for deliver to run.
 type Parameters struct {
-	Client       *connection.ClientConfig
+	ClientConfig *connection.ClientConfig
 	NextBlockNum uint64
 	OutputBlock  chan<- *common.Block
 }
 
-// ToChannel start receiving blocks starting from NextBlockNum to outputBlock.
-func ToChannel(ctx context.Context, cdp Parameters) error {
+// ToQueue start receiving blocks starting from NextBlockNum to outputBlock.
+func ToQueue(ctx context.Context, cdp Parameters) error {
 	cm, err := ordererconn.NewConnectionManager(&ordererconn.Config{
-		TLS:   ordererconn.TLSConfigToOrdererTLSConfig(cdp.Client.TLS),
-		Retry: cdp.Client.Retry,
+		TLS:   ordererconn.TLSConfigToOrdererTLSConfig(cdp.ClientConfig.TLS),
+		Retry: cdp.ClientConfig.Retry,
 		Organizations: map[string]*ordererconn.OrganizationConfig{
 			"org": {
 				Endpoints: []*commontypes.OrdererEndpoint{{
-					Host: cdp.Client.Endpoint.Host,
-					Port: cdp.Client.Endpoint.Port,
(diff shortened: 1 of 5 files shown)


utils/delivercommitter/committer.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Parameters needed for deliver to run.
+type Parameters struct {
+	Client       *connection.ClientConfig
+	NextBlockNum uint64
+	OutputBlock  chan<- *common.Block
+}
+
+// ToChannel start receiving blocks starting from NextBlockNum to outputBlock.
+func ToChannel(ctx context.Context, cdp Parameters) error {
```
cendhu: This `ToChannel` confused me initially with channel ID. We could name it something like `delivercommitter.ToQueue` or `delivercommitter.Deliver` or something similar.

cendhu (approved):
LGTM. A few suggestions on naming.

utils/delivercommitter/committer.go
liran-funaro: `delivercommitter.ToQueue` can work fine. `delivercommitter.Deliver` repeats the word "deliver".

PR #427, opened 2026-03-10
[bft-delivery] BFT delivery

### LINKED ISSUE #272: [deliver] Verify block prev-hash

### LINKED ISSUE #273: [deliver] Verify block's singatures

### LINKED ISSUE #274: [deliver] Monitor for block withholding attack

### LINKED ISSUE #295: [deliver] Increase test-coverage

### LINKED ISSUE #361: [sidecar] Load last config block from ledger
When working on the BFT delivery, We need to load the last block from the ledger to continue validation from there.
But we also need the last config block of this block because the one in the stateDB might be ahead of the last committed block in the ledger.
Since I already load the config block from the ledger, we don't need to also load it from the stateDB.

### LINKED ISSUE #396: [mock] A unified mock orderer test environment
All tests require common pattern with using the mock orderer.
1. Create multiple servers per ID
1. Create a genesis block
2. Initialising the appropriate TLS credentials
3. Starting/stopping servers per ID.

### LINKED ISSUE #397: [ordererconn] Always use a config/genesis block
Block delivery/broadcast, requires information that is available only via the config block.
Thus, it makes sense to require a config block for the configuration, instead of manually inputting the data to the YAML.

liran-funaro:
#### Type of change

- New feature
- Breaking change
 
#### Description

**[BFT block delivery]**
- Introduce BFT (Byzantine Fault Tolerant) block delivery with block withholding detection
- Use only the config block as the initialization mechanism (no separate bootstrap config)
- Split the delivery package into 3:
  - `deliver` (core library) - provides `ToQueue()` method for streaming blocks to Go channels
  - `deliverorderer` (BFT delivery from orderer) - implements fault-tolerant delivery with verification and block withholding detection
  - `delivercommitter` (delivery from sidecar/ledger) - implements delivery from peer ledger service (instead of `sidecarclient`)
- Replace the "client" notation with `ToQueue()` method that delivers to a channel
- Move the `NewBroadcastStream()` method to loadgen package (`loadgen/adapters/broadcast.go`) as it is the main use case for it

**[Test ENV]**
- Introduce a common `OrdererTestEnv` in `mock/test_exports.go` that prepares the orderer and artifacts for tests
- Use it everywhere the Orderer is used for testing
- `MockOrderer` now receives `ArtifactsPath` instead of `ConfigBlockPath` to support signing the blocks

**[ordererconn]**
- `LoadConfigBlock()/LoadConfigBlockFromFile()` now returns `ConfigBlockMaterial` which includes:
  - Channel ID
  - Config block
  - Bundle
  - Orderer organizations with endpoints and CAs
  - Application organizations with CAs
- Remove connection manager (`conn_manager.go`) and replace with `ConnectionMaterial` struct
- Add `OrdererConnectionMaterial()` method which returns `connection.ClientMaterial` instances to connect to the orderer
- Add comprehensive tests: `config_test.go`, `identity_test.go`, `load_test.go`

**[connection]**
- Introduce `ClientMaterial` struct for connection configuration without YAML config:
  - Contains `Endpoints`, `TLS`, and `Retry` fields
  - Factory method `NewClientMaterial()` from `MultiClientConfig`
- Add context parameter to `PreAllocateListener()` (in `server_util.go`)
- Add `ClosePreAllocatedListener()` method (in `server_util.go`)

**[sidecar]**
- Load config block from ledger instead of the state DB
- Remove config-updater (not needed as the delivery updates automatically)
- Remove `bootstrap` config field - it is now part of the Orderer config

**[coordinator]**
- Remove `GetConfigTransaction()` and `SetConfigTransaction()` methods from coordinator service (not used anymore)

**[Additional Changes]**
- Add mock streamer interface for testing (`utils/deliver/mocks/streamer.mock.go`)
- Update integration tests to use new delivery mechanism
- Update loadgen adapters to work with new delivery system
- Add comprehensive test coverage for new packages
- Update CI workflow and Makefile to support mocks

For reference, here is the output of the verification benchmark:
```
go test ./utils/deliverorderer/... -bench "Benchmark.*" -run "^$" | awk -f scripts/bench-tx-per-sec.awk
goos: linux
goarch: amd64
pkg: github.com/hyperledger/fabric-x-committer/utils/deliverorderer
cpu: Intel(R) Core(TM) i9-14900K
BenchmarkVerifyBlock/blockSize=100-32         	  169263	      6927 ns/op	144,363 TX/sec
BenchmarkVerifyBlock/blockSize=1,000-32       	 1490384	       866.4 ns/op	1,154,201 TX/sec
BenchmarkVerifyBlock/blockSize=10,000-32      	 6001669	       202.2 ns/op	4,945,598 TX/sec
BenchmarkVerifyBlock/blockSize=100,000-32     	 9214441	       173.0 ns/op	5,780,347 TX/sec
PASS
ok  	github.com/hyperledger/fabric-x-committer/utils/deliverorderer	41.489s
```

#### Related issues

- resolves #295 
- resolves #396 
- resolves #397 
- resolves #361 
- resolves #272 
- resolves #273 
- resolves #274


utils/deliverorderer/verify.go
```diff
@@ -0,0 +1,234 @@
...
+
+	"github.com/hyperledger/fabric-x-committer/utils/deliver"
+	"github.com/hyperledger/fabric-x-committer/utils/ordererconn"
+)
+
+type (
+	// blockProcessingState maintains the state required for processing and verifying blocks
+	// in sequence. It tracks the expected next block number, the last processed block,
+	// and configuration for content verification.
+	blockProcessingState struct {
```
cendhu: I think, it would be good to have a `newXXX` methods for these objects. Otherwise, it is unclear which fields are set by the user/creator and which fields are modified internally for the verification flow.

utils/deliverorderer/verify.go
```diff
@@ -0,0 +1,234 @@
...
+// verifyHashes checks that the block previous hash matches the previous block,
+// and that the internal data hash matches the one in the header.
+// It is only used internally.
+func (s *blockProcessingState) verifyHashes(block *common.Block) error {
+	blockNumber := block.Header.Number
+
+	// Verify header hash if we have the previous block hash.
+	if len(s.lastBlockHeaderHash) != 0 {
+		if !bytes.Equal(block.Header.PreviousHash, s.lastBlockHeaderHash) {
+			return errors.Errorf("previous block header hash mismatch on block [%d]", blockNumber)
```
cendhu:
```suggestion
			return errors.Newf("previous block header hash mismatch on block [%d]", blockNumber)
```
to be consistent with other usages.

utils/deliverorderer/verify.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// Block's data can be nil only if we only requested headers.
+	// However, the block's data can be non-nil even if we only requested headers in case of config block updates.
+	// Thus, we always verify the data hash if data is present.
+	if block.Data != nil {
+		dataHash := protoutil.ComputeBlockDataHash(block.Data)
+		// Verify that Header.DataHash is equal to the hash of block.Data
+		// This is to ensure that the header is consistent with the data carried by this block
+		if !bytes.Equal(dataHash, block.Header.DataHash) {
+			return errors.Errorf("block data hash mismatch on block [%d]", blockNumber)
```
cendhu:
```suggestion
			return errors.Newf("block data hash mismatch on block [%d]", blockNumber)
```
to be consistent with other usages. Please fix all such occurrences.

cendhu (commented):
Thanks for all the inline comments. It helped in the review process. 

I have completed review for deliverorderer/verify.go and deliverorderer/verify_test.go. Once we finalize on these files, I will move on to remaining files in the `deliverorderer`.  Not possible to do all at once as these two files itself took two and half hours for me.

utils/deliverorderer/verify.go
cendhu: Yes, `blockVerificationStateMachine` sounds good as each verification updates the internal states.

utils/deliverorderer/verify.go
```diff
@@ -0,0 +1,234 @@
...
+func (s *blockProcessingState) verifyBlockPolicy(block *common.Block) error {
+	if s.ConfigBlockMaterial == nil || s.verifierFunc == nil {
+		return nil
+	}
+
+	blockNumber := block.Header.Number
+
+	// If the config block is ahead of this block, something wrong has happened.
+	// This should never happen as we validate in-order block processing.
+	// But if we received a config block from external source, such error may occur.
```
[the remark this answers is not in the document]
cendhu: > This part protects against corrupted ledger, or bad input YAML.

Thanks for your patience. It is clear now. I think it is better to mention this in the comment so that in the future we know why we added this check.

utils/deliverorderer/orderer.go
```diff
@@ -0,0 +1,603 @@
...
+		// The buffer size is bounded by MaxBlocksAhead to prevent unbounded memory growth.
+		dataBlockDeadlineBuffer []time.Time
+
+		curDataBlockSourceID uint32
+		wg                   sync.WaitGroup
+		cancelWorkers        context.CancelFunc
+		activeWorkers        atomic.Int64
+		restartBackoff       *backoff.ExponentialBackOff
+		lastAction           streamAction
+		nextAllowedSuspicion time.Time
```
cendhu: Comments can help here. We can try to add comments for struct as much as possible.

utils/deliverorderer/orderer.go
```diff
@@ -0,0 +1,603 @@
...
+	//   - LastBlock: The most recently processed block (do NOT pass the genesis block here to allow the delivery to
+	//	   deliver it for processing)
+	//   - NextBlockVerificationConfig: Config block used to verify the next incoming block
+	//   - LastestKnownConfig: The newest known config block (can be ahead of NextBlockVerificationConfig)
+	//
+	// State Relationships:
+	//   - If LastestKnownConfig is not provided, NextBlockVerificationConfig is used as the latest config
+	//   - If NextBlockVerificationConfig is not provided, policy verification starts only when a config
+	//     block is delivered (LastestKnownConfig is NOT used for verification as it may not be relevant)
+	//   - LastestKnownConfig helps recover when delivery missed a config block that updated all endpoints
```
cendhu: There are two purposes:
``` 
  // - LastKnownConfigBlock enables connectivity when PreviousAppliedConfigBlock has stale orderer
  //   endpoints. This occurs in two scenarios:
  //   1. A new organization joining long after genesis, where the starting config block has
  //      completely outdated endpoints. The caller provides current endpoints obtained out-of-band
  //      (e.g., via configuration file).
  //   2. Session resumption after all old orderer endpoints were replaced while the data stream
  //      hadn't yet processed the config block with the new endpoints.
  //   This field is updated at the end of each delivery run for future runs
```

utils/deliverorderer/orderer.go
```diff
@@ -0,0 +1,603 @@
...
+	d.dataStream.dataBlockStream = true
+	return nil
+}
+
+func (d *ftDelivery) finalize() {
+	if d.cancelWorkers != nil {
+		// Cancel workers if any.
+		d.cancelWorkers()
+	}
+	d.wg.Wait()
```
cendhu: when wait() crosses function boundary, it generally becomes complex to navigate the code.

utils/deliverorderer/orderer.go
```diff
@@ -0,0 +1,603 @@
...
+}
+
+func (d *ftDelivery) run(ctx context.Context) error {
+	err := d.updateDeliveryStreams(ctx, streamActionConfigUpdate)
```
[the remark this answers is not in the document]
liran-funaro: >The verification state machine handles duplicates, but this is accidental complexity from the lifecycle model, not an inherent requirement.

It is an inherit requirement due to getting the same blocks from multiple parties.

Anyway, I'm fine with your suggestion. I will try to apply it. But it does require major changes to the code.

utils/deliverorderer/orderer.go
```diff
@@ -0,0 +1,603 @@
...
+			return errors.WithMessage(err, "error loading last block")
+		}
+	}
+
+	if session.NextBlockVerificationConfig != nil {
+		err := s.updateIfConfigBlock(session.NextBlockVerificationConfig)
+		if err != nil {
+			return errors.WithMessage(err, "error loading next block verification config")
+		}
+	}
```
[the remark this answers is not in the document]
liran-funaro: I went with `newBlockProcessingState(session *SessionInfo)` processes the session info and returns the state and latest config.
```go
func newBlockProcessingState(session *SessionInfo) (
	state blockVerificationStateMachine, latestConfig configState, err error,
)
```

utils/ordererconn/material.go
```diff
@@ -0,0 +1,83 @@
...
+	// MaterialParameters are the parameters to create connection material.
+	MaterialParameters struct {
+		TLS   connection.TLSMaterials
+		Retry *connection.RetryProfile
+		API   string
+	}
+)
+
+// OrdererConnectionMaterial returns the connection material given the config block material and the given parameters.
+func OrdererConnectionMaterial(m *channelconfig.ConfigBlockMaterial, p MaterialParameters) *ConnectionMaterial {
```
cendhu: `ordererconn.NewConfig`?

utils/ordererconn/material.go
```diff
@@ -0,0 +1,83 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package ordererconn
```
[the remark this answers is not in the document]
liran-funaro: I picked `ordererdial`

loadgen/adapters/broadcast.go
```diff
@@ -0,0 +1,198 @@
...
+
+	"github.com/hyperledger/fabric-x-committer/utils/connection"
+	"github.com/hyperledger/fabric-x-committer/utils/ordererdial"
+)
+
+type (
+	// BroadcastStream implements broadcast, with the addition of automatically signing the envelope.
+	BroadcastStream struct {
+		broadcaster
+	}
```
cendhu: Why do we need a struct with just an interface? Can't we replace `BroadcastStream` just with the interface. 

```
// CloseConnections closes all the connections for the stream and logs errors.
func (s *BroadcastStream) CloseConnections() {
	connection.CloseConnectionsLog(s)
}

// SendBatch sends a batch one by one.
func (s *BroadcastStream) SendBatch(envelopes []*common.Envelope) error {
	for _, env := range envelopes {
		err := s.Send(env)
		if err != nil {
			return err
		}
	}
	return nil
}
```

Very hard for me to navigate and understand. 

`Send()` implemented by both broadcaster is used only by the `SendBatch()`. Why can't we implement the `SendBatch()` directly on the broadcaster struct? Similarly, i think the close can be called directly. Additional indirection adds complexity. 

In addition, the interface should be defined at the receiver side -- runtime and orderer adapter.

loadgen/adapters/broadcast.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// bftBroadcaster supports broadcasting to multiple nodes (one per ID).
+	// It should be used only for testing as it does not validate successful broadcast to a valid quorum.
+	bftBroadcaster struct {
+		idToStream map[uint32]*cftBroadcaster
+	}
+)
+
+// NewBroadcastStream starts a new broadcast stream (non-production).
+func NewBroadcastStream(ctx context.Context, config *ordererdial.Config) (*BroadcastStream, error) {
```
cendhu: Return the actual struct and make the receiver receive it using interface.

cendhu (commented):
Final review.

loadgen/adapters/broadcast.go
liran-funaro: It seems that the linter no longer complains :)
I applied the changes.

PR #465, opened 2026-03-22
[connection] Upgrade to backoff/v5

liran-funaro:
#### Type of change

#### Description

- Upgrade to backoff/v5
- Introduce `ExecuteWithResult()` and `WaitForCondition()`

### DIFF
--- a/utils/retry/sustain.go
+++ b/utils/retry/sustain.go
@@ -54,6 +54,13 @@ func Sustain(ctx context.Context, p *Profile, op func() error) error {
 			return nil, errors.WithMessage(err, "sustained operation error")
 		})
 
+		// The above call should retry whenever ErrBackOff is returned.
+		// Thus, if the returned error is ErrBackOff, it means we reached the
+		// maximal attempts (MaxElapsedTime).
+		if errors.Is(err, ErrBackOff) {
+			return err
+		}
+
 		if errors.Is(err, ErrNonRetryable) {
 			return err
 		}
--- a/service/vc/database.go
+++ b/service/vc/database.go
@@ -174,15 +174,14 @@ func (db *database) queryVersionsIfPresent(ctx context.Context, nsID string, que
 	start := time.Now()
 	query := FmtNsID(queryVersionsSQLTempl, nsID)
 
-	foundKeysVersions, err := retryQueryAndReadTwoItems[[]byte, int64](ctx, db, query, queryKeys)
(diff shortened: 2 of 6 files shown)


utils/deliver/deliver_cft.go
```diff
@@ -84,45 +84,30 @@ func (c *CftClient) receiveFromBlockDeliverer(ctx context.Context, p *Parameters
...
+	status, err := c.deliverRelay(sCtx, stream, p)
+	if err != nil {
+		return err
 	}
-	return nil
+	if status == common.Status_SUCCESS {
+		// Indication that the seek range is fully delivered.
+		return nil
+	}
+	return errors.Newf("deliver failed with status: %s", status.String())
```
cendhu: ⚠️ **Critical verification needed**: The old code had a retry loop that would retry on non-SUCCESS status. The new code returns immediately on non-SUCCESS.

**Question:** Is there an outer retry loop that handles this? Please verify that delivery retry behavior is preserved at a higher level, or restore the retry loop here.

<details>
<summary>Code comparison</summary>

**Before:**
```go
for sCtx.Err() == nil {
    status, err := c.deliverRelay(sCtx, stream, p)
    if status == common.Status_SUCCESS {
        return nil
    }
    // Retry with backoff
}
```

**After:**
```go
status, err := c.deliverRelay(sCtx, stream, p)
if status == common.Status_SUCCESS {
    return nil
}
return errors.Newf("deliver failed with status: %s", status.String())
```
</details>

utils/retry/sustain.go
```diff
@@ -43,39 +40,24 @@ var (
...
-		}
-		if !errors.Is(opErr, ErrBackOff) {
-			b.Reset()
-		}
-		if err := WaitForNextBackOffDuration(ctx, b); err != nil {
+		_, err := executeWithResult(ctx, p, func() (any, error) {
+			err := op()
+			if err == nil {
+				err = errors.New("sustained operation ended unexpectedly")
+			}
```
cendhu: ✅ **Significant simplification**: The new implementation delegates backoff management to `executeWithResult()`, eliminating manual backoff state management.

**Why:** The old code manually managed `NextBackOff()` and timeout detection. The new code leverages backoff v5's built-in retry mechanism, making the logic clearer and less error-prone.

**Note:** Verify that timeout behavior matches v4 (MaxElapsedTime should be respected via functional options).

utils/deliver/deliver_cft.go
liran-funaro: This is no longer needed after we introduced the wait on the block storage implementation.

PR #515, opened 2026-04-01
[bench] Replace AWK tx/s script with b.ReportMetric

### LINKED ISSUE #514: [benchmark] Replace AWK-based tx/s reporting with b.ReportMetric in benchmarks
**Summary**

  Benchmark targets in the Makefile pipe output through scripts/bench-tx-per-sec.awk to add a TX/sec column. This
  can be done natively with Go's b.ReportMetric, which is more accurate for batch benchmarks and eliminates the
  external script dependency.

  **Changes**

  - Add test.ReportTxPerSecond(b) helper in utils/test/benchmark.go
  - Call it from all benchmarks across loadgen/, service/coordinator/, service/vc/, service/sidecar/,
  utils/deliverorderer/, and utils/serialization/
  - For batch benchmarks (dependency graph, preparer): cap input to exactly b.N transactions so the metric is
  accurate
  - Remove | awk -f scripts/bench-tx-per-sec.awk from all Makefile bench targets
  - Delete scripts/bench-tx-per-sec.awk

  **Why**

  - Accuracy: The AWK script computes 1e9/ns_per_op, which gives ops/sec — not tx/sec for batch benchmarks where one
   "op" processes multiple transactions. b.ReportMetric reports the true throughput.
  - Simplicity: No external script, no pipe — benchmark output includes tx/s natively.
  - Consistency: All benchmarks report the same metric from a single helper.

cendhu:
#### Type of change

#### Description

  Add test.ReportTxPerSecond(b) helper and call it from all benchmarks.
  Cap batch benchmarks to exactly b.N transactions for accurate reporting.
  Remove bench-tx-per-sec.awk and Makefile AWK pipes.

#### Related issues

 - resolves #514


service/coordinator/dependencygraph/manager_test.go
```diff
@@ -90,7 +90,7 @@ func BenchmarkDependencyGraph(b *testing.B) {
 								PrometheusMetricsProvider: monitoring.NewProvider(),
 							})
 
-							txPoll := workload.GenerateTransactions(b, p, max(b.N*3, batchSize*3))
+							txPoll := workload.GenerateTransactions(b, p, max(b.N, batchSize))
```
liran-funaro: This benchmark is designed such that the dependency graph internal queue will not be empty during the benchmark.
This is why we submit up to 3 times the number of TXs.

loadgen/metrics/tracker_test.go
liran-funaro: Unrelated: please remove `{ /* never */ }, {Portion: 1},` as they cannot be correctly measured (too simple).

loadgen/workload/stream_test.go
liran-funaro: Unrelated: please add `flogging.ActivateSpec("fatal")`

liran-funaro: Unrelated: please add `flogging.ActivateSpec("fatal")`

utils/serialization/envelope_wrapper_test.go
liran-funaro: unrelated: please add `flogging.ActivateSpec("fatal")`

liran-funaro (commented):
A major issue (personal preference) is that benchmark metrics in the standard output (like ns/op, B/op, or custom metrics) are not formatted with commas.
This makes the output difficult to read.
I suggest using the report metrics method as this PR suggest, but also having an optional post script to add the commas.

`BenchmarkMapBlockSize()` reports incorrectly because of the block size.
`b.N` indicated the number of iterations to test to ensure confidence in the results.
We usually use b.N as the number of TXs as the benchmark runs the same operation on each TX, which is what `b.N` indicates.
In `BenchmarkMapBlockSize()`, the block is the minimal tested unit as we test how the block size affects the performance.
Thus, we should use `b.N` as the number of blocks in this case.
However, I still think we should report TX/s as this is our actual unit of measurement we care about.
I suggest allowing `test.ReportTxPerSecond()` reporting TX/s with a different number of TXs than `b.N`.

Unrelated to this PR: it seems that the notifier and dependency graph benchmarks hangs. And the deliver benchmark have a error.
These should be fixed in a different PR.

liran-funaro (commented):
@cendhu Required your approval for my modifications.

liran-funaro (approved):

cendhu:
LGTM. Couldn't approve it via gh.

PR #516, opened 2026-04-01
feat(vc): add StateImporter for bulk state initialization from external sources

veyron-kairo:
Right now there's no way to initialize the committer's state database from an external source — state can only enter through the full block processing pipeline (orderer → sidecar → coordinator → vc). That pipeline does MVCC validation, endorsement verification, dependency tracking, etc., which makes no sense for trusted pre-validated state from a Fabric snapshot.

This PR adds a `StateImporter` to the VC service that lets you bulk-load state directly into the database using PostgreSQL's COPY protocol, skipping the validation pipeline entirely. It's meant to be the building block for the `committer --init-from-snapshot` bootstrap mode that the [Fabric → Fabric-X migration tool](https://github.com/LF-Decentralized-Trust-Mentorships/mentorship-program/issues/65) needs.

**What it does:**

- `InitSchema` / `CreateNamespace` — sets up system tables and per-namespace tables
- `ImportNamespaceState` — streams key-value-version triples into a namespace table via COPY (handles large datasets without loading everything into memory)
- `ImportPolicy` — writes namespace policies into `__meta` so the coordinator's `recoverPolicyManagerFromStateDB()` picks them up on startup
- `SetBlockHeight` — sets the last committed block number so the committer resumes from the right point
- `VerifyImport` — checks row counts and block height after import to catch integrity issues

No existing code is modified (except adding one test helper method to `test_exports.go`). The importer reuses the existing `database` struct, SQL templates, and table schema — imported data lands in the same `ns_<id>` tables the normal pipeline uses.

14 tests including a 10K-key bulk load and a multi-namespace end-to-end test that verifies imported policies are readable by `readNamespacePolicies()`.

### DIFF
--- a/service/vc/state_import.go
+++ b/service/vc/state_import.go
@@ -1,0 +1,217 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package vc
+
+import (
+	"context"
+	"encoding/binary"
+	"fmt"
+
+	"github.com/cockroachdb/errors"
+	"github.com/hyperledger/fabric-x-common/api/committerpb"
+	"github.com/yugabyte/pgx/v5"
+)
+
+// StateImporter provides bulk state import capabilities for initializing the
+// committer's state database from an external source such as a Fabric peer snapshot.
+//
+// Unlike the normal transaction processing pipeline (preparer -> validator -> committer),
+// imported state bypasses MVCC validation and endorsement verification since it originates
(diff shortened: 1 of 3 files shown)


sridhar-panigrahi:
@cendhu — the Exporter CLI that pairs with this StateImporter is now up at hyperledger/fabric-x#97. Between the two PRs the full export → bootstrap pipeline from the RFC is taking shape.

cendhu:
Thanks @veyron-kairo for your contributions. We are closing this PR as design is not finalized yet.

PR #521, opened 2026-04-06
[util] discover client user dir instead of hardcoding .com suffix

### LINKED ISSUE #520: `GetPeersMspDirs()` hardcodes `.com` domain suffix, breaks with custom crypto domains
`GetPeersMspDirs` in `utils/testcrypto/load_crypto.go` constructs user MSP paths by appending .com to the org directory name:

  `clientName := "client@" + mspItem.MspName + ".com"`

  This assumes `Domain == Name + ".com"`, which only holds for crypto generated by `gen_crypto.go` (which uses Domain: "peer-org-0.com"). When `cryptogen` is configured with a domain that doesn't include `.com` (e.g., Domain: peer-org-0), the constructed path `client@peer-org-0.com` doesn't match the actual directory `client@peer-org-0`.

  The fabric-x E2E integration test works around this with symlinks (`client@peer-org-0.com` -> `client@peer-org-0`).

  **Fix**: Scan the `users/` directory for a `client@*` entry instead of hardcoding the suffix. This works regardless of the domain convention used during crypto generation.

cendhu:
#### Type of change

- Bug fix
 
#### Description

  GetPeersMspDirs hardcoded ".com" when constructing user MSP paths,
  which broke when crypto-config used a domain without that suffix.
  Scan the users/ directory for a "client@" entry instead.

#### Related issues

 - resolves #520

### DIFF
--- a/utils/testcrypto/load_crypto.go
+++ b/utils/testcrypto/load_crypto.go
@@ -9,6 +9,7 @@ package testcrypto
 import (
 	"os"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/cockroachdb/errors"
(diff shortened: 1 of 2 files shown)


liran-funaro (approved):
LGTM

PR #528, opened 2026-04-08
[viper] Centralize configuration defaults in service packages

### LINKED ISSUE #426: [config] define default in one place
Define all defaults in cmd/config so Viper handles them directly. Otherwise, we are unnecessarily maintaining defaults in two locations.
Removing local defaults might require tests to pass explicit configurations, which helps prevent default divergence. To ensure consistency, we could export default values from here for cmd/config to use, or only set them here if a "second line of defense" is strictly necessary.

cendhu:
#### Type of change

#### Description

  Move hardcoded default values from `cmd/config/viper.go` to their respective
  service packages (coordinator, verifier, vc, query). Export these
  values as constants so they can be reused by the viper initialization logic.

  This consolidation reduces duplication and ensures that the defaults used
  by the CLI stay in sync with the service implementations.

  Key changes:
   - Exported Default... constants in service/coordinator, service/verifier, service/vc, and service/query.
   - Updated cmd/config/viper.go to use these constants.
   - Updated cmd/config/app_config_test.go to align with the centralized defaults.

  - Removed yaml and json tags as they are not required.
  - Added validator tags to validate the input.

#### Related issues

  - resolves #426


cmd/config/samples/query.yaml
```diff
@@ -24,11 +24,13 @@ server:
   rate-limit:
     # Maximum number of requests per second allowed.
     # Set to 0 to disable rate limiting.
+    # Default: 5000
     requests-per-second: 0
     # Maximum number of requests allowed in a single burst.
     # When rate limiting is enabled (requests-per-second > 0):
     #   - Must be greater than 0
     #   - Must be less than or equal to requests-per-second
+    # Default: 1000
```
liran-funaro: nit: please use thousands separator
```suggestion
    # Default: 1,000
```

utils/ordererdial/config.go
liran-funaro: I suggest adding `validate:"omitempty,oneof=tls mtls none"`

utils/testcrypto/load_crypto.go
```diff
@@ -119,6 +119,7 @@ func GetOrdererConnConfig(artifactsPath string, clientTLSConfig connection.TLSCo
 			Multiplier:      2,
 			MaxElapsedTime:  time.Second,
 		},
-		Identity: id,
+		Identity:                     id,
+		SuspicionGracePeriodPerBlock: time.Second,
```
liran-funaro: not needed

utils/deliverorderer/orderer_test.go
```diff
@@ -176,46 +176,50 @@ func TestNewOrdererDeliverEdgeCases(t *testing.T) {
 	t.Run("invalid FaultToleranceLevel", func(t *testing.T) {
 		t.Parallel()
 		_, err := deliverorderer.ToQueue(cancelledContext, deliverorderer.Parameters{
-			FaultToleranceLevel: "invalid",
-			TLS:                 *tls,
-			OutputBlock:         make(chan *common.Block, 10),
+			FaultToleranceLevel:          "invalid",
+			TLS:                          *tls,
+			OutputBlock:                  make(chan *common.Block, 10),
+			SuspicionGracePeriodPerBlock: time.Second,
```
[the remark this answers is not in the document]
liran-funaro: But it doesn't make sense to crash everything when this value is zero, as it is not required for CFT.
I suggest only failing when using BFT.

liran-funaro (approved):

PR #529, opened 2026-04-08
[sidecar] add notifier and relay throughput metrics

### LINKED ISSUE #250: [metrics] notifier
The notifier in sidecar does not have metrics. It would be good to find how many clients are connected and number of pending requests/timeout occurred.

cendhu:
#### Type of change

- New feature
 
#### Description

  Add Prometheus metrics for the notifier subsystem: active streams,
  pending subscriptions, unique pending tx IDs, status deliveries, and
  timeout deliveries. Move transaction_out_total from notifier to
  relay
  where it counts statuses actually processed from the coordinator,
  making it a true counterpart to transaction_in_total.

  Also adds per-request and global active tx ID limits with partial
  acceptance, and fixes removeAndEnqueueTimeoutEvents to count actual
  subscription removals instead of all matching tx IDs.

#### Related issues

  - resolves #250

### DIFF
--- a/service/sidecar/notify.go
+++ b/service/sidecar/notify.go
@@ -113,20 +113,20 @@ func (n *notifier) run(ctx context.Context, statusQueue chan []*committerpb.TxSt
 				})
 			}
 		case status := <-statusQueue:
-			removed, uniqueRemoved := notifyMap.removeAndEnqueueStatusEvents(status)
-			n.recordRemovals(removed, uniqueRemoved)
-			promutil.AddToCounter(n.metrics.notifierTxIDsStatusDeliveries, removed)
+			pendingTxIDsRemoved, uniquePendingTxIDsRemoved := notifyMap.removeAndEnqueueStatusEvents(status)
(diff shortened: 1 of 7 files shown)


service/sidecar/notify.go
```diff
@@ -116,26 +119,41 @@ func (n *notifier) run(ctx context.Context, statusQueue chan []*committerpb.TxSt
...
+}
+
 // OpenNotificationStream implements the [protonotify.NotifierServer] API.
 func (n *notifier) OpenNotificationStream(stream committerpb.Notifier_OpenNotificationStreamServer) error {
+	n.metrics.notifierActiveStreams.Inc()
+	defer n.metrics.notifierActiveStreams.Dec()
+
 	g, gCtx := errgroup.WithContext(stream.Context())
-	requestQueue := channel.NewWriter(gCtx, n.requestQueue)
+	globalRequestQueue := channel.NewWriter(gCtx, n.requestQueue)
```
liran-funaro: What is the reasoning for adding the "global" prefix? Is there a local one?

liran-funaro (approved):

PR #530, opened 2026-04-08
dynamic TLS CA rotation for sidecar and query service

### LINKED ISSUE #290: [sidecar, notification & query service] Dynamic updates of server root CAs according to config-block

cendhu:
#### Type of change

- New feature

#### Description

Add dynamic TLS support so that peer organization CA certificates are automatically updated when config blocks add or remove organizations.

- DynamicTLSManager in utils/connection/ watches config block updates and rebuilds the trusted CA pool on the fly
- Sidecar updates TLS immediately during config block processing
- Query service polls the DB periodically for config changes
- Integration test covers add/remove/restore of peer orgs with mTLS
- Infrastructure plumbing for QueryTLSRefreshInterval and PeerOrganizationCount in runner config

#### Related issues

 - resolves #290

### DIFF
--- a/service/query/query_service.go
+++ b/service/query/query_service.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"github.com/google/uuid"
+	"github.com/hyperledger/fabric-lib-go/common/flogging"
 	"github.com/hyperledger/fabric-x-common/api/applicationpb"
 	"github.com/hyperledger/fabric-x-common/api/committerpb"
 	"golang.org/x/sync/semaphore"
@@ -28,6 +29,8 @@ import (
 	"github.com/hyperledger/fabric-x-committer/utils/grpcerror"
 	"github.com/hyperledger/fabric-x-committer/utils/monitoring/promutil"
 )
+
+var logger = flogging.MustGetLogger("query")
 
 var (
 	// ErrInvalidOrStaleView is returned when attempting to use wrong, stale, or cancelled view.
@@ -58,16 +61,20 @@ type (
 		metrics     *perfMetrics
 		ready       *channel.Ready
 		healthcheck *health.Server
+		tlsUpdater  connection.TLSCertUpdater
 	}
 )
 
 // NewQueryService create a new QueryService given a configuration.
-func NewQueryService(config *Config) *Service {
+// The tlsUpdater is optional; when non-nil, it is used to push updated
+// root CA certificates read from the database.
+func NewQueryService(config *Config, tlsUpdater connection.TLSCertUpdater) *Service {
 	return &Service{
 		config:      config,
 		metrics:     newQueryServiceMetrics(),
 		ready:       channel.NewReady(),
 		healthcheck: connection.DefaultHealthCheckService(),
+		tlsUpdater:  tlsUpdater,
 	}
 }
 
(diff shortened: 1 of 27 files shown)


service/query/query_service.go
```diff
@@ -107,6 +114,10 @@ func (q *Service) Run(ctx context.Context) error {
 	}
 	q.ready.SignalReady()
 
+	// TLS refresh runs as a standalone goroutine rather than in an errgroup because
+	// a transient failure to read config from the DB should not stop the query service.
+	go q.refreshTLSFromDB(ctx, pool)
```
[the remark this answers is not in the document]
cendhu: Yes, I agree. As the impact is going to be negligible, this simple design is preferred. PostgreSQL supports LISTEN/NOTIFY. YugabyteDB has implemented it but yet to be released. Hence, this approach will be retained for a few more months and then we go with the cleaner approach.

service/query/query_service.go
```diff
@@ -320,6 +331,60 @@ func (q *Service) requestLatency(method string, start time.Time) {
...
+	defer ticker.Stop()
+
+	var lastVersion uint64
+	seen := false
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			configTX, err := queryConfig(ctx, pool)
```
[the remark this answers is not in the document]
cendhu: Not needed. `pgx` connections already have timeout protection configured in [ConfigureConnReadDeadline](https://github.com/hyperledger/fabric-x-committer/blob/5d2d7c4d79993b6a07163adb8c85e830163a13db/utils/dbconn/database_connection.go#L93): 10s dial timeout and 60s per-read deadline on the underlying connection. So queryConfig won't block indefinitely even without an explicit context timeout — the read deadline on the connection will fail it.

utils/connection/server.go
```diff
@@ -40,9 +40,9 @@ var (
 )
 
 // GrpcServer instantiate a [grpc.Server].
-func (c *ServerConfig) GrpcServer() (*grpc.Server, error) {
+func (c *ServerConfig) GrpcServer(tlsProvider TLSConfigProvider) (*grpc.Server, error) {
```
dean-amar, line 43: We are implicitly passing a function here as an argument. If this is something we want to avoid, we should think of another way to do so.

This is true for all the chain of functions using this argument.

liran-funaro: This implementation accepts an interface, not a method.
While systematically this is identical to passing a method, it adds clarity as we can easily discover all the implementations of this interface.
This is more difficult to achieve with a method (find all methods that have this specific signature).

cendhu: Valid point. I can explain why I am afraid of functions as argument or in struct. Building on Liran's comment, I generally try to avoid passing functions as arguments for core components. I'm speaking from some painful experience here—reading through the [gossip code](https://github.com/hyperledger/fabric/tree/main/gossip) in Fabric taught me that while passing functions is super quick to write initially, it makes tracing the execution path a real headache down the road. After a few years, only the people who wrote the code were able to make modifications to gossip, even though all Fabric maintainers were experienced and great at coding.

Go standard library does use func as an argument or func in struct (like in `tls.Config`). But it really comes down to the use case: if we are writing a broadly used library package, leaning towards functions for configuration options makes sense because it prioritizes the convenience of the people consuming the code. One does not need to write a whole new struct and implement an interface just to pass in custom behavior. A great example of this is `sort.Slice(mySlice, func(i, j int) bool { ... })`—it's vastly easier for a developer to drop in an anonymous function than it is to define a custom type just to satisfy the `sort.Interface`.

However, since we are not writing an external library, leaning towards interfaces is usually the better move. It prioritizes strict contracts, maintainability, and traceability. We don't really need that drop-in consumer flexibility; an interface guarantees we can easily track down exactly what is implementing the behavior without having to hunt through the codebase for anonymous functions.

liran-funaro (approved):
I don't understand some of the design decisions in this PR, but I'll allow this due to time constraints.

cendhu:
> I don't understand some of the design decisions in this PR, but I'll allow this due to time constraints.

Sure. We can discuss this further on an issue.

Issue #326, opened 2026-02-12
[configurations] Reduce YAML duplications


dean-amar:
Currently, we have repeating sections in our YAMLs (samples and templates). To reduce duplication, we need to parse the configuration differently and create a template file for each duplicated section. 

Note: if it adds too much complexity to the code rather than simple concatenation, it may not be worth the effort.

_Originally posted by @dean-amar in https://github.com/hyperledger/fabric-x-committer/pull/277#discussion_r2798074525_

PR #534, opened 2026-04-09
[config] rename template files from .yaml to .yaml.tmpl

cendhu:
#### Type of change

#### Description

Rename config template files to .yaml.tmpl to distinguish them from
regular YAML files and avoid yamllint false positives on Go template
syntax. Also rename loadgen client templates to match the service they
configure (e.g., loadgen_client_sidecar -> loadgen_committer).

#### Related issues

 - part of #326


liran-funaro (approved):
LGTM

PR #535, opened 2026-04-09
[config] Reduce duplicates in config templates

### LINKED ISSUE #326: [configurations] Reduce YAML duplications

cendhu:
#### Type of change

#### Description

The config templates had identical blocks (TLS, endpoint, monitoring,
logging, database) copy-pasted across 14 files. While these blocks
don't change often, when they do, every copy must be updated in sync —
a tedious and error-prone process. More importantly, reading any single
template required wading through boilerplate to find the service-specific
config that actually differs.

This refactoring extracts shared patterns into named template blocks.
Each service template now reads as a concise list of composed blocks
plus its unique config, making the service-specific parts immediately
visible. When a shared pattern does need to change, it's updated in
one place.

- Extract shared template blocks into shared.yaml.tmpl:
  tlsConfig, endpointWithTLS, endpointListWithIndividualTLS,
  endpointListWithSharedTLS, monitoring, logging, rateLimit,
  ordererClientConfig, and database
- Extract loadgen-specific shared block into loadgen_shared.yaml.tmpl:
  loadgenConfig
- Add custom "include" template function to enable piping template
  output through sprig's "indent" for correct YAML nesting at any level
- Restructure loadgen templates: replace string concatenation with
  define/template composition, each loadgen variant is now a standalone
  main template referencing shared blocks
- Switch from html/template to text/template (correct for YAML generation)

#### Related issues

 - resolves #326

### DIFF
--- a/cmd/config/create_config_file.go
+++ b/cmd/config/create_config_file.go
@@ -10,10 +10,11 @@ import (
 	"bytes"
 	_ "embed"
 	"fmt"
-	"html/template"
 	"os"
 	"path"
+	"strings"
 	"testing"
+	"text/template"
 	"time"
 
 	sprig "github.com/go-task/slim-sprig/v3"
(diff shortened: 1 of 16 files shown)


dean-amar (approved):
LGTM.

liran-funaro (approved):
LGTM. I left additional comments/questions.

Issue #536, opened 2026-04-09
[config] Extract server config from the services' config


liran-funaro:
This issue suggest removing the server config part from all services config.
Then, create a separate server config (e.g., in the `grpcservice` package).

```go
	Config struct {
		Server     *connection.ServerConfig `mapstructure:"server"`
		HTTPServer *connection.ServerConfig `mapstructure:"http-server"`
		Monitoring *connection.ServerConfig `mapstructure:"monitoring" `
		Logging    *logging.Config          `mapstructure:"logging"`
	}
```

The method `grpcservice.Serve()` will the relevant servers according to this config.
We will modify `cmd.config.ReadYamlAndSetupLogging()` to also load the server config.

This will create a clear separation between the server and the service.
The server part will start all the necessary servers, while the service will focus on the actual logic.

PR #540, opened 2026-04-13
Move Monitoring Server out of Service Logic

cendhu:
#### Type of change

#### Description

 The HTTP server for metrics was extracted from the servicei
 logic into utils/monitoring/http_server.go, with services
 now injected with a server rather than managing it themselves.
 The separation decouples metrics collection from HTTP serving
 across all services (coordinator, sidecar, query, vc, verifier, loadgen).

#### Related issues

### DIFF
--- a/cmd/committer/start_cmd.go
+++ b/cmd/committer/start_cmd.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 
 	"github.com/cockroachdb/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/cobra"
 
 	"github.com/hyperledger/fabric-x-committer/cmd/cliutil"
@@ -19,7 +20,9 @@ import (
 	"github.com/hyperledger/fabric-x-committer/service/sidecar"
 	"github.com/hyperledger/fabric-x-committer/service/vc"
 	"github.com/hyperledger/fabric-x-committer/service/verifier"
+	"github.com/hyperledger/fabric-x-committer/utils/connection"
 	"github.com/hyperledger/fabric-x-committer/utils/grpcservice"
+	"github.com/hyperledger/fabric-x-committer/utils/monitoring"
 )
 
 func startCMD() *cobra.Command {
@@ -56,9 +59,14 @@ func startService(ctx context.Context, name, configPath string) error {
 		return err
 	}
 
+	metricsProvider := monitoring.NewMetricsProvider()
+	if err := startMonitoringServer(ctx, conf, metricsProvider.Registry()); err != nil {
+		return err
+	}
+
 	switch c := conf.(type) {
 	case *sidecar.Config:
-		service, err := sidecar.New(c)
+		service, err := sidecar.New(c, metricsProvider)
 		if err != nil {
 			return errors.Wrap(err, "failed to create sidecar service")
 		}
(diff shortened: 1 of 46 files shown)


liran-funaro:
@cendhu I'm working on a PR that also solves the same issue.
We seem to have duplicated effort.
- #536

cendhu:
@liran-funaro I see. Maybe we can merge good things from both work. Can you push the changes so that we can see whether I should rebase on top of yours or the other-way?

liran-funaro:
@cendhu My PR is not ready.

liran-funaro:
@cendhu You can browse https://github.com/liran-funaro/fabric-x-committer/tree/refs/heads/external-server-config

cendhu:
> @cendhu You can browse https://github.com/liran-funaro/fabric-x-committer/tree/refs/heads/external-server-config

I like the approach of using single serve call to start both service and monitoring. I did not understand the difference between HTTP and Monitoring. I was thinking of providing a single http endpoint for metrics, go debug profile, and logging profile for simplicity. If we can reduce this struct to just GRPC and HTTP and grpcListener and httpListener, it would read well. 

```go
	Servers struct {
		GRPC       *grpc.Server
		HTTP       *runtime.ServeMux
		Monitoring *http.ServeMux

		httpServer       *http.Server
		monitoringServer *http.Server

		grpcListener       net.Listener
		httpListener       net.Listener
		monitoringListener net.Listener
	}
```

liran-funaro:
@cendhu  The HTTP is for the loadgen. Theoretically, it can use the same server as the monitoring.

cendhu:
> > @cendhu You can browse https://github.com/liran-funaro/fabric-x-committer/tree/refs/heads/external-server-config
> 
> I like the approach of using single serve call to start both service and monitoring. I did not understand the difference between HTTP and Monitoring. I was thinking of providing a single http endpoint for metrics, go debug profile, and logging profile for simplicity. If we can reduce this struct to just GRPC and HTTP and grpcListener and httpListener, it would read well.
> 
> ```go
> 	Servers struct {
> 		GRPC       *grpc.Server
> 		HTTP       *runtime.ServeMux
> 		Monitoring *http.ServeMux
> 
> 		httpServer       *http.Server
> 		monitoringServer *http.Server
> 
> 		grpcListener       net.Listener
> 		httpListener       net.Listener
> 		monitoringListener net.Listener
> 	}
> ```

While I like the utils/serve design, I am not sure about `c.metrics.Register(s.Monitoring)` from the service. I thought the goal was to decouple this logic from service.

liran-funaro:
>While I like the utils/serve design, I am not sure about c.metrics.Register(s.Monitoring) from the service. I thought the goal was to decouple this logic from service.

It is impossible to decouple.
Either we use my approach (register), or your approach (pass in the constructor). But in both ways, there is coupling.
It is just in a different place in the code.

cendhu:
> > While I like the utils/serve design, I am not sure about c.metrics.Register(s.Monitoring) from the service. I thought the goal was to decouple this logic from service.
> 
> It is impossible to decouple. Either we use my approach (register), or your approach (pass in the constructor). But in both ways, there is coupling. It is just in a different place in the code.

In this PR, we are only sending metricsProvider to each service. No knowledge about the server config or http server is sent.

liran-funaro:
@cendhu Does this PR have additional contributions that were not available in PR #544 ?

cendhu:
@liran-funaro There is one minor refactoring related to monitorQueues which I would push as a separate PR.

PR #542, opened 2026-04-14
Introduce new process for testing

### LINKED ISSUE #545: [integration] Remove dependency on `ifrit` and `gomega`
`ifrit` is not maintained and is using an old deprecated version of `gomega`.
It requires a global fail handler registration which causes race condition with parallel tests.

liran-funaro:
#### Type of change

#### Description

- Introduce new process for testing

#### Additional details

`ifrit` is not maintained and is using an old deprecated version of `gomega`.
This PR removes the dependency in `ifrit` and `gomega`.

#### Related issues

- resolves #545


utils/test/process.go
```diff
@@ -0,0 +1,90 @@
...
+	}()
+
+	return p, nil
+}
+
+func (p *Process) streamOutput(reader io.Reader) {
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		line := scanner.Text()
+		fmt.Printf("\033[0;31m[%-15s]\033[0m %s\n", p.name, line)
```
cendhu: Difficult to decode what this means - `\033[0;31m[%-15s]\033[0m %s\n"`

cendhu (approved):
LGTM.

utils/test/process.go
liran-funaro: addressed

PR #544, opened 2026-04-16
Split server and service configs

### LINKED ISSUE #441: [metrics] Extract monitoring server startup from service implementation
Currently, Prometheus monitoring server startup is embedded in each service's Run() method, tightly coupling service logic with the server's activation. This should be refactored to follow the same pattern as the gRPC server, where the server lifecycle is managed externally by connection.StartService().

## Current Architecture

- Each service creates its own metrics provider internally (e.g., `newPerformanceMetrics()`)
- The Prometheus server is started inside `service.Run()` as a goroutine
- The CMD layer has no control over the monitoring server lifecycle
- Services are responsible for both business logic and monitoring infrastructure

## Proposed Architecture

Similar to how `connection.StartService()` manages the gRPC server lifecycle:

1. Add a new method to the `Service` interface (e.g., `StartMonitoringServer(ctx context.Context) error`)
2. Extract monitoring server startup logic from `service.Run()` into this new method
3. Update `connection.StartService()` to start both:
   - The gRPC server(s) via `RunGrpcServer()`
   - The monitoring server via the new interface method

Please share your thoughts @cendhu, @liran-funaro.

### LINKED ISSUE #536: [config] Extract server config from the services' config

liran-funaro:
#### Type of change

#### Description

- Split server and service configs
- Move all server handling to `serve` package

#### Related issues

- resolves #536 
- resolves #441


utils/serve/config.go
```diff
@@ -0,0 +1,86 @@
...
+type (
+	// Config holds serving infrastructure configuration.
+	// This is parsed separately from service-specific configuration.
+	// The GRPC serves the GRPC service.
+	// The HTTP serves the service's HTTP API.
+	// The Monitoring serves the Prometheus metrics service.
+	Config struct {
+		GRPC       ServerConfig `mapstructure:"server"`
+		HTTP       ServerConfig `mapstructure:"http-server"`
+		Monitoring ServerConfig `mapstructure:"monitoring"`
```
cendhu: gRPC and HTTP are clearer terms than monitoring. We should limit each service to one gRPC and one HTTP endpoint, regardless of the specific HTTP methods used. Loadgen is the only exception requiring two servers, but we shouldn't let that clutter our core service code. Using one of each avoids confusion and keeps the architecture simple. If you would like to retain two HTTP endpoints for the loadgen, start one HTTP endpoint within the loadgen package itself.

cmd/cliutil/test_exports.go
```diff
@@ -71,21 +70,28 @@ func StartDefaultSystem(t *testing.T) config.SystemConfig {
...
 		LedgerPath: t.TempDir(),
 	}
 }
 
+func convertServiceConfig(c *serve.Config) config.ServiceConfig {
+	return config.ServiceConfig{
+		GrpcEndpoint:    &c.GRPC.Endpoint,
+		MetricsEndpoint: &c.Monitoring.Endpoint,
+		GrpcTLS:         c.GRPC.TLS,
+		MetricsTLS:      c.Monitoring.TLS,
```
cendhu: Maybe in a separate PR, we should rename MetricsEndpoint to HTTPEndpoint and MetricsTLS to httpTLS.

loadgen/client.go
```diff
@@ -163,10 +155,13 @@ func (c *Client) WaitForReady(ctx context.Context) bool {
...
-func (c *Client) RegisterService(server *grpc.Server) {
-	servicepb.RegisterLoadGenServiceServer(server, c)
-	healthgrpc.RegisterHealthServer(server, c.healthcheck)
+// RegisterService registers the loadgen's gRPC services and monitoring server.
+func (c *Client) RegisterService(s serve.Servers) {
+	servicepb.RegisterLoadGenServiceServer(s.GRPC, c)
+	healthgrpc.RegisterHealthServer(s.GRPC, c.healthcheck)
+	monitoring.RegisterMonitoringServer(s.Monitoring, c.resources.Metrics.Provider)
+	// The following call returns error, but always returns nil.
+	_ = servicepb.RegisterLoadGenServiceHandlerServer(context.Background(), s.HTTP, c)
```
cendhu: If we rename s.Monitoring to s.HTTP similar to s.GRPC< we can use one HTTP endpoint.

service/sidecar/block_delivery_test.go
```diff
@@ -15,25 +15,32 @@ import (
...
 	"github.com/hyperledger/fabric-x-committer/utils/test"
 )
 
+type blockDeliveryWrapper struct {
+	*blockDelivery
+}
+
+func (w *blockDeliveryWrapper) RegisterService(s serve.Servers) {
+	peer.RegisterDeliverServer(s.GRPC, w.blockDelivery)
+}
```
cendhu, line 31: Not introduced by this PR, but I am wondering why we have multiple grpcServices for the sidecar alone. What is the advantage with that approach over single grpcService? It does require refactoring of APIs on fabric-x-common but adds simplicity to the codebase.

utils/serve/tls_provider.go
```diff
@@ -0,0 +1,127 @@
...
+		return false
+	}
+
+	mergedPool := d.staticCertPool.Clone()
+	// We ignore failed certificates here, and process the rest.
+	// This ensures we allow partial updates to succeed even if some certs are invalid.
+	// Otherwise, a single bad certificate could block access to the system to new clients,
+	// or allow access to invalid clients.
+	connection.ExtendCertPool(mergedPool, d.updater.Load()...)
+	d.clientConfig.ClientCAs = mergedPool
```
cendhu: As per tls config doc comment
```
// A Config structure is used to configure a TLS client or server.
// After one has been passed to a TLS function it must not be
// modified. A Config may be reused; the tls package will also not
// modify it.
```
we sholud do 

```go
   newConfig := d.clientConfig.Clone()
   newConfig.ClientCAs = mergedPool
   d.clientConfig = newConfig
 ```

File: utils/serve/start_serve.go [hunk 6]
```diff
@@ -0,0 +1,335 @@
...
+		return s, errors.Wrapf(err, "failed creating HTTP server")
+	}
+
+	s.Monitoring = http.NewServeMux()
+	s.monitoringServer, err = newHTTPServer(&conf.Monitoring, s.Monitoring)
+	if err != nil {
+		return s, errors.Wrapf(err, "failed creating monitoring server")
+	}
+
+	if !conf.GRPC.Endpoint.Empty() {
```
cendhu, line 138: I do not understand why do we create TLS provider and server when the endpoint is empty? This is applicable for HTTP server too.

liran-funaro: We create the server even if we do not deploy it. This allows us to avoid `nil` checks in the `service.Register()` method in case the endpoint is empty.

cendhu: Please add this as reasoning comment.

utils/serve/config.go
liran-funaro: Good point. I consolidated the monitoring server and HTTP server.
The field is called `HTTP` but for now, I did not change the YAML field name "monitoring".

loadgen/client.go
liran-funaro: done

service/sidecar/block_delivery_test.go
liran-funaro: I don't know. This was introduced by your PR, and I assumed it was either a constraint or a convenience to reduce code in `sidecar.go`.

utils/serve/tls_provider.go
liran-funaro: done

cmd/cliutil/test_exports.go
liran-funaro: We can also do it in this PR. Whatever you prefer.

service/sidecar/block_delivery_test.go
cendhu: peer delivery was introduced as a separate service as it was exposing the Fabric delivery APIs. We can also embed this in the sidecar object along with notifier and block query.

utils/serve/start_serve.go
liran-funaro: done

cmd/cliutil/test_exports.go
cendhu: I have created an issue. Let's address in a separate PR.

cendhu (approved):
LGTM.

Please add changes made due to review comments in a separate PR as earlier. We have two force pushes but they are not showing the change with respect to the removal of monitoring http endpoint from the config.

PR #550, opened 2026-04-20
[docs] add site url to mkdocs

cendhu:
#### Type of change

- Documentation update
 
#### Description

Added site_url (https://hyperledger.github.io/fabric-x-committer/ ) to the mkdocs so that the theme is applied.


cendhu:
@liran-funaro I am merging this as I need to share this with Gabi (if it works).

PR #551, opened 2026-04-20
Fixes sidecar diagram and add queyr and verifier

liran-funaro:
#### Type of change

- Documentation update
 
#### Description

- Minor fixes to the sidecar diagram
- Add Query service and Verifier diagrams


cendhu (approved):
I did not go over line by line. We can read it via the hosted docs and fix if there are any issues.

PR #552, opened 2026-04-22
feat(deliverorderer): add skip-block-signature-verification escape hatch

tinystone007:
## Problem

\`fabric-x-committer\` v0.2.0 introduced block-signature verification in \`utils/deliverorderer\` (the \`verifyBlockPolicy\` path inside \`blockVerificationStateMachine\`). The check evaluates the channel's BlockValidation policy against the block metadata signatures delivered by the orderer.

\`hyperledger/fabric-x-orderer\` (all released lines — v0.0.24, v0.1.0, v1.0.0-alpha) still carries this in \`node/crypto/signer.go\`:

\`\`\`go
// TODO: implement correct Serialize
// Serialize is called when a SignatureHeader.Creator is created. Since this creator is placeholder, the SignatureHeader.Creator must be updated with correct creator.
func (s ECDSASigner) Serialize() ([]byte, error) {
    return []byte("creator"), nil
}
\`\`\`

As a result every block metadata signature carries a literal \`"creator"\` placeholder as its Creator, which does not unmarshal as an \`msp.SerializedIdentity\` and satisfies zero sub-policies of the Orderer group's Writers. Every non-genesis block is rejected with:

\`\`\`
block signature verification failed on block [N]:
implicit policy evaluation failed - 0 sub-policies were satisfied
\`\`\`

v0.1.9 worked by accident — \`deliverorderer\` did not yet do this check, so the mismatched signatures were never evaluated. v0.2.0 added the check, but the orderer side was never updated, leaving the two halves of the ecosystem mutually incompatible on default config.

## This PR

Adds an optional yaml field:

\`\`\`yaml
orderer:
  skip-block-signature-verification: true   # default: false
\`\`\`

When set, \`configState.updateIfConfigBlock\` leaves \`verifierFunc\` nil, and the existing \`if cs.verifierFunc == nil { return nil }\` guard inside \`verifyBlockPolicy\` short-circuits the check. Default is \`false\`, so anyone running against a fixed orderer keeps strict verification.

### Plumbing

- \`ordererdial.Config\` gains \`SkipBlockSignatureVerification\`.
- \`deliverorderer.Parameters\` + \`deliverorderer.SessionInfo\` mirror the flag.
- \`LoadParametersFromConfig\` copies it through.
- \`newBlockProcessingState\` seeds the state machine + latest-config.
- \`configState.updateIfConfigBlock\` leaves \`verifierFunc\` nil when the flag is set.

4 files, +38/-13.

## Why ship this as an escape hatch rather than waiting for the orderer fix

We run \`fabric-x-committer\` against the upstream orderer today and cannot proceed without an opt-in workaround. The alternative — fixing \`ECDSASigner.Serialize\` — requires protocol-level coordination (the orderer writes \`common.MetadataSignature\` with \`IdentifierHeader\` rather than \`SignatureHeader\`, so the real path to "correct" is either orderer-side \`SignatureHeader.Creator\` population or committer-side \`IdentifierHeader\` + consenter-map resolution). That's a larger change across two repos.

Shipping this flag:
- lets operators keep the deploy unblocked (required for us, probably for anyone else integrating v0.2.0+),
- defaults off so no behavior change for anyone on a fixed orderer,
- is strictly additive — trivially reversible when the orderer side is fixed. At that point, downstream config simply drops the flag and normal verification resumes.

## Testing

Running locally against \`fabric-x-orderer\` v0.1.0 + this patch:
- With \`skip-block-signature-verification: true\`: committer tails the orderer deliver stream, commits all non-genesis blocks. Token Issue / Redeem / Transfer flows through (10+ end-to-end scenarios pass).
- With flag omitted / \`false\`: reproduces \`implicit policy evaluation failed - 0 sub-policies were satisfied\` on block 1.

Happy to extend an existing test, add a unit test in \`utils/deliverorderer\`, or split the patch as reviewers prefer.

### DIFF
--- a/utils/deliverorderer/config.go
+++ b/utils/deliverorderer/config.go
@@ -91,6 +91,9 @@ type (
 		OutputBlockWithSourceID chan<- *deliver.BlockWithSourceID
 
 		SuspicionGracePeriodPerBlock time.Duration
+
+		// SkipBlockSignatureVerification mirrors ordererdial.Config.SkipBlockSignatureVerification.
+		SkipBlockSignatureVerification bool
 	}
 
 	// SessionInfo represents the processing state of blocks to support recovery and resumption.
@@ -99,6 +102,9 @@ type (
 		LatestKnownConfig           *common.Block
 		NextBlockVerificationConfig *common.Block
 		LastBlock                   *common.Block
+
+		// SkipBlockSignatureVerification mirrors ordererdial.Config.SkipBlockSignatureVerification.
+		SkipBlockSignatureVerification bool
 	}
 )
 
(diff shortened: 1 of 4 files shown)


liran-funaro:
@tinystone007 Thanks for taking an interest in our project.
To skip block signature verification, one can use the `ToQueueWithNoFT()` method. The committer must be FT.
For the future, we suggest starting by creating an issue that describes the problem before submitting a PR.

PR #568, opened 2026-05-11
[Service Startup] Add configurable readiness timeouts

### LINKED ISSUE #586: [StartAndServe] Make readiness timeout configurable
Currently, the StartAndServe function uses a hardcoded 5-minute timeout when waiting for the service to become ready. This is inflexible for different environments or service types that may require shorter or longer grace periods.

We need to replace this hardcoded value with a configurable duration.

dean-amar:
#### Type of change

- Improvement (configurability)
 
#### Description

- Added configurable service readiness timeouts based on individual service settings.

#### Related issues

- Resolves #586


cmd/config/samples/coordinator.yaml
```diff
@@ -39,6 +39,10 @@ monitoring:
     #   - Must be less than or equal to requests-per-second
     burst: 0
 
+# Maximum time to wait for the coordinator service to become ready before startup fails.
+# This covers initialization that happens after Run starts and before the gRPC server begins serving.
+readiness-timeout: 5m
```
liran-funaro: Please make the comment user-oriented rather than developer-oriented.
Do not mention the "Run" method, but use terms that relate to the user (e.g., initialization before the server starts).
You can specify what we are waiting for from the coordinator to make this explanation clearer to the user.
Also, mention the default value as we do for other configurations.

This applies to all the services. If a service doesn't need to wait (WaitForReady() returns true), then we can omit this configuration from the YAML.

utils/serve/start_serve.go
```diff
@@ -61,6 +61,9 @@ type (
 	}
 )
 
+// DefaultReadinessTimeout is the default timeout for waiting for service readiness.
+const DefaultReadinessTimeout = 5 * time.Minute
```
liran-funaro: The defaults should be in `config.go` like in the other services.

utils/serve/config.go
```diff
@@ -23,6 +23,9 @@ type (
 	Config struct {
 		GRPC ServerConfig `mapstructure:"server"`
 		HTTP ServerConfig `mapstructure:"monitoring"`
+		// ReadinessTimeout is the maximum time to wait for a service
+		// to become ready before startup fails.
+		ReadinessTimeout time.Duration `mapstructure:"readiness-timeout" validate:"required,gt=0"`
```
liran-funaro: Please consider other names. `ReadinessTimeout` doesn't sound right.
E.g. `ServiceStartupTimeout`. But feel free to suggest alternatives.

utils/serve/start_serve.go
```diff
@@ -76,7 +79,16 @@ func StartAndServe(ctx context.Context, service Service, serverConfig ...*Config
 		return service.Run(gCtx)
 	})
 
-	ctxTimeout, cancelTimeout := context.WithTimeout(gCtx, 5*time.Minute) // TODO: make this configurable.
+	// Extract the readiness timeout from the first config that defines one.
+	var readinessTimeout time.Duration
```
liran-funaro: Ensure we have a positive value, even if we use this method with no server config.
```suggestion
	readinessTimeout := DefaultReadinessTimeout
```

utils/serve/start_serve.go
dean-amar: Done.

utils/serve/config.go
dean-amar: I went with this name. Done.

utils/serve/start_serve.go
dean-amar: Done.

cmd/config/samples/coordinator.yaml
dean-amar: Done.

utils/serve/start_serve.go
```diff
@@ -76,7 +76,16 @@ func StartAndServe(ctx context.Context, service Service, serverConfig ...*Config
 		return service.Run(gCtx)
 	})
 
-	ctxTimeout, cancelTimeout := context.WithTimeout(gCtx, 5*time.Minute) // TODO: make this configurable.
+	// Extract the readiness timeout from the first config that defines one.
```
liran-funaro: nit:
```suggestion
	// Extract the service startup timeout from the first config that defines one.
```

utils/serve/config.go
```diff
@@ -68,6 +71,9 @@ type (
 	}
 )
 
+// DefaultServiceStartupTimeout is the default timeout for waiting for service readiness.
```
liran-funaro: nit
```suggestion
// DefaultServiceStartupTimeout is the default timeout for waiting for service startup.
```

liran-funaro (approved):
LGTM. Need to fix some comments.

PR #569, opened 2026-05-11
[cmd] Fix automatic env

### LINKED ISSUE #557: TLS sidecars ignore certificate path environment variable overrides
## Summary
When enabling TLS for the Fabric-X committer container, the sidecar components appear to ignore the configured environment variable overrides for certificate/key paths and instead fall back to hardcoded paths inside the container.

## Observed behavior
From investigation in hyperledger-labs/fabric-smart-client PR #1298:
- Verifier, Query Service, and Orderer default to paths such as `/client-certs/public-key.pem` and `/server-certs/public-key.pem`
- the configured environment variable overrides do not appear to be used
- this leads to file-not-found startup failures unless crypto material is mounted directly into those hardcoded locations

## Expected behavior
The sidecar components should honor the provided environment variable overrides for certificate/key locations.

## Context
This was noticed while wiring TLS support for the Fabric-X committer container from:
- https://github.com/hyperledger-labs/fabric-smart-client/pull/1298

A workaround on the FSC side is to mount generated crypto material into the expected hardcoded directories, but this issue tracks the underlying behavior in `fabric-x-committer`.

### LINKED ISSUE #558: [BUG] Environment variables fail to override TLS settings due to YAML Aliases in sidecar.yaml
## Description
This issue was identified during the integration of the `fabric-x-committer` sidecar into the **Fabric Smart Client (FSC)** integration test suite (specifically [FSC PR #1298](https://github.com/hyperledger-labs/fabric-smart-client/pull/1298)).

We found that environment variable overrides (e.g., `SC_SIDECAR_ORDERER_TLS_MODE`) are ignored by the application, preventing successful TLS orchestration within the FSC framework. 

The root cause appears to be the use of **YAML Anchors and Aliases** in the default `sidecar.yaml` configuration file included in the Docker image.

## Technical Details
In `cmd/config/samples/sidecar.yaml`, the TLS configuration is defined as an anchor and then aliased across multiple services:

```yaml
server:
  tls: &tlsConfig
    mode: mtls
    cert-path: /root/artifacts/...

orderer:
  tls: *tlsConfig

committer:
  tls: *tlsConfig

### LINKED ISSUE #567: [cmd] Keys that exist in the config struct but are NOT in the YAML and have NO default CANNOT be overridden
Viper's `AutomaticEnv()` feature only works for configuration keys that Viper already knows about. This means:
- Keys present in the YAML file can be overridden by environment variables ✓
- Keys with defaults set via `SetDefault()` can be overridden by environment variables ✓
- Keys that exist in the config struct but are NOT in the YAML and have NO default CANNOT be overridden ✗

### PR DESCRIPTION
**Pull request by liran-funaro**
#### Type of change

#### Description

- Allow overriding with env variables keys that exist in the config struct but are NOT in the YAML and have NO default

#### Related issues

- resolves #567
- resolves #557 
- resolves #558

### DIFF
--- a/cmd/config/app_config.go
+++ b/cmd/config/app_config.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 
 	"github.com/cockroachdb/errors"
@@ -29,6 +30,10 @@ import (
 	"github.com/hyperledger/fabric-x-committer/utils/serve"
 )
 
+type loggincConfig struct {
+	Logging flogging.Config `mapstructure:"logging"`
+}
+
 var (
 	logger   = flogging.MustGetLogger("config-reader")
 	validate = validator.New()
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: cmd/config/app_config.go [hunk 1]
```diff
@@ -29,6 +30,10 @@ import (
 	"github.com/hyperledger/fabric-x-committer/utils/serve"
 )
 
+type loggincConfig struct {
```
cendhu:
```suggestion
type loggingConfig struct {
```

cmd/config/app_config.go
```diff
@@ -93,46 +98,46 @@ func ReadYamlAndSetupLogging[T any](v *viper.Viper, configPath, servicePrefix st
...
-	}
-	flogging.Init(*loggingWrapper.Logging)
 
-	// Parse server configuration separately (similar to logging)
-	serverConfig := &serve.Config{}
-	if err = unmarshal(v, serverConfig); err != nil {
-		return nil, nil, err
+	// Check if the YAML is overridden by an environment variable.
+	if envValue, ok := os.LookupEnv(v.GetEnvPrefix() + "_YAML"); ok && envValue != "" {
+		logger.Debugf("Overridding config YAML from env: %s", envValue)
```
cendhu: Should we print the envValue? This can leak some sensitive info in the logs.

cendhu (approved):
LGTM.

cmd/config/app_config.go
liran-funaro: Good point. I'll remove this.

cmd/config/app_config.go
liran-funaro: good catch

PR #595, opened 2026-05-17
[sidecar] Single object for delivery

### LINKED ISSUE #593: [sidecar] Use single object for all block-store API
Not introduced by this PR, but I am wondering why we have multiple grpcServices for the sidecar alone. What is the advantage with that approach over single grpcService? It does require refactoring of APIs on fabric-x-common but adds simplicity to the codebase.

_Originally posted by @cendhu in https://github.com/hyperledger/fabric-x-committer/pull/544#discussion_r3193270021_

### LINKED ISSUE #594: [sidecar] Potential resource leak (require close)
In case the user does not manually call `.Close()`, the sidecar will hold the block store

liran-funaro:
#### Type of change

#### Description

- Use a single Sidecar object for all delivery API
- Create block store object on `run()` and close it when it ends

#### Related issues

- resolves #594 
- resolves #593


cendhu (approved):
LGTM

PR #598, opened 2026-05-19
[ordererdial] Refactor ordererdial package to use connection.TLSConfig

### LINKED ISSUE #339: [Orderer setup] Remove common-ca-cert-paths
Common-ca-cert-paths is a temporary workaround to inject the CA into all organizations.
We already added the ability to read the TLS certificates from the config-block, so we need to remove it for proper testing and align with the production setup.

dean-amar:
#### Type of change
- Improvement (duplication reduction)
 
#### Description

- This PR refactors the `ordererdial` package to eliminate code duplication by leveraging the existing `connection.TLSConfig` structure instead of maintaining a separate TLS configuration implementation.

#### Additional details

- As discussed in #339, static local CAs are used across the committer components, and the sidecar keeps the orderer CA certificate paths as static configuration for consistency, even though it can load the orderer CAs from the genesis block at startup. 

#### Related issues

- Resolves #339


liran-funaro (approved):
LGTM

PR #601, opened 2026-05-21
[sidecar] Fix empty ledger newest delivery seek

### LINKED ISSUE #600: [sidecar] Empty ledger newest delivery seek causes uint64 underflow
When SeekPosition_Newest is used on an empty ledger (height=0):
   1. FileLedger.Iterator(Newest) internally computes Height()-1, underflowing to math.MaxUint64
   2. Stop=Newest computes height - 1, also underflowing

   Additionally, for BLOCK_UNTIL_READY behavior, the cursor was created before waiting for block zero, which could
   fail on FileLedger since the block-zero index doesn't exist yet.

   Fix: wait for ledger height > 0 before creating cursor; substitute Start=Newest with Specified{Number:0} on empty
   ledger; guard Stop=Newest against zero height.

cendhu:
#### Type of change

- Bug fix

#### Description

When SeekPosition_Newest is used on an empty ledger (height=0), FileLedger.Iterator(Newest) computes Height()-1, underflowing to math.MaxUint64. Similarly, Stop=Newest computes height-1, also underflowing. Additionally, for BLOCK_UNTIL_READY behavior, the cursor was created before waiting for block zero, which could fail on FileLedger since the block-zero index does not exist yet.

- Move WaitForCondition (height > 0) before cursor creation for BLOCK_UNTIL_READY
- In getCursor, replace Start=Newest with Specified{Number:0} when height is 0
- Guard Stop=Newest against zero height, returning 0 instead of height-1

#### Related issues

 - resolves #600

### DIFF
--- a/service/sidecar/block_delivery_test.go
+++ b/service/sidecar/block_delivery_test.go
@@ -7,7 +7,9 @@ SPDX-License-Identifier: Apache-2.0
 package sidecar
 
 import (
+	"context"
 	"testing"
+	"time"
 
 	"github.com/hyperledger/fabric-protos-go-apiv2/common"
 	ab "github.com/hyperledger/fabric-protos-go-apiv2/orderer"
@@ -18,6 +20,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	"github.com/hyperledger/fabric-x-committer/utils/connection"
 	"github.com/hyperledger/fabric-x-committer/utils/serve"
 	"github.com/hyperledger/fabric-x-committer/utils/test"
 )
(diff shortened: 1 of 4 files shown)


service/sidecar/block_delivery.go
```diff
@@ -135,7 +137,14 @@ func (s *blockDelivery) deliverBlocks( //nolint:gocognit
 }
 
 func (s *blockDelivery) getCursor(seekInfo *ab.SeekInfo) (blockledger.Iterator, uint64, error) {
-	cursor, number := s.blockStore.ledger.Iterator(seekInfo.Start)
+	start := seekInfo.Start
+	if _, ok := start.Type.(*ab.SeekPosition_Newest); ok && s.blockStore.ledger.Height() == 0 {
+		// FileLedger.Iterator(Newest) computes Height()-1. On an empty ledger,
+		// request block 0 through a local start value so seekInfo.Start stays unchanged.
+		start = &ab.SeekPosition{Type: &ab.SeekPosition_Specified{Specified: &ab.SeekSpecified{Number: 0}}}
+	}
```
liran-funaro: Wouldn't it make more sense to fix `fileledger.FileLedger` under `fabric-x-common`?

cendhu: Root underflow in `fileledger.FileLedger.Iterator(Newest)` fixed in companion common PR: hyperledger/fabric-x-common#116

Keeping committer-side fixes here for two reasons:
1. `Stop=Newest` underflow is in sidecar `getCursor`, not in common.
2. Until committer consumes a common release with that fix, `BLOCK_UNTIL_READY` must avoid cursor creation while ledger is empty.

After common release + dependency update, local `Start=Newest` empty-ledger workaround can be removed. Sidecar `Stop=Newest` guard still belongs here.

liran-funaro (commented):
I suggest updating to the latest fabric-x-common with the relevant fix.

service/sidecar/sidecar.go
liran-funaro: Unrelated to this PR:
Can you rename the variable `number` to `stopNum` as this was unclear when reviewing it?

service/sidecar/sidecar.go
```diff
@@ -540,24 +540,26 @@ func (s *Service) deliverBlocks(
...
-		return common.Status_BAD_REQUEST, err
-	}
-	defer cursor.Close()
-	logger.Debugf("Received seekInfo.")
 
 	ctx := srv.Context()
 
 	if seekInfo.Behavior == ab.SeekInfo_BLOCK_UNTIL_READY {
-		// We use a retry backoff here to avoid busy waiting when blocks are not yet available.
+		// Wait before creating the cursor. FileLedger cannot create a waiting block-zero
```
liran-funaro, line 547: Good catch

cendhu:
> I suggest updating to the latest fabric-x-common with the relevant fix.

Done.

service/sidecar/block_delivery_test.go
```diff
@@ -131,18 +134,117 @@ func newBlockStoreWithBlocks(t *testing.T, numBlocks int) (*blockStore, [][3]str
...
+			ensureAtLeastHeight(t, bs, 1)
+
+			ctx, cancel := context.WithTimeout(t.Context(), 2*time.Second)
+			defer cancel()
+			var result deliverResult
+			select {
+			case result = <-resultCh:
+			case <-ctx.Done():
+				require.NoError(t, ctx.Err())
+			}
```
liran-funaro: We can shorten the notation using existing tools.
```suggestion
			result, ok := channel.NewReader(t.Context(), resultCh).ReadWithTimeout(2*time.Second)
			require.True(t, ok)
```

liran-funaro (approved):
LGTM. Just some minor comments

service/sidecar/sidecar.go
cendhu: Renamed to `startNum` instead of `stopNum` — this variable holds the start block number returned by `Iterator()`, not the stop position. Using `stopNum` would be misleading since the stop number is computed in the switch cases below.

service/sidecar/block_delivery_test.go
cendhu: Done. Simplified to `channel.NewReader(t.Context(), resultCh).ReadWithTimeout(2 * time.Second)`.

service/sidecar/sidecar.go
```diff
@@ -578,22 +580,26 @@ func (s *Service) deliverBlocks(
 }
 
 func (s *Service) getCursor(seekInfo *ab.SeekInfo) (blockledger.Iterator, uint64, error) {
-	cursor, number := s.blockStore.ledger.Iterator(seekInfo.Start)
+	cursor, startNum := s.blockStore.ledger.Iterator(seekInfo.Start)
```
liran-funaro, line 583: Is it `startNum` or `stopNum`? The caller (line 556) uses `stopNum`.

cendhu: Yes, `startNum` is correct. It's the value returned by `Iterator(seekInfo.Start)` — it is the start position.

The reason it works as the stop number in two branches (seekOldest, seekNewest) is a logical coincidence: when `Stop=Oldest` or `Start==Stop==Newest`, the stop block equals the start block (deliver exactly one block then stop). So using `startNum` as the stop number is correct in those cases.

The variable describes what it is (the start number from the iterator), not how it's used (sometimes returned as the stop number).

PR #603, opened 2026-05-21
[sidecar] Fix empty timeout notifications

### LINKED ISSUE #602: [sidecar] Fix empty timeout notifications
**Problem**: The notifier emits empty NotificationResponse messages (TimeoutTxIds: []) to clients when a timeout
   callback races with a status delivery.

   **Root cause**: removeAndEnqueueTimeoutEvents iterates over the request's TX IDs and skips any that are no longer in
   the subscription map (already removed by a prior status delivery). However, after the loop, it unconditionally
   writes a NotificationResponse to the stream event queue, even when all TX IDs were skipped and txIDs is empty.

   **Fix**: Add an early return in removeAndEnqueueTimeoutEvents when len(txIDs) == 0, skipping the write when there are
   no remaining TX IDs to report.

cendhu:
#### Type of change

- Bug fix
 
#### Description

When a timeout callback is ordered after a status delivery, the status handler removes all TX IDs from the subscription map first. The subsequent timeout handler then iterates over the request's TX IDs but finds none still subscribed, resulting in an empty TimeoutTxIds list being written to the response channel. Add an early return when no TX IDs remain to skip the write, and add a test covering this event ordering.

#### Related issues

 - resolves #601

### DIFF
--- a/service/sidecar/notify.go
+++ b/service/sidecar/notify.go
@@ -274,13 +274,12 @@ func (m *subscriptions) removeAndEnqueueTimeoutEvents(
 		}
 	}
 	// txIDs can be empty when the timeout fires after a status event has already
-	// removed all of this request's subscriptions (e.g., removeAndEnqueueStatusEvents
+	// removed all of this request's subscriptions (i.e., removeAndEnqueueStatusEvents
 	// ran first). In that case there is nothing to report, so we skip the write.
-	if len(txIDs) == 0 {
-		return 0, 0
+	if len(txIDs) > 0 {
+		req.streamEventQueue.Write(&committerpb.NotificationResponse{
+			TimeoutTxIds: txIDs,
+		})
 	}
-	req.streamEventQueue.Write(&committerpb.NotificationResponse{
-		TimeoutTxIds: txIDs,
-	})
 	return pendingTxIDsRemoved, uniquePendingTxIDsRemoved
 }
--- a/service/sidecar/notify.go
+++ b/service/sidecar/notify.go
@@ -273,6 +273,12 @@ func (m *subscriptions) removeAndEnqueueTimeoutEvents(
 			delete(requests, req)
 		}
 	}
+	// txIDs can be empty when the timeout fires after a status event has already
+	// removed all of this request's subscriptions (e.g., removeAndEnqueueStatusEvents
+	// ran first). In that case there is nothing to report, so we skip the write.
+	if len(txIDs) == 0 {
+		return 0, 0
+	}
 	req.streamEventQueue.Write(&committerpb.NotificationResponse{
 		TimeoutTxIds: txIDs,
 	})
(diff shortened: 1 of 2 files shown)


service/sidecar/notify.go
```diff
@@ -273,6 +273,12 @@ func (m *subscriptions) removeAndEnqueueTimeoutEvents(
 			delete(requests, req)
 		}
 	}
+	// txIDs can be empty when the timeout fires after a status event has already
+	// removed all of this request's subscriptions (e.g., removeAndEnqueueStatusEvents
```
liran-funaro: nit: These is only one way this could happen.
```suggestion
	// removed all of this request's subscriptions (i.e., removeAndEnqueueStatusEvents
```

service/sidecar/notify.go
liran-funaro: minor: While it is true that in this case, these values should be zero. Having a dedicated return path seems confusing.

I suggest the following pattern

```go
	if len(txIDs) > 0 {
		req.streamEventQueue.Write(&committerpb.NotificationResponse{TimeoutTxIds: txIDs})
	}

	return pendingTxIDsRemoved, uniquePendingTxIDsRemoved
```

service/sidecar/notify_test.go
```diff
@@ -475,6 +475,69 @@ func TestNotifierGlobalLimit(t *testing.T) {
...
+		txIDToRequests: make(map[string]map[*notificationRequest]any),
+		availableSlots: 1,
+	}
+
+	// Subscribe: the request is now tracked in the map under "tx1".
+	rejected, uniqueNew := subs.addRequest(req)
+	require.Empty(t, rejected)
+	require.Equal(t, 1, uniqueNew)
+
+	// Simulate the race: the timeout timer has already queued this request for
```
liran-funaro: nit: Using the term race can be confusing as we use "race" to describe memory access race-condition.
This is an event ordering race.

liran-funaro (approved):
Using the term race can be confusing as we use "race" to describe memory access race-condition.
This is an event ordering race.

We should also test for the opposite order:
1. The timer put a timeout event in the queue
2. The status event was processed
3. The timeout event is processed

This scenario is already handled well in the code, but we should do a similar simulation for this scenario.

cendhu:
@liran-funaro Updated the test.

service/sidecar/notify_test.go
```diff
@@ -475,6 +475,104 @@ func TestNotifierGlobalLimit(t *testing.T) {
...
+			if tc.queueTimeoutBeforeStatus {
+				timeoutReq = req
+			}
+
+			removeAndRequireStatusEvent(t, subs, responses)
+
+			if !tc.queueTimeoutBeforeStatus {
+				timeoutReq = req
+			}
+			removeAndRequireNoTimeoutEvent(t, subs, timeoutReq, responses)
```
liran-funaro: This flow doesn't make sense. It is equivalent to:
```suggestion
			subs, req, responses := newSingleTxSubscription(t)
			removeAndRequireStatusEvent(t, subs, responses)
			removeAndRequireNoTimeoutEvent(t, subs, req, responses)
```

cendhu: Done. I inlined the function too.

PR #605, opened 2026-05-21
[coordinator] Fix numWaitingTxsForStatus not counting rejected transactions

### LINKED ISSUE #604: [coordinator] numWaitingTxsForStatus counter goes negative when block contains rejected transactions
**Problem**

   The coordinator's numWaitingTxsForStatus atomic counter only increments by len(blk.Txs), ignoring
   len(blk.Rejected). When a block contains rejected transactions, sendTxStatus decrements the counter by the total
   number of statuses returned by the VC service (which includes rejected txs), but the counter was never
   incremented for those rejected txs. This causes the counter to go negative, breaking the
   NumberOfWaitingTransactionsForStatus gRPC API and any callers that rely on it to determine when all transactions
   have been processed.

   **Root Cause**

   In receiveAndProcessBlock (service/coordinator/coordinator.go:346):

   ```go
     promutil.AddToCounter(c.metrics.transactionReceivedTotal, len(blk.Txs)+len(blk.Rejected))
     c.numWaitingTxsForStatus.Add(int32(len(blk.Txs)))
   ```

   The metrics counter correctly accounts for both normal and rejected txs, but numWaitingTxsForStatus only counts
   blk.Txs. Since rejected txs flow through the VC service and produce status responses (via PrelimInvalidTxStatus),
   sendTxStatus decrements for them, creating a mismatch.

   **Fix**

   Change the increment to include rejected transactions:

   ```go
     c.numWaitingTxsForStatus.Add(int32(len(blk.Txs) + len(blk.Rejected)))
   ```

   This matches the metrics line above and ensures the counter correctly returns to zero after all transaction
   statuses (normal + rejected) are delivered to the sidecar.

cendhu:
#### Type of change

- Bug fix
 
#### Description

**Problem**:
The coordinator numWaitingTxsForStatus counter only incremented by len(blk.Txs), ignoring len(blk.Rejected). When a block contained rejected transactions, the counter was never decremented back to zero because sendTxStatus decrements by the actual number of statuses returned (which includes rejected txs). This caused the waiting transaction count to go negative, breaking the
NumberOfWaitingTransactionsForStatus API and any callers that rely on it to determine when all transactions have been processed.

**Root cause**:
In receiveAndProcessBlock, the counter increment only used len(blk.Txs) without len(blk.Rejected), despite the metrics counter on the line above correctly including both via len(blk.Txs)+len(blk.Rejected).

**Fix**:
Change numWaitingTxsForStatus to add len(blk.Txs) + len(blk.Rejected), matching the metrics and ensuring the counter is correctly decremented to zero when all transaction statuses (including rejected) are sent back to the sidecar.

Added TestWaitingTxsCountReturnsToZeroForBlockWithRejectedTxs to verify the counter returns to zero after processing a block that contains both normal and rejected transactions.

#### Related issues

  - resolves #604

### DIFF
--- a/service/coordinator/coordinator.go
+++ b/service/coordinator/coordinator.go
@@ -343,7 +343,7 @@ func (c *Service) receiveAndProcessBlock(
 		// still be forwarded to downstream components for complete processing.
 
 		promutil.AddToCounter(c.metrics.transactionReceivedTotal, len(blk.Txs)+len(blk.Rejected))
-		c.numWaitingTxsForStatus.Add(int32(len(blk.Txs))) //nolint:gosec
+		c.numWaitingTxsForStatus.Add(int32(len(blk.Txs) + len(blk.Rejected))) //nolint:gosec
 
 		if len(blk.Txs) > 0 {
 			// TODO: make it configurable.
--- a/service/coordinator/coordinator_test.go
+++ b/service/coordinator/coordinator_test.go
@@ -929,6 +929,33 @@ func TestChunkSizeSentForDepGraph(t *testing.T) {
 	}, 4*time.Second, 100*time.Millisecond)
 }
 
+func TestWaitingTxsCountReturnsToZeroForBlockWithRejectedTxs(t *testing.T) {
+	t.Parallel()
+	env := newCoordinatorTestEnv(t, &testConfig{numSigService: 1, numVcService: 1, mockVcService: true})
+
+	ctx, cancel := context.WithTimeout(t.Context(), 2*time.Minute)
+	t.Cleanup(cancel)
+	env.startServiceAndOpenStream(ctx, t)
+	env.createNamespaces(t, 0, "1")
+
+	b, expectedTxsStatus := makeTestBlock(1)
+	rejectedStatus := committerpb.NewTxStatus(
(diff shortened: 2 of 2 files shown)


liran-funaro (approved):
LGTM.
Unrelated to this PR, line 283 has an unexplained behaviour regarding `numWaitingTxsForStatus`.
```go
	return &servicepb.WaitingTransactions{
		Count: c.numWaitingTxsForStatus.Load() - int32(len(c.queues.vcServiceToCoordinatorTxStatus)), //nolint:gosec
	}, nil
```

Can you please add the comment explaining this?
This seems to mix units. The left is \#TXs and the right is \#TX-Batches.

cendhu:
> LGTM. Unrelated to this PR, line 283 has an unexplained behaviour regarding `numWaitingTxsForStatus`.
> 
> ```go
> 	return &servicepb.WaitingTransactions{
> 		Count: c.numWaitingTxsForStatus.Load() - int32(len(c.queues.vcServiceToCoordinatorTxStatus)), //nolint:gosec
> 	}, nil
> ```
> 
> Can you please add the comment explaining this? This seems to mix units. The left is #TXs and the right is #TX-Batches.

Good catch. This is a bug https://github.com/hyperledger/fabric-x-committer/issues/612

I will address it in a separate PR.

PR #607, opened 2026-05-21
[vc] Fix timer reset to drain stale ticks before Reset

### LINKED ISSUE #606: [vc] Timer.Reset called without draining stale ticks, causing spurious immediate firings
**Problem**

   In batchReceivedTransactionsAndForwardForProcessing, sendLargeBatch deferred
   timer.Reset(vc.timeoutForMinTxBatchSize) directly. Per Go docs (https://pkg.go.dev/time#Timer.Reset), Reset
   should only be called on stopped or expired timers whose channel has been drained. If the timer has already fired
   and the channel contains a stale tick, Reset does not drain it — the old tick remains in the channel and is read
   immediately on the next select iteration, causing a spurious batch flush.

   **Root Cause**

   time.Timer.Reset() does not drain the timer's channel. When the timer has already expired and a tick is pending
   in the channel, calling Reset starts the new period but leaves the stale tick in place. On the next select { case
   <-timer.C: ... }, the stale tick is consumed immediately, triggering an unwanted sendLargeBatch() call with an
   empty or partial batch.

   **Fix**

   Extracted a resetBatchTimer helper that:
   1. Calls timer.Stop() — if it returns false, the timer already expired
   2. Drains any stale tick from the channel with select { case <-timer.C: default: }
   3. Calls timer.Reset(timeout)

   Replaced the bare timer.Reset(vc.timeoutForMinTxBatchSize) call in sendLargeBatch with resetBatchTimer(timer,
   vc.timeoutForMinTxBatchSize).

cendhu:
#### Type of change

- Bug fix
 
#### Description

Use resetBatchTimer helper instead of bare timer.Reset in batchReceivedTransactionsAndForwardForProcessing. Per Go docs, Reset must only be called on stopped/expired timers with the channel drained; otherwise a stale tick may fire immediately.

Place resetBatchTimer after its caller (caller before callee) and add a reasoning comment. Simplify the test by replacing the cap(timer.C) check with a brief sleep to let the nanosecond timer expire, and use t.Cleanup for timer stop.

#### Related issues

 - resolves #606

### DIFF
--- a/service/vc/validator_committer_service.go
+++ b/service/vc/validator_committer_service.go
@@ -317,7 +317,7 @@ func (vc *ValidatorCommitterService) batchReceivedTransactionsAndForwardForProce
 	toPrepareTxs := channel.NewWriter(ctx, vc.toPrepareTxs)
 
 	sendLargeBatch := func() {
-		defer timer.Reset(vc.timeoutForMinTxBatchSize)
+		defer resetBatchTimer(timer, vc.timeoutForMinTxBatchSize)
 		if len(largerBatch.Transactions) == 0 {
 			return
 		}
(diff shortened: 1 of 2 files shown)


liran-funaro (commented):
This is not an issue since Go 1.23. From Go docs:
```
// For a chan-based timer created with NewTimer, as of Go 1.23,
// any receive from t.C after Reset has returned is guaranteed not
// to receive a time value corresponding to the previous timer settings;
// if the program has not received from t.C already and the timer is
// running, Reset is guaranteed to return true.
// Before Go 1.23, the only safe way to use Reset was to call [Timer.Stop]
// and explicitly drain the timer first.
// See the [NewTimer] documentation for more details.
```

cendhu:
Closing this PR due to the above stated reason.

liran-funaro:
#606 should also be closed? Or does it address a different issue?

cendhu:
> #606 should also be closed? Or does it address a different issue?

Yes, I have closed it.

PR #609, opened 2026-05-25
[coordinator] Introduce a unified service manager

### LINKED ISSUE #611: [coordinator] Duplicate implementation in VC an Verifier manager
The code for both managers is largely duplicated, with only minor adaptations between them.
Both managers provide the same core functionality.

This issue proposes extracting the shared logic in order to:

* Eliminate code duplication
* Reduce the overall code size and maintenance overhead
* Enable bug fixes (such as #608) to be implemented in a single place instead of multiple locations
* Improve code clarity by separating management logic from service-specific logic
* Simplify future enhancements by centralizing common behavior
* Improve maintainability and reduce the likelihood of behavioral drift between the two managers over time

It is also worth noting that the current implementation was originally approved under time constraints.
Under normal circumstances, code with this level of duplication and mixed responsibilities would likely not be accepted as-is.

In addition, the VC manager currently has an extra responsibility: storing the shared, load-balanced VC connection.
This issue proposes moving that responsibility into a separate class, since it is unrelated to task management and violates separation of concerns.

liran-funaro:
#### Type of change

#### Description

This commit introduces a unified service manager.
This way, we avoid cases where one instance is fixed while the other remains buggy.

#### Related issues

- resolves #611

### DIFF
--- a/docs/coordinator-block-diagram.md
+++ b/docs/coordinator-block-diagram.md
@@ -78,11 +78,13 @@ Inside coordinator, control is split across five main components:
 
 - `Service` owns external gRPC surface and stream lifecycle.
 - `dependencygraph.Manager` decides which transactions are free to run.
-- `signatureVerifierManager` talks to Verifier service pool.
-- `validatorCommitterManager` talks to Validator-Committer pool and sends final results to two consumers.
+- The verifier `serviceManager` talks to Verifier service pool.
+- The validator-committer `serviceManager` talks to Validator-Committer pool and sends final results to two consumers.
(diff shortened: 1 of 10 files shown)


cendhu:
I am not comfortable with adding more abstractions to the code. Bug can be fixed in the current flow. We need to discuss on the abstraction such as servicemanager.

liran-funaro:
> I am not comfortable with adding more abstractions to the code. Bug can be fixed in the current flow. We need to discuss on the abstraction such as servicemanager.

I added an issue to address this. Please review it.
- #611

PR #610, opened 2026-05-25
[coordinator-test] Remove the need for a real VC

### LINKED ISSUE #93: [coordinator test] Use mock VC for all coordinator tests
Currently, the mock VC is missing some functionality required for certain coordinator tests.
This issue suggests improving the mock VC and removing the need to run a real VC in the coordinator tests.

### LINKED ISSUE #187: [coordinator test] Enable verify keys in the mock VC
This allows using the mock VC in the following test
- TestCoordinatorServiceDependentOrderedTxs
- TestCoordinatorRecovery
- TestCoordinatorServiceDependentOrderedTxs

liran-funaro:
#### Type of change

- Test update
 
#### Description

- Add support for injecting predefined TX status
- Keep tx execution order for validation in tests
- Remove the use of a real VC and DB in the coordinator testing 

#### Related issues

- resolves #187 
- resolves #93


cendhu (commented):
I am fine with this change but add unit-test for this mock vcservice as it has some core logic such as MVCC.

cendhu:
@liran-funaro Not sure whether you missed my comment. Please add unit-test for this mock vcservice.

liran-funaro:
@cendhu Added tests

mock/vcservice.go
```diff
@@ -218,3 +219,123 @@ func (v *VcService) SubmitTransactions(ctx context.Context, txsBatch *servicepb.
...
+
+	// We simulate a faulty node by not responding to the first X TXs.
+	skip := max(0, min(v.MockFaultyNodeDropSize, len(txs)))
+
+	v.worldStateMu.Lock()
+	defer v.worldStateMu.Unlock()
+	for _, tx := range txs[skip:] {
+		code := v.validate(tx)
+		if code == committerpb.Status_COMMITTED {
+			for _, w := range getWrites(tx) {
```
cendhu: validate() does not differentiate between duplicate and non-duplicate. As a result, we reapply the writes even for duplicate transactions. This should have been detected by the test in coordinator. If not, we may be not be checking the version.

mock/vcservice.go
```diff
@@ -218,3 +219,123 @@ func (v *VcService) SubmitTransactions(ctx context.Context, txsBatch *servicepb.
...
+			return committerpb.Status_ABORTED_MVCC_CONFLICT
+		}
+	}
+
+	return code
+}
+
+func getReads(tx *servicepb.VcTx) (reads []read) {
+	for _, ns := range tx.Namespaces {
+		if ns.NsId != committerpb.MetaNamespaceID && ns.NsId != committerpb.ConfigNamespaceID {
```
cendhu: For metaNamespaceID, we need to add the config version to the read dependency.


```go
				switch nsOperations.NsId {
				case committerpb.MetaNamespaceID:
					// Meta TX is dependent on the config TX.
					prepTxs.addReadsOnly(tID, &applicationpb.TxNamespace{
						NsId: committerpb.ConfigNamespaceID,
						ReadsOnly: []*applicationpb.Read{{
							Key:     []byte(committerpb.ConfigKey),
							Version: &nsOperations.NsVersion,
						}},
					})
				case committerpb.ConfigNamespaceID:
					// A config TX is independent.
				default:
					prepTxs.addReadsOnly(tID, &applicationpb.TxNamespace{
						NsId: committerpb.MetaNamespaceID,
						ReadsOnly: []*applicationpb.Read{{
							Key:     []byte(nsOperations.NsId),
							Version: &nsOperations.NsVersion,
						}},
					})
				}
```

cendhu (commented):
We need to address two issues with this approach:

1. Logic Duplication & Correctness: We are duplicating the core logic of vcservice here, which makes me question if the trade-off is worth it. It already allowed two bugs to slip past our coordinator and mock tests. Since correctness is strictly more important than reducing test runtime, we should evaluate these changes carefully to avoid introducing hard-to-detect bugs.
2. Future Maintenance: When we implement snapshots and checkpoints, I'm concerned we will have to add even more complexity to this mock.

As a general rule, mocks should be kept as simple (or 'dumb') as possible. While I supported this change earlier, I am skeptical about it now.

liran-funaro:
@cendhu I understand your concern. The motivation is not performance, but test encapsulation.
We don't want VC issues to fail the coordinator.
The alternative is to inject responses manually without really doing the validation. Also, injecting the world-state manually.
This will reduce significantly the mock complexity and can arguably make the tests more explicit as the test will explicitly set the VC behaviour.

cendhu:
> The alternative is to inject responses manually without really doing the validation. Also, injecting the world-state manually. This will reduce significantly the mock complexity and can arguably make the tests more explicit as the test will explicitly set the VC behaviour.

Yes, this sounds interesting and similar to how most mock frameworks work. How can we approach this?

liran-funaro:
@cendhu I modified the PR according to the discussion in this PR

mock/vcservice_test.go
```diff
@@ -0,0 +1,453 @@
...
+	vc      *VcService
+	servers *test.Servers
+	conns   []*grpc.ClientConn
+	clients []servicepb.ValidationAndCommitServiceClient
+	streams []grpc.BidiStreamingClient[servicepb.VcBatch, committerpb.TxStatusBatch]
+}
+
+const testNS = "ns1"
+
+func startVCTestEnv(t *testing.T, p test.StartServerParameters) *mockVCTestEnv {
```
cendhu: Usually, we name it newVCTestEnv.

mock/vcservice_test.go
```diff
@@ -0,0 +1,453 @@
...
+			tx := &servicepb.VcTx{
+				Ref: committerpb.NewTxRef("batch-tx-"+string(rune(i)), uint64(100+i), 0),
+				Namespaces: []*applicationpb.TxNamespace{{
+					NsId: testNS,
+					BlindWrites: []*applicationpb.Write{
+						{Key: []byte("batch-key"), Value: []byte("batch-value")},
+					},
+				}},
+			}
+			e.sendTX(t, tx, committerpb.Status_COMMITTED)
```
cendhu: Should we prefix sendTx and sendBatchTx with require? Initially, I was wondering why we send and not check anything. After looking at the sendTx and sendBatchTx, I understood it does require too.

mock/vcservice.go
```diff
@@ -218,3 +212,50 @@ func (v *VcService) SubmitTransactions(ctx context.Context, txsBatch *servicepb.
 	channel.NewWriter(ctx, s.q).Write(txsBatch)
 	return nil
 }
+
+// SetTxStatus injects the status the mock will return for the TX with the given ref.
+// This lets a test declare the VC's outcome explicitly (e.g. ABORTED_MVCC_CONFLICT or
+// REJECTED_DUPLICATE_TX_ID) instead of having the mock compute it. A TX with no injected
+// status defaults to COMMITTED (unless it carries a PrelimInvalidTxStatus).
```
cendhu, line 211: Can we mention the priority order? PrelimInvalidTxStatus takes precedence over others, etc... This should be documented. I am also worried whether we would miss this in future and make mistakes.

mock/vcservice_test.go
```diff
@@ -0,0 +1,453 @@
...
+	e.sendTX(t, &servicepb.VcTx{
+		Ref:                   prelimRef,
+		PrelimInvalidTxStatus: new(committerpb.Status_MALFORMED_NO_WRITES),
+		Namespaces: []*applicationpb.TxNamespace{{
+			NsId:      testNS,
+			ReadsOnly: []*applicationpb.Read{{Key: []byte("key3")}},
+		}},
+	}, committerpb.Status_MALFORMED_NO_WRITES)
+
+	t.Log("Case 4: Duplicate transaction number detection (error, not a status)")
```
cendhu: Is this possible? If not, can we remove this subtest?

cendhu (approved):
LGTM. This read much simpler. Thanks for making this change.

mock/vcservice_test.go
liran-funaro: I don't think this can happen in our real coordinator if the blocks are correct. But the mock VC checks this, so we need to test this. We can remove this check and this test. What do you think?

cendhu: Yes, we can remove this.

PR #613, opened 2026-05-26
[Coordinator] fix waiting status count

### LINKED ISSUE #612: [coordinator] waiting transaction count subtracts status batches instead of statuses
### Problem

  `NumberOfWaitingTransactionsForStatus` mixes units when calculating waiting transaction count:

   ```go
     Count: c.numWaitingTxsForStatus.Load() - int32(len(c.queues.vcServiceToCoordinatorTxStatus))
   ```

   numWaitingTxsForStatus counts transactions, but len(vcServiceToCoordinatorTxStatus) counts queued TxStatusBatch objects. A single TxStatusBatch can contain multiple transaction statuses.

   This can cause the API to overreport waiting transactions during recovery/backlog scenarios.

   ### Example

   If:

   - numWaitingTxsForStatus = 7
   - vcServiceToCoordinatorTxStatus has 2 queued batches
   - those batches contain 5 total transaction statuses

   Expected waiting count:

   ```text
     7 - 5 = 2
   ```

   Current code returns:

   ```text
     7 - 2 = 5
   ```

   ### When this happens

   Common path hides this because NumberOfWaitingTransactionsForStatus returns error while BlockProcessing stream is active.

   Bug appears when:

   1. sidecar/coordinator stream becomes inactive
   2. VC finishes processing transactions
   3. status batches are queued in vcServiceToCoordinatorTxStatus
   4. sidecar calls NumberOfWaitingTransactionsForStatus before reopening stream
   5. one or more queued batches contain multiple statuses

   ### Suggested fix

   Wrap vcServiceToCoordinatorTxStatus channel in a concrete queue struct that tracks total queued transaction statuses with an atomic counter.

   Do not scan/drain the channel to count statuses. Reading from channel consumes messages and can reorder or lose delivery.

   Possible shape:

   ```go
     type txStatusQueue struct {
         ch    chan *committerpb.TxStatusBatch
         count atomic.Int32
     }
   ```

   The queue should:

   - increment count when enqueueing a status batch
   - rollback count if enqueue fails
   - decrement count when dequeueing a status batch
   - expose readyCount() for NumberOfWaitingTransactionsForStatus

   Then API should subtract queued status count, not queued batch count.

   ### Test

   Add test covering:

   - numWaitingTxsForStatus = 7
   - queue has 2 batches containing 5 total statuses
   - NumberOfWaitingTransactionsForStatus returns 2, not 5
   ```

cendhu:
#### Type of change

- Bug fix
 
#### Description

Track VC-produced transaction statuses separately from the status queue length so NumberOfWaitingTransactionsForStatus does not count statuses that are already ready to be sent back to the sidecar.

Use a txStatusQueue to count statuses before enqueueing and roll the count back if a blocked enqueue is canceled. This avoids a race where a fast reader could decrement the count before the writer increments it.

Simplify queue handling by removing impossible nil/closed-channel cases and keep tests focused on count accounting and context cancellation.

#### Related issues

 - resolves #612

### DIFF
--- a/docs/coordinator.md
+++ b/docs/coordinator.md
@@ -186,7 +186,7 @@ GetConfigTransaction(ctx context.Context, in *Empty, opts ...grpc.CallOption) (*
   * This API retrieves the latest system configuration transaction by querying a Validator-Committer service, which reads it from the database.
 
 ```go
-NumberOfWaitingTransactionsForStatus(ctx context.Context, in *Empty, opts ...grpc.CallOption) (*WaitingTransactions, error)
+NoPendingTransactionProcessing(ctx context.Context, in *Empty, opts ...grpc.CallOption) (*wrapperspb.BoolValue, error)
 ```
   * This API returns the number of transactions that are currently in the pipeline awaiting a final status.
 
--- a/docs/coordinator.md
+++ b/docs/coordinator.md
@@ -188,7 +188,7 @@ GetConfigTransaction(ctx context.Context, in *Empty, opts ...grpc.CallOption) (*
 ```go
 NoPendingTransactionProcessing(ctx context.Context, in *Empty, opts ...grpc.CallOption) (*wrapperspb.BoolValue, error)
 ```
-  * This API returns the number of transactions that are currently in the pipeline awaiting a final status.
+  * This API returns true when all previously submitted transactions have been processed.
 
 ## 7. Failure and Recovery
 
--- a/docs/sidecar-block-diagram.md
+++ b/docs/sidecar-block-diagram.md
@@ -375,7 +375,7 @@ When the sidecar restarts after a crash, it follows a systematic recovery proces
    └─► Initialize TLS configuration
 
 2. Wait for Coordinator to Become Idle
-   ├─► Query: NumberOfWaitingTransactionsForStatus()
+   ├─► Query: NoPendingTransactionProcessing()
    ├─► Wait until count == 0
    └─► Ensures all in-flight transactions are processed
 
--- a/docs/sidecar-block-diagram.md
+++ b/docs/sidecar-block-diagram.md
@@ -376,7 +376,7 @@ When the sidecar restarts after a crash, it follows a systematic recovery proces
 
 2. Wait for Coordinator to Become Idle
    ├─► Query: NoPendingTransactionProcessing()
-   ├─► Wait until count == 0
+   ├─► Wait until true
    └─► Ensures all in-flight transactions are processed
 
 3. Query Coordinator State
--- a/service/coordinator/coordinator.go
+++ b/service/coordinator/coordinator.go
@@ -86,10 +86,10 @@ type (
 		//       rename vcServiceToDepGraphValidatedTxs to committedOrAbortedTxsNode.
 		vcServiceToDepGraphValidatedTxs chan dependencygraph.TxNodeBatch
 
-		// sender: validator committer manager sends transaction status to this channel. For each validator committer
-		// 	       server, there is a goroutine that sends transaction status to this channel.
-		// receiver: coordinator receives transaction status from this channel and forwards them to the sidecar.
-		vcServiceToCoordinatorTxStatus chan *committerpb.TxStatusBatch
+		// sender: validator committer manager sends transaction status to this queue. For each validator committer
+		// 	       server, there is a goroutine that sends transaction status to this queue.
+		// receiver: coordinator receives transaction status from this queue and forwards them to the sidecar.
+		vcServiceToCoordinatorTxStatus *txStatusQueue
 	}
 )
 
@@ -122,7 +122,7 @@ func NewCoordinatorService(c *Config) *Service {
 		depGraphToSigVerifierFreeTxs:       make(chan dependencygraph.TxNodeBatch, bufSzPerChanForValCommitMgr),
 		sigVerifierToVCServiceValidatedTxs: make(chan dependencygraph.TxNodeBatch, bufSzPerChanForSignVerifierMgr),
 		vcServiceToDepGraphValidatedTxs:    make(chan dependencygraph.TxNodeBatch, bufSzPerChanForValCommitMgr),
-		vcServiceToCoordinatorTxStatus:     make(chan *committerpb.TxStatusBatch, bufSzPerChanForValCommitMgr),
+		vcServiceToCoordinatorTxStatus:     newTxStatusQueue(bufSzPerChanForValCommitMgr),
 	}
 
 	metrics := newPerformanceMetrics()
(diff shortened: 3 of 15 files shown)


service/coordinator/coordinator.go
```diff
@@ -280,7 +280,7 @@ func (c *Service) NumberOfWaitingTransactionsForStatus(
 	defer c.streamActive.Unlock()
 
 	return &servicepb.WaitingTransactions{
-		Count: c.numWaitingTxsForStatus.Load() - int32(len(c.queues.vcServiceToCoordinatorTxStatus)), //nolint:gosec
+		Count: c.numWaitingTxsForStatus.Load() - c.queues.vcServiceToCoordinatorTxStatus.readyCount(),
```
liran-funaro: major: We still have a race condition when reading each value, leading to an incoherent state.
I still miss a comment here explaining why we need to decrease `vcServiceToCoordinatorTxStatus` from `numWaitingTxsForStatus`.
Can't we wait for the coordinator to finish processing all the TXs, and then `numWaitingTxsForStatus` would be zero?
We don't even need `NumberOfWaitingTransactionsForStatus` in practice. We just need `IsNoWaitingTransactionsForStatus` method.

liran-funaro (commented):
I can't understand the reasoning of the following line, so I can't properly review this PR.
Please clarify.
```go
c.numWaitingTxsForStatus.Load() - c.queues.vcServiceToCoordinatorTxStatus.readyCount(),
```

service/coordinator/coordinator.go
cendhu: I have added a comment. Yes, I agree that we can have a `NoPendingTransactionProcessing()`. Initial goal of introducing this waiting transactions count was for observability but we didn't utilize that.

cendhu: I can do the following:

1. Rename `numWaitingTxsForStatus` to `numTxsInProgress` 
2. Rename `NumberOfWaitingTransactionForStatus` to `NoPendingTransactionProcessing() bool`

@liran-funaro If you agree on this, I can make an update to this PR.

liran-funaro: Sounds good. But I still don't understand the claim that the result cannot be negative. Because we don't read the two values atomically, isn't it possible to reach a negative value?

syndbg: Out of curiosity commenting. Isn't this a good opportunity to re-think the approach of using two atomics to derive the state?

`WaitingTransactions.count` is exposed through the coordinator proto, but it is derived from `numWaitingTxsForStatus -
  vcServiceToCoordinatorTxStatus.readyCount()`. This assumes `readyCount()` is always a subset of `numWaitingTxsForStatus`: statuses already produced by VC but not yet sent to the sidecar.

That invariant is not enforced by the two atomics themselves. Atomics only make each counter race-free; they do not guarantee the combined state is coherent. The calculation is only expected to be safe because this RPC holds `streamActive`, so no active stream can concurrently receive new blocks or
  drain status batches, at least from what I see.


Could we move this into a named helper/tracker that documents and enforces the invariant, and defensively handle `readyCount() > numWaitingTxsForStatus`? Returning a negative `WaitingTransactions.count` over the public API would be misleading; silently clamping is also bad (i.e   `Count: max(0, waiting-ready)`)
 since it would also hide an invariant violation.


That protects the proto response from a negative value, but it also hides the fact that the coordinator state became inconsistent. If ready > waiting, VC
  has apparently produced more statuses than the coordinator believes are outstanding. That should be observable, because it points to a lifecycle/counting
  bug.

  **Simpler design**: keep the proto as-is, but move the logic into one internal method that owns the invariant.

```
  func (c *Service) waitingTransactionsForStatusCount() int32 {
  	waiting := c.numWaitingTxsForStatus.Load()
  	ready := c.queues.vcServiceToCoordinatorTxStatus.readyCount()

  	if ready > waiting {
  		logger.Errorf("inconsistent waiting status counters: waiting=%d ready=%d", waiting, ready)
  		return 0
  	}

  	return waiting - ready
  }
```

  Then:

```
  return &servicepb.WaitingTransactions{
  	Count: c.waitingTransactionsForStatusCount(),
  }, nil
```

Note: It would enable more analysis how and why the state gets inconsistent, but it's a start at least.

WDYT?

liran-funaro: After reviewing the code, it is clear that `numWaitingTxsForStatus` can only be updated when the stream is active. Thus, it is expected to be constant.
Thus, we have no race condition.
`vcServiceToCoordinatorTxStatus` can grow as the Coordinator receives responses from the VC, but cannot decrease as the stream is closed, so there is no puller from this queue.

This should be documented in the method. We used to know that when we wrote this code :)

Another note is that if `vcServiceToCoordinatorTxStatus` cap is not larger than `numWaitingTxsForStatus`, we will reach a deadlock in the sidecar.
We should address this. Maybe in a different PR.

service/coordinator/coordinator.go
```diff
@@ -279,8 +279,19 @@ func (c *Service) NumberOfWaitingTransactionsForStatus(
 	}
 	defer c.streamActive.Unlock()
 
+	// When the sidecar stream is inactive, numWaitingTxsForStatus is not decremented
```
liran-funaro: I suggest broadening the statement
```suggestion
	// When the sidecar stream is inactive, numWaitingTxsForStatus is not updated
```

service/coordinator/coordinator.go
cendhu: @liran-funaro Sure, will add the reasoning. 

@syndbg Both are atomic types. Waiting would always be higher or equal to the ready.

docs/coordinator.md
```diff
@@ -186,7 +186,7 @@ GetConfigTransaction(ctx context.Context, in *Empty, opts ...grpc.CallOption) (*
   * This API retrieves the latest system configuration transaction by querying a Validator-Committer service, which reads it from the database.
 
 ```go
-NumberOfWaitingTransactionsForStatus(ctx context.Context, in *Empty, opts ...grpc.CallOption) (*WaitingTransactions, error)
+NoPendingTransactionProcessing(ctx context.Context, in *Empty, opts ...grpc.CallOption) (*wrapperspb.BoolValue, error)
 ```
   * This API returns the number of transactions that are currently in the pipeline awaiting a final status.
```
liran-funaro: Fix description

docs/sidecar-block-diagram.md
```diff
@@ -375,7 +375,7 @@ When the sidecar restarts after a crash, it follows a systematic recovery proces
    └─► Initialize TLS configuration
 
 2. Wait for Coordinator to Become Idle
-   ├─► Query: NumberOfWaitingTransactionsForStatus()
+   ├─► Query: NoPendingTransactionProcessing()
    ├─► Wait until count == 0
```
liran-funaro:
```suggestion
   ├─► Wait until true

File: service/vc/validator_committer_service.go [hunk 5]
```diff
@@ -61,7 +61,7 @@ type ValidatorCommitterService struct {
 	// streams were allowed concurrently, transaction status might not reach the client
 	// reliably, as we are not currently associating requests and their corresponding responses
 	// (i.e., status updates) with specific streams.
-	// Further, when isStreamActive is active, NumberOfWaitingTransactionsForStatus would return an
+	// Further, when isStreamActive is active, NoPendingTransactionProcessing would return an
 	// error as this gRPC api can be called only when the stream is inactive.
```
liran-funaro: I think this part of the comment is out of place. It has nothing to do with the VC.

liran-funaro (approved):

service/vc/validator_committer_service.go
cendhu: Yes. I do not know how it got here. This was a pre-existing comment.

PR #614, opened 2026-05-27
[bob] Reduce initial context from 25K to 19K

liran-funaro:
#### Type of change

#### Description

This commit reduces Bob's initial context from 25K tokens to 19K.

- Moving the testing rules to a skill
- Removing some repeating lines


cendhu (approved):
We can also caveman skill to compress this even further.

PR #617, opened 2026-06-01
[BREAKING] Add CLI command for database initialization

### LINKED ISSUE #487: [db] Admin DB initialisation
Currently the sidecar initializes the DB by a cascading calls to the VC.
Desired: The admin should do this manually.
We should add a CMD `committer init-state-db`.

dean-amar:
#### Type of change

- New feature
- Improvement (code improvement)
- Test update
- Documentation update
- Breaking change
 
#### Description

- Introduces a new committer `init-db` CLI command that allows administrators to initialize the database (create system tables, metadata tables, and system namespaces) as a standalone operation before starting services. The change includes a **BREAKING** API modification: the SetupSystemTablesAndNamespaces RPC method has been removed from the `VC` service as database initialization is now performed via CLI instead of gRPC.

#### Additional details

- Usage example: ```committer init-db --config vc.yaml```

#### Related issues

- Resolves #487


cmd/cliutil/flags.go
```diff
@@ -7,10 +7,17 @@ SPDX-License-Identifier: Apache-2.0
...
-// SetDefaultFlags registers the --config flag for the cmd parameter.
+// SetDefaultFlags registers the --config flag on the given command.
 func SetDefaultFlags(cmd *cobra.Command, configPath *string) {
 	cmd.PersistentFlags().StringVarP(configPath, "config", "c", "", "set the config file path")
 }
+
+// SetDurationFlag registers a duration flag with the given name on the given command.
+func SetDurationFlag(cmd *cobra.Command, name string, duration *time.Duration) {
+	cmd.Flags().DurationVar(duration, name, 5*time.Minute, "Timeout for the initialization operation")
+}
```
liran-funaro: minor: We don't need a dedicated method for this

cmd/config/app_config.go
```diff
@@ -84,6 +91,17 @@ func ReadLoadGenYamlAndSetupLogging(
...
 
+// ReadDBInitYamlAndSetupLogging reading the YAML config file for database initialization.
+func ReadDBInitYamlAndSetupLogging(
+	v *viper.Viper, configPath string,
+) (*db.Config, *serve.Config, error) {
+	cfg, srvCfg, err := readYamlAndSetupLogging[databaseInitConfig](v, configPath)
+	if err != nil {
+		return nil, nil, err
+	}
+	return cfg.Database, srvCfg, nil
```
liran-funaro: minor: This method should not return `srvCfg`

utils/statedb/dbinit.go
```diff
@@ -77,17 +73,25 @@ func NewDatabasePool(ctx context.Context, config *DatabaseConfig) (*pgxpool.Pool
...
+// SetupSystemTablesAndNamespaces creates the required system tables and namespaces in the database.
+// This function should be called before operating against the database.
+// It is safe to run multiple times.
+// TODO: merge this file with vc/database.go.
+func SetupSystemTablesAndNamespaces(
+	ctx context.Context,
+	pool *pgxpool.Pool,
+	retryProfile *retry.Profile,
+	tablePreSplitTablets int,
+) error {
```
liran-funaro, line 86: This method can create the pool and include the call to GetTablePreSplitTablets().
So it will only accept db config.
```suggestion
func SetupSystemTblesAndNamespaces(
	ctx context.Context,
	config db.Config,
) error {
```

utils/db/dbinit.go
```diff
@@ -97,7 +101,8 @@ func (db *database) setupSystemTablesAndNamespaces(ctx context.Context) error {
 	return nil
 }
 
-func createNsTables(nsID string, tablets int, queryFunc func(q string) error) error {
+// CreateNsTables creates the table and functions for a namespace.
+func CreateNsTables(nsID string, tablets int, queryFunc func(q string) error) error {
```
liran-funaro: I suggest refactoring this method to return the query and let the caller execute the query.
```suggestion
func CreateNsTablesQuery(nsID string, tablets int) string {
```

cmd/config/samples/dbinit.yaml
```diff
@@ -0,0 +1,34 @@
...
+    # Path to the CA certificate file for verifying the database server's certificate.
+    ca-cert-path: /root/artifacts/peerOrganizations/peer-org-0.com/msp/tlscacerts/tlspeer-org-0-CA-cert.pem
+  # Maximum size of the database connection pool. Limits concurrent database operations.
+  # Should be tuned based on database capacity and expected transaction load.
+  # Default: 20
+  max-connections: 10
+  # Minimum size of the database connection pool. Keeps connections warm for faster operations.
+  # Should be less than or equal to max-connections.
+  # Default: 1
+  min-connections: 5
```
liran-funaro: The config should be changed to reflect the fact that we don't need many connections.
We should also modify the comment to explain this.
```suggestion
  # Maximum size of the database connection pool. Limits concurrent database operations.
  # Should be tuned based on database capacity and expected transaction load.
  # Default: 20
  max-connections: 10
  # Minimum size of the database connection pool. Keeps connections warm for faster operations.
  # Should be less than or equal to max-connections.
  # Default: 1
  min-connections: 1
```

cmd/config/viper.go
```diff
@@ -106,6 +107,13 @@ func NewViperWithOrdererDefaults() *viper.Viper {
 	return v
 }
 
+// NewViperWithDBInitDefaults returns a viper instance with the database initialization default values.
+func NewViperWithDBInitDefaults() *viper.Viper {
+	v := NewViperWithLoggingDefault("dbinit")
```
liran-funaro: This means that VC env variables overrides will not apply to DB init.

docker/images/test_node/Dockerfile
```diff
@@ -39,10 +40,13 @@ ENV SC_VC_DATABASE_DATABASE="postgres"
 ENV SC_QUERY_DATABASE_USERNAME="postgres"
 ENV SC_QUERY_DATABASE_PASSWORD="postgres"
 ENV SC_QUERY_DATABASE_DATABASE="postgres"
-
+ENV SC_DBINIT_DATABASE_USERNAME="postgres"
+ENV SC_DBINIT_DATABASE_PASSWORD="postgres"
+ENV SC_DBINIT_DATABASE_DATABASE="postgres"
```
liran-funaro: This can be avoided if we opt to use the VC config instead of a dedicated db-init config.

docker/images/test_node/run
```diff
@@ -59,6 +59,10 @@ if [[ $ops == *"loadgen"* ]]; then
   "$BINS_PATH/loadgen" start --config "$CONFIGS_PATH/loadgen.yaml" &
 fi
 
+if [[ $ops == *"init-db"* ]]; then
+  "$BINS_PATH/committer" init-db --config "$CONFIGS_PATH/dbinit.yaml"
+fi
```
liran-funaro: This should run before the committer.

docker/test/common.go
```diff
@@ -87,19 +104,42 @@ func createAndStartContainerAndItsLogs(
...
-	logs, err := dockerClient.ContainerLogs(ctx, resp.ID, container.LogsOptions{
+// runContainerToCompletion creates, starts, and waits for a one-shot container to finish,
+// asserting a zero exit code.
+func runContainerToCompletion(
+	ctx context.Context,
+	t *testing.T,
+	params createAndStartContainerParameters,
+) {
+	t.Helper()
+	dockerClient, id := createContainer(ctx, t, params)
```
liran-funaro: We can reuse the `createAndStartContainerAndItsLogs()`.
`createAndStartContainerAndItsLogs()` can return the statusCh.
Since this is only used by `runDatabaseInitWithReleaseImage()`, we can remove `runContainerToCompletion()` and embed the wait logic to runDatabaseInitWithReleaseImage()`.

docker/test/container_release_image_test.go
```diff
@@ -181,24 +184,57 @@ func startSecuredDatabaseNode(ctx context.Context, t *testing.T, params startNod
 	return conn.Password
 }
 
+// runDatabaseInitWithReleaseImage runs init-db command in a temporary container.
+func runDatabaseInitWithReleaseImage(ctx context.Context, t *testing.T, params startNodeParameters) {
```
liran-funaro: We should also test the case where we fail to initialize the DB using the container.
Run this command without an active DB and a short timeout.

docker/test/common.go
```diff
@@ -111,6 +151,25 @@ func createAndStartContainerAndItsLogs(
...
+
+	select {
+	case err := <-errCh:
+		require.NoError(t, err)
+	case status := <-statusCh:
+		require.Zero(t, status.StatusCode, "container %s failed with exit code %d", params.name, status.StatusCode)
+	}
+}
+
+func createContainer(
```
liran-funaro: After removing `runContainerToCompletion()`, there is no justification for this method.

service/vc/validator_committer_service_test.go
```diff
@@ -48,7 +49,7 @@ func TestVCSecureConnection(t *testing.T) {
 			return func(ctx context.Context, t *testing.T, cfg connection.TLSConfig) error {
 				t.Helper()
 				client := createValidatorAndCommitClientWithTLS(t, &env.ServerConfigs[0].GRPC.Endpoint, cfg)
-				_, err := client.SetupSystemTablesAndNamespaces(ctx, nil)
+				_, err := client.SetLastCommittedBlockNumber(ctx, nil)
```
liran-funaro: I don't think this is needed

service/vc/test_exports.go
```diff
@@ -176,6 +171,13 @@ func NewDatabaseTestEnvFromConnection(t *testing.T, cs *testdb.Connection, loadB
 	require.NoError(t, err, "%+v", err)
 	t.Cleanup(dbObject.close)
 
+	// Initialize the database with system tables and namespaces
+	tablets, err := db.GetTablePreSplitTablets(sCtx, dbObject.pool, config)
+	require.NoError(t, err, "failed to get table pre-split tablets: %+v", err)
+
+	err = db.SetupSystemTablesAndNamespaces(sCtx, dbObject.pool, config.Retry, tablets)
+	require.NoError(t, err, "failed to initialize database: %+v", err)
```
liran-funaro: unify to a single call

utils/db/dbinit.go
```diff
@@ -77,17 +73,25 @@ func NewDatabasePool(ctx context.Context, config *DatabaseConfig) (*pgxpool.Pool
...
 
-// TODO: merge this file with database.go.
-func (db *database) setupSystemTablesAndNamespaces(ctx context.Context) error {
-	logger.Info("Created tx status table, metadata table, and its methods.")
-	if execErr := retry.ExecuteSQL(ctx, db.retryProfile, db.pool,
-		fmtSplitIntoTablets(dbInitSQLStmt, db.tablePreSplitTablets)); execErr != nil {
+// SetupSystemTablesAndNamespaces creates the required system tables and namespaces in the database.
+// This function should be called before operating against the database.
+// It is safe to run multiple times.
+// TODO: merge this file with vc/database.go.
```
liran-funaro: This todo is obselete

cmd/cliutil/flags.go
dean-amar: Done.

cmd/config/app_config.go
dean-amar: Done.

utils/statedb/dbinit.go
dean-amar: Done.

docker/images/test_node/run
dean-amar: Done.

utils/db/dbinit.go
dean-amar: Done.

docker/test/common.go
dean-amar: Done.

docker/test/common.go
dean-amar: Done.

cmd/config/samples/dbinit.yaml
dean-amar: Done.

cmd/config/viper.go
dean-amar: That's true, but the VC and Query database configurations should match, and we explicitly set each of their ENV variables. The current implementation is intentional - DB init is an administrative operation that should have its own configuration namespace IMO.

docker/images/test_node/Dockerfile
dean-amar: Addressed in the comment above.

docker/test/container_release_image_test.go
dean-amar: Done.

service/vc/validator_committer_service_test.go
dean-amar: In the next commit, I changed this to a read-RPC.

liran-funaro: Then please document as a comment that this is the intention of this call, and it is not necessary for correct execution of the test.

cendhu (commented):
I recommend reusing the vc config for the db-init command, as it isn't a standalone component and relies on vc code anyway. We already follow this pattern by reusing the service config for health checks, so we can take the same approach here. Furthermore, since this command will be used very rarely (primarily during initial setup and occasional version upgrades), introducing a dedicated config is overkill and adds unnecessary code.

utils/db/config.go
```diff
@@ -0,0 +1,64 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package db
```
cendhu: The new package `db` is not an utility. It is a service. Hence, I would suggest to revert moving db to utility.

liran-funaro: We use the term "service" for a process that is deployed to run continuously and serve requests.
This `db` package does not satisfy this description.

liran-funaro: It is also used in the query service and the CMD package. So having it in the VC subtree is not optimal.
This actually belongs in a "common" package (like in Fabric). We used "utils" for that.
We can rename "utils" to "common", and solve this semantic issue.

dean-amar: I'll revert the change and move the `db` package back into utils.
If there is agreement to rename `utils` to `common`, I'll do it in a separate PR.

cmd/committer/db_init_cmd.go
```diff
@@ -0,0 +1,62 @@
...
+// databaseInitializationCMD creates the "init-db" command.
+func databaseInitializationCMD() *cobra.Command {
+	var configPath string
+	var timeout time.Duration
+	cmd := &cobra.Command{
+		Use:   initDBCommand,
+		Short: "Initialize the database with required tables and namespaces",
+		Long: `Initialize the state database by creating system tables, metadata tables,
+				and system namespaces (meta and config). This is a one-time administrative
+				operation that must be performed before doing any operation against the database`,
```
liran-funaro:
```suggestion
				operation that must be performed before booting the committer for the first time`,
```

cmd/committer/db_init_cmd.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	cliutil.SetDefaultFlags(cmd, &configPath)
+	// registers a duration flag for database initialization.
+	cmd.Flags().DurationVar(&timeout, "timeout", 5*time.Minute, "Timeout for the initialization operation")
+	return cmd
+}
+
+// runDatabaseInitialization reads the validator-committer config,
+// connects directly to the database using its database section, and initializes it.
+func runDatabaseInitialization(cmd *cobra.Command, configPath string, timeout time.Duration) error {
```
liran-funaro: nit: Why is this not embedded inside `databaseInitializationCMD()` as we do in all the other commands?

docker/test/common.go
```diff
@@ -98,14 +115,20 @@ func createAndStartContainerAndItsLogs(
...
+	})
 	_, err = dockerClient.ContainerStart(ctx, resp.ID, client.ContainerStartOptions{})
 	require.NoError(t, err)
-
-	//nolint:contextcheck // We want to ensure cleanup when the test is done.
-	t.Cleanup(func() {
-		stopAndRemoveContainerByID(context.Background(), t, dockerClient, resp.ID)
-	})
-
+	if !params.hostConfig.AutoRemove {
```
liran-funaro: major: Why is this optional now? We always want to STOP the container. Not just removing it.

docker/test/common.go
```diff
@@ -118,6 +141,7 @@ func createAndStartContainerAndItsLogs(
 			t.Logf("[%s] logs ended with: %v", params.name, err)
 		}
 	}()
+	return resultChannels.Result, resultChannels.Error
```
liran-funaro: Why not just return `resultChannels`?

docker/test/container_release_image_test.go
```diff
@@ -326,10 +387,40 @@ func startCommitterNodeWithTestImage(
...
+
 // mustGetWD returns the current working directory.
 func mustGetWD(t *testing.T) string {
 	t.Helper()
 	wd, err := os.Getwd()
 	require.NoError(t, err)
 	return wd
 }
+
+func requireSuccessfulExecution(
```
liran-funaro: minor: This is used only once and does not justify exporting it to a function.
It makes the code more complicated to read.

docker/test/container_release_image_test.go
```diff
@@ -326,10 +387,40 @@ func startCommitterNodeWithTestImage(
...
+	t.Helper()
+	select {
+	case err := <-errCh:
+		require.NoError(t, err)
+	case status := <-resultCh:
+		require.Zero(
+			t,
+			status.StatusCode,
+			"container failed with exit code %d",
+			status.StatusCode,
```
liran-funaro: nit: the extra message here is redundant. `testify` will include the actual status code in the error message.

docker/test/container_release_image_test.go
```diff
@@ -134,6 +131,29 @@ func TestCommitterReleaseImagesWithTLS(t *testing.T) {
...
+func TestDatabaseInitFailureWithoutActiveDB(t *testing.T) {
+	t.Parallel()
+
+	params := startNodeParameters{
+		tlsMode:       connection.NoneTLSMode,
+		dbType:        "none_activated_database",
+		dbInitTimeout: "10s",
+		artifactsPath: generateArtifacts(t),
+	}
+	resultCh, errorCh := runDatabaseInitWithReleaseImage(t.Context(), t, params.asNode(vcName))
```
liran-funaro: minor: Why use `asNode()` if we only use these parameters once?
```suggestion
	resultCh, errorCh := runDatabaseInitWithReleaseImage(t.Context(), t, startNodeParameters{
		node: vcName,
		tlsMode:       connection.NoneTLSMode,
		dbType:        "none_activated_database",
		dbInitTimeout: "10s",
		artifactsPath: generateArtifacts(t),
	})
```

docker/test/container_release_image_test.go
```diff
@@ -134,6 +131,29 @@ func TestCommitterReleaseImagesWithTLS(t *testing.T) {
...
+
+	// Expect the container to fail since there's no database available.
+	select {
+	case status := <-resultCh:
```
liran-funaro: major: Also wait on a timeout context.

docker/test/container_release_image_test.go
```diff
@@ -326,10 +387,40 @@ func startCommitterNodeWithTestImage(
...
 	require.NoError(t, err)
 	return wd
 }
+
+func requireSuccessfulExecution(
+	t *testing.T, resultCh <-chan container.WaitResponse, errCh <-chan error,
+) {
+	t.Helper()
+	select {
+	case err := <-errCh:
```
liran-funaro: major: Also wait on a timeout context.

docs/deployment-guide.md
```diff
@@ -148,7 +148,23 @@ Start the database cluster first and wait for it to be healthy before starting a
...
+     ```shell
+     committer init-db --config <path-to-vc-config> --timeout <duration>
+     ```
+     **Parameters:**
+      - `--config`: Path to the VC service configuration file (required)
+      - `--timeout`: Timeout for the initialization operation (default: 5m)
+
+   - This command connects directly to the database and creates:
+     - System tables (`tx_status`, `metadata`)
+     - System namespaces (`meta`, `config`)
```
liran-funaro:
```suggestion
     - System namespaces (`_meta`, `_config`)
```

docs/deployment-guide.md
```diff
@@ -148,7 +148,23 @@ Start the database cluster first and wait for it to be healthy before starting a
...
+     - Required stored procedures
+   - This is a one-time administrative operation that must be performed before any db operation
```
liran-funaro:
```suggestion
   - This is a one-time administrative operation that must be performed before booting the committer for the first time
```

utils/db/config.go
liran-funaro: Maybe `statedb`? @cendhu What do you think?

utils/db/dbinit.go
```diff
@@ -77,34 +73,46 @@ func NewDatabasePool(ctx context.Context, config *DatabaseConfig) (*pgxpool.Pool
...
+			)
 		}
 		logger.Infof("namespace %s: created table and its methods.", nsID)
 	}
 	return nil
 }
 
-func createNsTables(nsID string, tablets int, queryFunc func(q string) error) error {
+// CreateNsTables creates the table and functions for a namespace.
+func CreateNsTables(nsID string, tablets int) string {
```
liran-funaro: minor: The name and comment were not updated.

```suggestion
// MakeNsTablesQuery returns the query to create the table and functions for a namespace.
func MakeNsTablesQuery(nsID string, tablets int) string {
```

service/vc/database_test.go
```diff
@@ -237,9 +228,36 @@ func TestValidateNamespaceReads(t *testing.T) {
 	}
 }
 
+func TestDBInit(t *testing.T) {
```
liran-funaro, line 238: Why is TestDBInit here and not under dbinit_test.go ?

utils/db/config.go
cendhu: It is not an util. We can keep it as /statedb too. Not a trivial package to hide it under util/. It is one of the key component for us.

cendhu: In my view, I do not see utils/signature, utils/deliverorderer, and utils/db as utilities. Signature holds a key signature verification component. Usually, we expect it to reside under /service/verifier. Many a times, I wondered where is the signature verification code while browsing /service/verifier. Similarly, deliverorderer holds an important bft puller. I personally do not see it as utility. The same goes for the statedb/db package. I see utility as simple reusable code such as protoutil for block traversal, channels, connections, grpcerrors, etc...

liran-funaro: We agreed that `statedb` is a good name.

I don't think we can objectively pin down what qualifies as "utils".
For me, it means "common building blocks for the application," which is
exactly what this package is today.
But it clearly reads more narrowly for you, and that's fair; the name is doing us a disservice here.

Two things make me hesitant to promote `statedb` to the top level:
 - It's shared by the VC, query, and cmd trees, so it can't sit under any single service.
 - `utils/` already holds several of our key cross-cutting components: `deliverorderer`,
   `serve`, `signature`, `connection`. If we promote `statedb` to the top level because
   it's important, the same logic applies to those, and the root turns into a flat list
   of packages. I'd rather keep shared building blocks grouped.

Since the real friction is that "utils" undersells these components, how about we
address that directly: rename `utils/` to `common/` (as in Fabric). That removes the
"hidden under util" concern, keeps everything grouped, and fixes it repo-wide rather
than special-casing one package.
Since it is a repo-wide rename, we can do it in a separate PR.

If you feel strongly about top-level, I won't block on it, but I think
`common/statedb` is the cleaner end state.

cendhu: @liran-funaro How about `pkg/` or `internal/pkg`. `utils` and `common` are anti-pattern in go.

liran-funaro: Building on your `internal/pkg` suggestion‚ I'd go with `internal/` and drop the `pkg` sublayer. `internal` is idiomatic Go; the extra `pkg` doesn't add meaning and pushes us to three levels. So: rename `utils/` to `internal/` with no other changes, landing this package at `internal/statedb`. That keeps everything in one navigable place, drops the `util` name. WDYT?

cmd/committer/db_init_cmd.go
dean-amar: Done.

cmd/committer/db_init_cmd.go
dean-amar: We can, changed.

docker/test/common.go
dean-amar: Done.

docker/test/container_release_image_test.go
dean-amar: Inlined.

docker/test/container_release_image_test.go
dean-amar: Done.

docker/test/container_release_image_test.go
dean-amar: True, done.

docker/test/container_release_image_test.go
dean-amar: Added, thanks.

docker/test/container_release_image_test.go
dean-amar: Added.

utils/db/dbinit.go
dean-amar: Done.

service/vc/database_test.go
dean-amar: It's testing a runtime database operation (inserting keys), not initialization, so it belongs with other database operation tests in database_test.go.
I'll change its name to a more suitable one.

docker/test/common.go
dean-amar: I originally added this term to avoid 'container not found' error logs when AutoRemove had already cleaned up the container. But you're right, we should always register cleanup to handle cases where containers hang or tests are interrupted.

utils/db/config.go
cendhu: Yes. Sounds good. `internal` instead of `utils`.

liran-funaro: I'll open an issue for this. This PR will put `statedb` under `utils` and we can unblock it.

dean-amar: Done.

cmd/committer/db_init_cmd_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/hyperledger/fabric-x-committer/cmd/config"
+	"github.com/hyperledger/fabric-x-committer/utils/statedb"
+	"github.com/hyperledger/fabric-x-committer/utils/testdb"
+)
+
+func TestMain(m *testing.M) {
+	testdb.RunTestMain(m)
+}
+
+func TestInitStateDBCMD(t *testing.T) {
```
liran-funaro: We don't need to test the CMD functionality here against the DB.
We have a unit test in the `vc` package, and a integration test that tests the CMD, and the docker test as well.

utils/statedb/dbinit_test.go
```diff
@@ -4,45 +4,19 @@ Copyright IBM Corp. All Rights Reserved.
...
 	"time"
 
-	"github.com/hyperledger/fabric-x-common/api/committerpb"
 	"github.com/stretchr/testify/require"
 
 	"github.com/hyperledger/fabric-x-committer/utils/connection"
 	"github.com/hyperledger/fabric-x-committer/utils/retry"
 )
 
-func TestDBInit(t *testing.T) {
```
liran-funaro, line 21: It is understandable we don't put this test here as the CI runs this tests without a DB.
Two issues:
1. Add a detached comment here that explains why we don't have a unit test in this package as stated above.
2. Put this test in the VC package instead for completeness (not in the CMD package). The comment should mention where to find the relevant test.

service/vc/database_test.go
```diff
@@ -237,9 +228,36 @@ func TestValidateNamespaceReads(t *testing.T) {
 	}
 }
 
+func TestInsertNamespaceKeys(t *testing.T) {
```
liran-funaro: DB Init test should be right above this method.

cmd/committer/db_init_cmd.go
```diff
@@ -0,0 +1,56 @@
...
+	cmd := &cobra.Command{
+		Use:   initDBCommand,
+		Short: "Initialize the database with required tables and namespaces",
+		Long: `Initialize the state database by creating system tables, metadata tables,
+				and system namespaces (meta and config). This is a one-time administrative
+				operation that must be performed before booting the committer for the first time`,
+		Args:         cobra.NoArgs,
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			vcConfig, _, err := config.ReadVCYamlAndSetupLogging(config.NewViperWithVCDefaults(), configPath)
```
liran-funaro: The previous review did not ask to limit the config to be of the VC, just that we can use the existing VC/query service config instead of creating a new one.
We should have a dedicated viper config reader that return `statedb.Config` that expect only the `database` field, thus is agnostic to the other fields.

service/vc/database_test.go
liran-funaro: Yes. The test should be here. Please return it here. Please add a comment here and in `statedb` tests package why it is here and not in `statedb` package.

cmd/committer/db_init_cmd_test.go
dean-amar: So should we remove this test entirely?

cmd/committer/db_init_cmd.go
dean-amar: Done.

PR #620, opened 2026-06-03
[notify] Introduce StreamAllTransactions stream

### LINKED ISSUE #618: [notification] Streaming API for Committed Transactions
### Summary
Extend the Notification Service with a new streaming API that allows clients to subscribe to and receive real-time notifications of **all** committed transactions, not just those with pre-registered transaction IDs.

### Background
Currently, the Sidecar provides three mechanisms for accessing transaction data:

1. **Block Delivery Service** - Streams committed blocks to clients, but requires manual parsing of the block structure to extract individual transactions
2. **Notification Service** - Allows clients to subscribe to **specific transaction IDs** they know in advance, receiving asynchronous status updates when those transactions complete
3. **Query Service** - Provides read-only access to the committed world state with configurable isolation levels

The Notification Service currently requires clients to know transaction IDs upfront to subscribe. There is no mechanism to observe **all** committed transactions in a parsed, transaction-level format without manually parsing blocks from the Block Delivery API.

### Use Cases

1. **Audit and Monitoring**: External systems need to track all state changes for compliance, auditing, or analytics purposes without implementing block parsing logic
2. **Event-Driven Applications**: Applications that react to any state change in the blockchain, not just their own transactions, with minimal parsing overhead
3. **Replication and Synchronization**: Secondary systems that need a transaction-level view of all commits
4. **Real-time Dashboards**: Monitoring tools that display live transaction throughput and patterns across all namespaces

### Proposed Solution

Add a new streaming RPC to the **Notification Service** (same server as Block Delivery, different API) that provides a **transaction-level stream** of all committed transactions.

**Option 1:** Batched into blocks, in the order they are committed.
**Option 2:** Batched according to processing order. Reported once committed.

The stream will start from the currently processed block, and will include the transaction reference (Transaction ID, block number, transaction index) and the commit status (committed/aborted/etc...).

**Optional features:**
1. **Supports filtering** by:
   - Namespace(s)
   - Transaction status (committed, aborted, rejected)
2. **Provides parsed transaction details** including:
   - Read/write sets (optional, configurable)
   - Endorsements (optional, configurable)

### API Design Considerations

Extend the existing `Notifier` service in the Notification Service:

```protobuf
service Notifier {
    // Existing API - subscribe to specific transaction IDs
    rpc OpenNotificationStream (stream NotificationRequest) returns (stream NotificationResponse);
    
    // NEW API - subscribe to all committed transactions
    rpc StreamAllTransactions(stream StreamAllRequest) returns (stream TxBatch);
}

message StreamAllRequest {
    // The following are optional. These will be implemented if requested.
    repeated string filter_namespaces = 1;  // Empty = all namespaces
    repeated Status filter_status = 2; // Empty = all statuses
    bool include_read_write_sets = 3;
    bool include_endorsments = 4;
}

message TxBatch {
    repeated TxEvent = 1;
}

message TxEvent {
    TxRef ref = 1;
    Status status = 2;

    // A list of namespaces that define the transaction's scope.
    // If namespace filtering is activated, only the relevant namespaces will be included.
    // Empty if not requested. 
    repeated TxNamespace namespaces = 3;

    // A list of endorsements.
    // Empty if not requested. 
    repeated Endorsements endorsements = 4;
}
```

### Implementation Considerations

1. **Service Location**: This will be part of the **Notification Service**
2. **Leverage Existing Infrastructure**: Build on top of the Sidecar's internal committed block stream - consume committed blocks and emit individual transaction events
3. **Performance Impact**: Streaming all transactions could create a significant load - consider:
   - Share the existing `max-concurrent-streams` limit with Block Delivery and existing Notification streams
   - Backpressure mechanisms if clients can't keep up (optional)
   - Memory buffering limits
4. **Consistency Guarantees**: Transactions should be streamed in deterministic block order
5. **No Recovery**: Disconnected clients might miss blocks

### Benefits Over Block Delivery API

- **Simplified Client Implementation**: No need to parse complex block structures
- **Block-Level Granularity**: Clients receive events per block
- **Built-in Filtering**: Server-side filtering reduces network traffic and client processing
- **Consistent with Existing Notification Service**: Similar event structure and configuration to the existing transaction ID-based notification API
- **Unified Service**: Both notification mechanisms (specific TXs and all TXs) in the same service

### Related Documentation

- notification-service.md - Existing notification mechanism for specific transaction IDs (same service where this will be added)
- sidecar.md - Block Delivery Service (requires manual parsing) and Notification Service architecture

### Questions for Discussion

1. Stream order: block order (batched into blocks), or processing order.
   **Decision:** Block order.
2. Filtering support.
    **Decision:** Allow filtering by namespace and commit status.
4. Stream content (optional TX content and endorsements).
    **Decision:** Allow optional TX content and endorsements.

liran-funaro:
#### Type of change

- New feature
 
#### Description

- Introduce StreamAllTransactions stream to the notification service

#### Related issues

- resolves #618


service/sidecar/notify.go
```diff
@@ -277,9 +360,161 @@ func (m *subscriptions) removeAndEnqueueTimeoutEvents(
...
+		if s.includeEndorsements {
+			event.Endorsements = tx.Content.Endorsements
+		}
+		events = append(events, event)
+	}
+
+	return events
+}
+
+func getMatchingNamespaces(
```
cendhu: getMatching and filterNamespaces are confusing.Let's use filter everywhere as it denotes what we let in as per our context. My suggestion would be `applyFilter` or something similar.

service/sidecar/notify.go
```diff
@@ -168,6 +216,40 @@ func (n *notifier) OpenNotificationStream(stream committerpb.Notifier_OpenNotifi
...
 
+// StreamAllTransactions implements the [committerpb.NotifierServer] API.
+// It streams all committed transactions to the client, with optional filtering
+// by namespace and transaction status.
+func (n *notifier) StreamAllTransactions(
+	req *committerpb.StreamAllRequest,
+	stream committerpb.Notifier_StreamAllTransactionsServer,
+) error {
+	n.metrics.notifierActiveStreams.Inc()
+	defer n.metrics.notifierActiveStreams.Dec()
```
cendhu, line 227: I assume the MaxConcurrentStream and stream interceptor are applicable for this stream too automatically.

cendhu (approved):
LGTM. In a separate PR, we should also update our loadgen to use this stream to avoid processing blocks.

service/sidecar/notify.go
liran-funaro: Yes. It is tested in the integration test.

service/sidecar/notify.go
liran-funaro: I used `filterAndBuildEvents()` and `filterNamespaces()`

liran-funaro:
@cendhu There is a problem with the linter download. I ran the linter locally and it passes.

PR #629, opened 2026-06-16
[coordinator] BUGFIX coordinator does not pass metadata to verifier

liran-funaro:
#### Type of change

- Bug fix
 
#### Description

- The coordinator did not pass the metadata field to the verifier
- This commit modifies the coordinator to store the full TX object, thus, any future changes will automatically pass to the verifier


liran-funaro:
@cendhu I'm merging this without review to speed up the process.

Issue #10, opened 2025-07-03
Do not allow other namespaces in tx along with metanamespace


cendhu:
While duplicate namespaces are currently prohibited, a user could attempt to circumvent this by including an entry for a new namespace within the same metanamespace transaction that creates it. This would result in an invalid transaction.

However, for simplicity (especially to implement #562), we can restrict a transaction to contain only metanamespace tx.

PR #631, opened 2026-06-22
[integration-test] create multiple namespaces from a single tx

cendhu:
#### Type of change

- Test update
 
#### Description

This commit adds a test case to check whether we can create multiple namespaces from a single transaction.


liran-funaro (approved):
LGTM. Is issue #10 relevant here?
- #10

cendhu:
> LGTM. Is issue #10 relevant here?
> 
> * #10

@liran-funaro No. Ale asked for sample code to create multiple namespaces within a transactions. As there was no test with such a scenario, I had added this. We yet to decide on allowing user-namespaces along with meta-namespace within a transaction.

PR #642, opened 2026-07-06
[coordinator, sidecar] Make retry profile configurable for sustained operations

### LINKED ISSUE #625: TODO items mentioned in the codebase
# TODO Items - Fabric-X Committer

**Total Items: 23**

---

## Configuration (9 items)

- [x] **Make retry configurable in Signature Verifier Manager**  
  **File**: `service/coordinator/signature_verifier_manager.go:113`
  ```go
  // TODO: initialize retry from config.
  return retry.Sustain(eCtx, nil, func() error {
  ```

- [x] **Make retry configurable in Validator Committer Manager**  
  **File**: `service/coordinator/validator_committer_manager.go:123`
  ```go
  // TODO: initialize retry from config.
  return retry.Sustain(eCtx, nil, func() (err error) {
  ```

- [x] **Make retry configurable in Sidecar**  
  **File**: `service/sidecar/sidecar.go:158`
  ```go
  // TODO: initialize retry from config.
  return retry.Sustain(gCtx, nil, func() error {
  ```

- [x] **Make chunk size configurable in Coordinator**  
  **File**: `service/coordinator/coordinator.go:349`
  ```go
  // TODO: make it configurable.
  chunkSizeForDepGraph := min(c.config.DependencyGraph.WaitingTxsLimit, 500)
  ```

- [x] **Make sampling time configurable in Coordinator queue monitor**  
  **File**: `service/coordinator/coordinator.go:403`
  ```go
  // TODO: make sampling time configurable
  ticker := time.NewTicker(100 * time.Millisecond)
  ```

- [x] **Make sampling time configurable in Dependency Graph Manager**  
  **File**: `service/coordinator/dependencygraph/manager.go:79`
  ```go
  // TODO: make sampling time configurable
  ticker := time.NewTicker(100 * time.Millisecond)
  ```

- [ ] **Make queue multiplier configurable in VC Service**  
  **File**: `service/vc/validator_committer_service.go:81`
  ```go
  // TODO: make queueMultiplier configurable
  queueMultiplier := 1
  ```

- [ ] **Make sampling time configurable in VC Service queue monitor**  
  **File**: `service/vc/validator_committer_service.go:171`
  ```go
  // TODO: make sampling time configurable
  ticker := time.NewTicker(250 * time.Millisecond)
  ```

- [ ] **Move MaxMsgSize constant to fabric-x-common**  
  **File**: `utils/connection/client.go:41`
  ```go
  // TODO: All services including the orderer must use the same default maximum message size.
  //       Hence, we need to move this constant to fabric-x-common.
  ```

---

## Code Refactoring (4 items)

- [ ] **Rename channel variable in Coordinator**  
  **File**: `service/coordinator/coordinator.go:85`
  ```go
  // TODO: As signValidatedTxsNode and vcServiceToDepGraphValidatedTxs can cause confusion, it would be better to
  //       rename vcServiceToDepGraphValidatedTxs to committedOrAbortedTxsNode.
  ```

- [ ] **Split signature verification file**  
  **File**: `utils/signature/verify.go:21`
  ```go
  // TODO: Move channelPolicyVerifier, mspPolicyVerifier, and keyVerifier to their own files.
  //       It is difficult to read the current file due to the interleaving of methods from different objects.
  ```

- [x] **Merge database files**  
  **File**: `service/vc/dbinit.go:80`
  ```go
  // TODO: merge this file with database.go.
  ```

- [ ] **Remove unused ChannelHeader fields**  
  **File**: `service/sidecar/mapping.go:104`
  ```go
  // skips unused ChannelHeader fields (version, timestamp, channel_id, epoch,
  // extension, tls_cert_hash) and the Header's signature_header. Corruption in
  // those fields will go undetected. This is acceptable because the committer
  // does not use them, and for the same reason, they are not validated in the
  // sidecar. TODO: remove unused fields from the ChannelHeader proto.
  ```

---

## Testing (3 items)

- [ ] **Use real signature verifier in coordinator tests**  
  **File**: `service/coordinator/coordinator_test.go:381`
  ```go
  // TODO: Use real signature verifier instead of mocks.
  ```

- [ ] **Add test for commit retry**  
  **File**: `service/vc/committer.go:80`
  ```go
  // TODO: Add test to ensure commit is retried.
  ```

- [ ] **Improve block store test verification**  
  **File**: `service/sidecar/block_store_test.go:93`
  ```go
  // TODO: appendBlock forces fsync (Append) for single-tx blocks since they may be config blocks,
  //       but we cannot verify Append vs AppendNoSync here because the ledger field is a concrete
  //       *fileledger.FileLedger (not an interface). To properly test this, the ledger dependency
  //       would need to be behind an interface so we can assert which method was called.
  ```

---

## Performance (2 items)

- [ ] **Evaluate parallel namespace validation**  
  **File**: `service/vc/validator.go:98`
  ```go
  // TODO: We can run per namespace validation in parallel. However, we should not
  // 		 over parallelize and make contention among preparer, validator, and committer
  //       goroutines to acquire the CPU. Based on performance evaluation, we can decide
  //       to run per namespace validation in parallel.
  ```

- [ ] **Evaluate parallel blind write processing**  
  **File**: `service/vc/committer.go:190`
  ```go
  // TODO: Though we could run the following in a goroutine per namespace, we restrain
  // 		 from doing so till we evaluate the performance
  ```

---

## Security (2 items)

- [ ] **Pass database password via environment variable - VC Config**  
  **File**: `cmd/config/samples/vc.yaml:58`
  ```yaml
  # TODO: pass password via environment variable for better security.
  password: "yugabyte"
  ```

- [ ] **Pass database password via environment variable - Query Config**  
  **File**: `cmd/config/samples/query.yaml:62`
  ```yaml
  # TODO: pass password via environment variable for better security.
  password: "yugabyte"
  ```

---

## Documentation (1 item)

- [ ] **Add channel flow diagram**  
  **File**: `service/coordinator/dependencygraph/global_dependency_manager.go:38`
  ```go
  // TODO: add a text figure explaining how the input and output channels
  //       are being used.
  ```

---

## Feature Implementation (1 item)

- [ ] **Implement seek info for ledger service**  
  **File**: `utils/deliver/delivery.go:143`
  ```go
  // TODO: We have seek info only for the orderer but not for the ledger service. It needs
  //       to implemented as fabric ledger also allows different seek info.
  ```

---

## Metrics (1 item)

- [ ] **Add lock duration metrics**  
  **File**: `service/coordinator/signature_verifier_manager.go:318`
  ```go
  // TODO: introduce metrics to measure the lock wait/holding duration.

dean-amar:
#### Type of change

- Improvement

#### Description

- This PR addresses TODOs in the coordinator's Validator Committer Manager, Signature Verifier Manager and Sidecar by making retry behavior configurable through the existing reconnect configuration.

#### Additional Details

- The retry profiles used for these managers are the ones defined in the `MultiClientConfig`, which are used for connection retries. 

- Note: We can also introduce a new configurable retry profile specifically for the coordinator that will be passed on to the managers if separate control over stream retry vs connection retry is needed in the future.

#### Related issues

- Resolves partly #625


cendhu:
Till we have dynamic scaling capability through discovery service, we cannot add a retry profile for this operation. Currently, we can add `n` verifiers' IP to the config and dynamically start/stop a few depending on the load. For the internal GA, this is what we have recommended if any client really need them.

liran-funaro:
@cendhu, we already have a retry profile for this (default). I don't understand the reasoning for your comment.
How does not having dynamic scaling justify using the default retry profile, rather than allowing the user to decide?

cendhu:
> @cendhu, we already have a retry profile for this (default). I don't understand the reasoning for your comment.
> How does not having dynamic scaling justify using the default retry profile, rather than allowing the user to decide?

I didn't know we have a default retry profile. Only if we retry indefinitely, we can support dynamic scaling for now.

liran-funaro:
> Only if we retry indefinitely, we can support dynamic scaling for now.

I suggest opening an issue for dynamic scaling via the retry mechanism.
The only parameter that should be handled with care is `max-elapsed-time`. We can set it to infinity when we need auto scaling. The default is 15 minutes.
The other retry parameters are still applicable and can/should be configured by the user.

liran-funaro:
@cendhu The retry config will be addressed by #679 
We can unblock this PR.

PR #649, opened 2026-07-14
[coordinator] fix deadlock waiting for pending transactions

### LINKED ISSUE #616: [coordinator-sidecar] possible deadlock when waiting for pending transactions
If `vcServiceToCoordinatorTxStatus` cap is not larger than `numWaitingTxsForStatus`, we will reach a deadlock in the sidecar.
We should address this. Maybe in a different PR.

_Originally posted by @liran-funaro in https://github.com/hyperledger/fabric-x-committer/pull/613#discussion_r3309978828_

liran-funaro:
#### Type of change

- Bug fix

#### Description

While the sidecar stream is inactive, the VC keeps filling the coordinator's
bounded status queue, but its only drainer (`sendTxStatus`) is gone with the
stream. If the in-flight backlog exceeds the queue capacity, the VC blocks on the
full queue and recovery deadlocks: the sidecar won't reconnect until
`NoPendingTransactionProcessing` reports idle, which can't happen while the VC is
blocked.

Fix by draining and discarding queued statuses in `NoPendingTransactionProcessing`
(polled during recovery). This unblocks the VC so it can finish. Discarding is
safe — statuses are durable in the state DB and recovered via
`GetTransactionsStatus`, never from this queue.

Adds a regression test reproducing the deadlock.

#### Related issues

 - resolves #616

### DIFF
--- a/docs/coordinator.md
+++ b/docs/coordinator.md
@@ -188,7 +188,7 @@ GetConfigTransaction(ctx context.Context, in *Empty, opts ...grpc.CallOption) (*
 ```go
 NoPendingTransactionProcessing(ctx context.Context, in *Empty, opts ...grpc.CallOption) (*wrapperspb.BoolValue, error)
 ```
-  * This API returns true when all previously submitted transactions have been processed.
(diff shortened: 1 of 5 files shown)


cendhu (approved):
Good catch.

PR #669, opened 2026-07-15
[config] Default value annotation

### LINKED ISSUE #570: [cmd] Add default config values as a config struct annotations
This issue suggest using struct annotations to set default values for configurations.

For example:
```go
type Config struct {
    Number int `mapstructure:"logging" default:"5"`
}
```

liran-funaro:
#### Type of change

#### Description

- Declare config defaults with `default:"..."` struct tags, applied to Viper by a single reflection walk (unified with the existing env-override walk)
- Support struct-assignment tags (`key=value`) for scoped nested defaults on shared types (retry, orderer suspicion window, logging spec)
- Move per-service default ports into `cmd/config`; remove all other defaults-only constants from service packages
- Drop `required` from fields that carry a default
- Scope rate-limit and max-concurrent-streams defaults to the client-facing services (query, sidecar)
- Align sample-YAML`# Default:` comments with the actual defaults

#### Related issues

- resolves #570


cmd/config/viper.go
```diff
@@ -11,111 +11,80 @@ import (
...
+	sidecarMonitoringPort     = 2114
+	verifierServerPort        = 5001
+	verifierMonitoringPort    = 2115
+	vcServerPort              = 6001
+	vcMonitoringPort          = 2116
+	queryServerPort           = 7001
+	queryMonitoringPort       = 2117
+	loadgenServerPort         = 8001
+	loadgenMonitoringPort     = 2118
+)
```
cendhu, line 37: When I placed these defaults within each service, I had the following choices:

1. place them within each service -- through the config is not defined there but we have all constants in one place and know where to look for.
2. place them within cmd/config -- place where the default is used but the config is not defined here.
3. place them with utils/connection -- where we define the connection configs. 

To me, none of them looked perform and hence, I went with 1. Do you have any reasoning for choosing 2 over the other?

cmd/config/config_preload.go
```diff
@@ -0,0 +1,108 @@
+/*
```
cendhu, line 1: Was the file renamed before applying changes? Difficult to compare with env_vars.go as it is not showing as rename.

cmd/config/viper.go
liran-funaro: The reasoning is as follows:
1. They are used only in the CMD package
2. They are part of the configuration, not part of the service code. The service code does not confined to a specific port. Only the CMD part is what defines these.
3. The other default values fits well inside the service as they are the default that fits the service behaviour. This is the rational you made and it is correct for internal service values. It just doesn't fit for the service port, which are a global decision (we need to prevent port collision).

cmd/config/config_preload.go
liran-funaro: Option 1: I'll rename back to env_vars.go (but this will remain the fixed name for this file).
Option 2: Keep it as such, but use external tools to compare.
Please let me know what do you prefer.

cendhu: I will go with option 2. In the future, we can do this with two commits in the same PR.

liran-funaro: Good idea. I'll remember this for next time.

cendhu (approved):

PR #676, opened 2026-07-16
[loadgen] Allow bare number as const distribution shorthand

### LINKED ISSUE #673: [loadgen] Allow simple const config for disribution
Currently to set a const distribution, we have to set it manually. For example:
```yaml
    read-write-count:
      const: 2
```

This issue suggest supporting a syntactic sugar that will allow the following form:
```yaml
    read-write-count: 2
```

liran-funaro:
#### Type of change

- New feature
 
#### Description

Add a mapstructure decode hook so a bare number is accepted as a shorthand for a const distribution. For example:

    read-write-count: 2

is now equivalent to:

    read-write-count:
      const: 2

The hook targets the workload.Distribution type, so the shorthand applies uniformly to every distribution config field (read/write counts, query-size, min-invalid-keys-portion, dependency gap). Non-numeric values (maps, strings, booleans) are left untouched, so the full distribution forms (const, uniform, normal, discrete, bernoulli) keep working.

- Documented the shorthand in the loadgen sample config and switched the read-write-count example to it, which makes the existing sample-loading test cover the shorthand end-to-end through the real config loader.
- Added decoder unit tests for the shorthand (int/float/negative/zero), regression tests for the explicit forms, and a guard test ensuring a non-numeric scalar surfaces a decode error rather than being coerced.

#### Related issues

 - resolves #673

### DIFF
--- a/cmd/config/config_decoder.go
+++ b/cmd/config/config_decoder.go
@@ -19,6 +19,7 @@ import (
 	"github.com/hyperledger/fabric-x-common/common/viperutil"
 	"github.com/spf13/viper"
 
+	"github.com/hyperledger/fabric-x-committer/loadgen/workload"
 	"github.com/hyperledger/fabric-x-committer/utils/connection"
 	"github.com/hyperledger/fabric-x-committer/utils/serve"
 )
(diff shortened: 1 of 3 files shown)


cmd/config/config_decoder.go
```diff
@@ -42,10 +43,31 @@ const (
...
+	case rv.CanInt():
+		return workload.Distribution{Const: float64(rv.Int())}, nil
+	case rv.CanUint():
+		return workload.Distribution{Const: float64(rv.Uint())}, nil
+	case rv.CanFloat():
+		return workload.Distribution{Const: rv.Float()}, nil
+	default:
+		return rawData, nil
+	}
+}
```
cendhu, line 69: I know these are pre-existing but do we really need all these distributions. We have never used anything other than const till now. Hence, I wonder whether we can simplify this by removing the support for various distributions. When there is a real need, we can introduce back only the required distribution. I doubt such a need would arise anytime soon or maybe never.

liran-funaro: We never used it so far, I agree. But we should. In a real-world workload, I doubt the TX layout is fixed. It changes and fits a certain distribution.
We use it in the TX rw-set size and for the dependency injection gap.
It makes sense to make the gap a distribution, not a fixed number.
Regarding the TX rw-set sizes, we can make it a choice of fixed layouts.
That is, instead of a distribution for each part (r/rw/w), we can list a few fixed layouts and have a probability for each. I argue that this is even more realistic and usable than the option we have at the moment.

liran-funaro:
```yaml
transaction:
  - weight: 7                    # simple transfer (70%)
    read-write: 2
    read-write-value-size: 32
  - weight: 2                    # read-heavy query-then-write (20%)
    read-only: 4
    read-write: 1
    read-write-value-size: 32
  - weight: 1                    # bulk blind insert of large values (10%)
    blind-write: 5
    blind-write-value-size: 256
```

cendhu: Realistic should be tested along with the application/endorsers. For example, idap load generator would test the realistic needs. We just do a constant test for peak capacity, identify bottleneck, and fix them. Workload mix, tx rate pattern over time, etc... are out of scope for us.

liran-funaro: Mixed workload can affect the bottleneck. For example, how rare TX with many keys affects the dependency graph, or how a less common TX with a large value affects the storage layer.
I'm sure there are more examples. These are just the ones I came up with now.

cendhu: These are very fine trained testing. Without knowing what application does, I am not sure how useful these tests are. I would simply start with a const. We then evolve. It may be needed in the future.

liran-funaro: Please open an issue requesting the removal of the distribution in the loadgen.
We can discuss this there.

liran-funaro:
Closed due to 
- #682

PR #680, opened 2026-07-19
[utils] Move gRPC server bootstrapping packages to fabric-x-common

### LINKED ISSUE #675: Discussion: publish `utils/serve` to fabric-x-common
## Context

fabric-x-evm is expected to need the same gRPC server bootstrapping that `utils/serve` provides here (health-check service, TLS provider, interceptors, config) for the endorsement feature. Rather than evm copying the code or depending directly on fabric-x-committer, it can be worthwhile to publish `serve` into fabric-x-common, which fabric-x-committer already depends on.

Advantages:

- Single source of truth - bug/security fixes to server bootstrapping (TLS, retry, health checks) are shared.
- Consistency - evm, committer and future components get the same battle-tested gRPC serving behavior, with similar configuration flags.
- Right dependency direction - extends the existing common ← committer relationship instead of evm depending on committer or hand-copying code.
- Shared maintenance burden - the code is reviewed/owned by more than one team, rather than being committer's problem alone.

It's possible that other (official or unofficial) components will be created later that also need GRPC connectivity and want to fit in the Fabric ecosystem. Moving the code to fabric-x-common might even make that more inviting. As an example, the [custom endorser sample](https://github.com/hyperledger/fabric-x-samples/tree/main/custom-endorser) was easy to build with few lines of code thanks to depending on the committer `utils/serve` package. At the same time you can also see it as an example of a sub-optimal dependency direction...

## What would move

Just three packages, forming a clean, shallow chain: `serve → connection → retry`. `retry` has no internal fabric-x-committer dependencies at all; `connection` depends only on `retry`; `serve`'s production code (`server.go`, `config.go`, `interceptors.go`, `start_serve.go`, `tls_provider.go`) depends only on `connection`. No coordinator/sidecar/VC concepts anywhere in the chain.

## Decoupling check

At first sight, the dependency graph seems to suggest a reliance on committer specific code. However this is only through tests, but not through production code. So some rewriting of tests would be necessary to keep the same coverage, but the scope is limited.

## Alternatives considered (for fabric-x-evm)

- **Copy the code into fabric-x-evm**: avoids the dependency question but risks silent drift — bug fixes and security patches applied in one repo won't reach the other.
- **evm depends directly on fabric-x-committer**: wrong direction, and drags in unrelated coordinator/sidecar/VC concerns that evm has no use for.

## Ask

Discuss whether publishing `serve` (+ `connection` and `retry`) to fabric-x-common is agreeable.

dean-amar:
#### Type of change

- Improvement (code reusability)
- Test update

#### Description

- Committer side of the cross-repo move of the gRPC server-bootstrapping chain into `fabric-x-common`. 
 This PR removes the implementations from `fabric-x-committer` and repoints every import to its new home under `github.com/hyperledger/fabric-x-common/utils/...`.

Packages removed from this repo (now provided by fabric-x-common):

- `utils/serve`      — gRPC/HTTP server lifecycle, interceptors, TLS provider, config
- `utils/connection` — gRPC client dialing, endpoints, TLS, health checks
- `utils/retry`      — backoff/retry primitives
- `utils/channel`    — readiness signaling

#### Related issues

- Partially resolves #675


utils/serialization/transaction_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/hyperledger/fabric-x-common/tools/configtxgen"
+	"github.com/hyperledger/fabric-x-common/tools/cryptogen"
+	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger/fabric-x-committer/utils/serialization"
+)
+
+const localhost = "localhost"
+
+func TestExtractAppTLSCAsFromEnvelope(t *testing.T) {
```
liran-funaro, line 21: How is this relevant to this PR?

utils/test/cryptogen_tls.go
```diff
@@ -0,0 +1,55 @@
...
+		cryptogen.MSPDir, cryptogen.TLSCaCertsDir, "tlspeer-org-0-CA-cert.pem")
+
+	// OrdererRootCATLSPath is the path to organization 0's orderer TLS credentials in the crypto materials directory.
+	OrdererRootCATLSPath = filepath.Join(cryptogen.OrdererOrganizationsDir,
+		"orderer-org-0.com", cryptogen.MSPDir, cryptogen.TLSCaCertsDir, "tlsorderer-org-0-CA-cert.pem")
+)
+
+// NewServiceTLSConfig creates a server TLS configuration with certificates loaded from the artifact path.
+// This function constructs paths to TLS certificates for a given service within the peer organization structure.
+func NewServiceTLSConfig(artifactsPath, serviceName, mode string) connection.TLSConfig {
```
liran-funaro: How is this relevant to this PR? Was this moved from someplace else?

dean-amar: This has been changed.

utils/serialization/transaction_test.go
dean-amar: This test previously lived under `serve/tls_provider_test.go` and depended on the `utils/serialization` package, which resides in the committer. To avoid a circular dependency, the test will remain in the committer. Its original location was also a mistake: it has nothing to do with the `dynamic_tls` mechanism and only tests `ExtractAppTLSCAsFromEnvelope`, which is part of the serialization package.

PR #684, opened 2026-07-20
[retry] Support unlimited retries via max-elapsed-time

### LINKED ISSUE #679: Support dynamic scaling of verifiers/VCs via the retry mechanism (indefinite max-elapsed-time)
#### Type

- Improvement / Design

#### Context

Follow-up from the discussion on #642 ("Make retry profile configurable for sustained
operations"). That PR makes the retry profile configurable for the coordinator's
Signature Verifier Manager, Validator Committer Manager, and the Sidecar's coordinator
client, reusing the `reconnect` profile from `MultiClientConfig`/`ClientConfig` (partly
resolves #625).

During review, @cendhu noted that dynamic scaling of verifiers/VCs is a desired
capability: operators can pre-configure `n` endpoints and start/stop instances depending
on load. For this to work without failing operations, the coordinator must keep retrying
reconnection to an endpoint that is temporarily down (being scaled) instead of giving up.
This issue tracks supporting that via the (now configurable) retry mechanism.

#### How retry drives reconnection today

Each verifier/VC endpoint runs a `retry.Sustain` loop that maintains its streaming
connection and restarts it on transient failure:

- `service/coordinator/signature_verifier_manager.go` (`sendTransactionsAndForwardStatus`, one Sustain per verifier)
- `service/coordinator/validator_committer_manager.go` (`sendTransactionsAndForwardStatus`, one Sustain per VC)
- `service/sidecar/sidecar.go` (`sendBlocksAndReceiveStatus`, sidecar ↔ coordinator block delivery)

`Sustain` keeps the operation alive until the context is cancelled, a permanent error
occurs, or continuous transient failures exceed `Profile.MaxElapsedTime`. So
`max-elapsed-time` is effectively the window during which an endpoint may be down (e.g.
being scaled) before the coordinator gives up and the run fails.

#### The gap

For dynamic scaling we want **indefinite** retry (`max-elapsed-time` = infinity) so
instances can be stopped/started at will without crashing the coordinator/sidecar. This
cannot be expressed through config today:

- The intended semantics are "0 = unlimited" (`utils/retry/profile.go` package/struct
  docs; `backoff/v5`'s `WithMaxElapsedTime(0)`).
- But `Profile.WithDefaults()` coerces a `0`/unset `max-elapsed-time` back to the
  15-minute default (`utils/retry/profile.go`, the `if p.MaxElapsedTime > 0` guard), and
  every entry point (`Sustain`, `executeWithResult`) runs `WithDefaults()` first. So
  `max-elapsed-time: 0` silently becomes 15 minutes — infinity is unreachable. The only
  current workaround is a very large finite value.

#### Proposal

1. Provide a way to configure indefinite retry for `max-elapsed-time` (e.g. a sentinel
   value, or distinguishing an explicitly-set `0` from unset), so operators can opt into
   never-give-up reconnection for dynamic-scaling deployments.
2. Ensure the coordinator manager and sidecar `Sustain` loops honor this and keep
   reconnecting to endpoints across scale-down/up. The other retry parameters
   (`initial-interval`, `randomization-factor`, `multiplier`, `max-interval`) remain
   applicable and user-configurable.
3. Document the dynamic-scaling workflow (pre-configure endpoints; enable infinite retry;
   start/stop instances) and the care needed around `max-elapsed-time` — a too-short
   budget breaks scaling, while infinity means the coordinator never gives up on a
   permanently-dead endpoint.

#### Notes / open questions

- #642 reuses the connection `reconnect` profile for the sustained stream loop. As noted
  there, a separate retry profile for the coordinator (stream retry vs connection retry)
  may be warranted if independent control is needed.
- Today scaling is static (`docs/deployment-guide.md`: endpoints are pre-configured;
  scale by starting/stopping pre-configured instances). There is no service discovery;
  the `Sustain` reconnect loops are the natural place to hang dynamic re-resolution later.

#### Related

- Follows up on #642
- Related to #625 (retry configurability TODOs)

liran-funaro:
#### Type of change

- Improvement

#### Description

- Make `retry.Profile.MaxElapsedTime` a `*time.Duration` so an operator can request unlimited retries via `max-elapsed-time: 0`. This enables dynamic scaling of pre-configured instances: the coordinator/sidecar keep reconnecting to instances that are stopped/started at runtime instead of giving up after the default 15m budget.

#### Changes

- Semantics: `nil` (unset) → default 15m (unchanged); `0` → unlimited; positive → that budget. A `validate:"omitempty,gte=0"` tag rejects negative values at config load.
- Unlimited retries are provided by the backoff/`Sustain` layer (`WithMaxElapsedTime(0)`). The gRPC retry policy has no unlimited option, so it stays intentionally bounded at `defaultGrpcMaxAttempts` and relies on `Sustain` to re-establish the connection (service reconnection only).
- Rewrite `connection.CalcMaxAttempts` as a closed-form formula instead of a loop (behavior-preserving; validated by the unchanged existing test vectors).
- Document the dynamic-scaling retry-budget guidance and the `0 = unlimited` option in the deployment guide and sample configs.

#### Related issues

- Resolves #679


docs/deployment-guide.md
```diff
@@ -113,6 +113,8 @@ For detailed service descriptions and architecture, see the [Architecture Guide]
 
 All service endpoints must be pre-configured before deployment. To scale up, start additional pre-configured instances. To scale down, stop the instances no longer needed.
 
+**Retry budget for dynamic scaling.** The Coordinator maintains its connections to the Verifier and Validator-Committer instances (and the Sidecar its connection to the Coordinator) with an exponential-backoff retry loop bounded by the `reconnect.max-elapsed-time` budget (default 15 minutes). If a pre-configured instance stays down longer than that budget — for example, while it is stopped for scaling — the sustained operation gives up. For deployments that stop and start instances at runtime, set `reconnect.max-elapsed-time: 0s` on the relevant client configuration to retry **indefinitely** (never give up), so the Coordinator keeps reconnecting as instances come and go. The other retry parameters (`initial-interval`, `randomization-factor`, `multiplier`, `max-interval`) still apply. Note that this only removes the time limit from the reconnection loop; the underlying gRPC retry policy remains bounded by design.
+
```
cendhu, line 119: > the underlying gRPC retry policy remains bounded by design

I think we need to give some additional details here. The comment above line `defaultGrpcMaxAttempts = 1024` can be incorporated here with some modification.

utils/connection/client.go
```diff
@@ -344,20 +357,34 @@ func MakeGrpcRetryPolicyJSON(p *retry.Profile) string {
...
+
+	if maxElapsedTime < geometricSum {
+		// The budget is exhausted while the interval is still growing: the smallest n with
+		// initialInterval*(multiplier^n - 1)/(multiplier - 1) > maxElapsedTime.
+		q := 1 + maxElapsedTime*(multiplier-1)/initialInterval
+		return int(math.Floor(math.Log(q)/math.Log(multiplier))) + 1
 	}
-	return attempts
+	// The budget is exhausted during the capped phase, where every interval equals maxInterval.
+	return int(geometricSteps) + int(math.Floor((maxElapsedTime-geometricSum)/maxInterval)) + 1
```
cendhu, line 387: The default max interval is 10 seconds. When we scale up, it would take more than 10 seconds. Maybe we can add this to the deployment guide doc.

cendhu (approved):
LGTM.

docs/deployment-guide.md
liran-funaro: I'm not sure this note should be here. It is a deployment guide. The deployer doesn't care about internal design.

utils/connection/client.go
liran-funaro: The maximal interval means that after this interval value, we stop doing exponential backoff and just wait for each interval with this maximal value.
I don't see how this is affected by scaling up. 
This applies to the deployer only in terms that it is the de facto value that we pull to check if the service has come up again.

docs/deployment-guide.md
cendhu: When I read `the underlying gRPC retry policy remains bounded by design.`, I did not understand what it means or refers too. Only when I read connection.go, I understood what it means. I agree that it is an internal design and hence suggest to remove `; the underlying gRPC retry policy remains bounded by design.` as it occurs suddenly without explaining what this gRPC retry policy is.

utils/connection/client.go
cendhu: When the interval reaches the maximal value, it will continue to wait for maximal value every time. By default, it is set to 10 seconds but configurable by the user. The user should set the maximum interval as per this understanding.

liran-funaro: I'll add this explanation to the deployment guide.

Issue #658, opened 2026-07-15
I-8: [foundations] System namespace handling


cendhu:
# PR-08 — committer: system-namespace recognition + policy-namespace reservation

**Repo**: `fabric-x-committer` (this repo)
**Tasks**: T011, T012, T013, T014, T015, T016 (see `the feature task breakdown`)
**Depends on**: PR-05 (constants), PR-07 (schema present).
**Blocks**: PR-10 (sidecar), and all VC snapshot/checkpoint work — this is the foundational logic that treats `_snapshot`/`_checkpoint` as system namespaces.

## Scope

Make `_snapshot` and `_checkpoint` first-class system namespaces across the dependency graph, the preparer, and policy validation — mirroring existing `_meta`/`_config` handling.

- **T011** — `validateNamespaceIDInPolicy` (`service/verifier/policy/policy.go`): reject `committerpb.SnapshotNamespaceID`/`CheckpointNamespaceID` for ordinary namespace-policy updates (alongside `_meta`/`_config`).
- **T013** — `service/coordinator/dependencygraph/transaction_node.go`: do NOT add the normal `_meta` namespace-version read dependency for these two namespaces (exempt them alongside the existing `_meta`/`_config` check at the `ns.NsId != MetaNamespaceID && ns.NsId != ConfigNamespaceID` guard), per `the applicable API contract` §2.
- **T014** — `service/vc/preparer.go`: skip the `_meta` namespace-version dependency for both; for `_snapshot` bypass read/write prep (no reads/writes); `_checkpoint` retains its single `ReadWrite` for checkpoint-specific handling, per `the applicable API contract` §4.

> **Graphify**: before editing `transaction_node.go`/`preparer.go` dependency construction, run `graphify_explain` on the functions to see callers/tests. After changes, run `graphify update . --force`.

## TDD (tests land in this PR)

- **T012** [P] — `service/verifier/policy/policy_test.go`: success/failure cases that `_snapshot`/`_checkpoint` are rejected as ordinary application namespace policies (mirror `_meta`/`_config`). `make test-no-db`.
- **T015** [P] — `service/coordinator/dependencygraph/transaction_node_test.go`: no `_meta` dependency created for `_snapshot`/`_checkpoint` TXs. `make test-no-db`.
- **T016** [P] — `service/vc/preparer_test.go`: `_snapshot` skips read/write prep; `_checkpoint` retains one `ReadWrite`; no `_meta` dependency. `make test-no-db`.

[...]

PR #685, opened 2026-07-20
[vc, coordinator, verifier] treat _snapshot and _checkpoint as system namespaces

### LINKED ISSUE #657: I-7: [vc] Snapshot schema and checkpoint namespace
# PR-07 — committer: `_snapshot`/`_checkpoint` namespace bootstrapping

**Tasks**: T008, T009, T010 (see `the feature task breakdown`)
**Depends on**: PR-05 (go.mod bump — uses `committerpb.SnapshotNamespaceID`/`committerpb.CheckpointNamespaceID`).
**Blocks**: PR-08 (system-ns logic references the namespace tables), PR-11 (VC writes `_snapshot` rows).

## Scope

Schema-only. DB layer — run the **DB** test target. There is no `snapshot_state` SQL table and no `ns__checkpoint`-specific DDL: both `_snapshot` and `_checkpoint` are ordinary namespaces, bootstrapped through the existing namespace/table creation path (the same `ns_<id>` convention already used for `_meta`/`_config`), no new SQL DDL beyond that standard path, and no timestamp columns anywhere.

- **T008/T009** — ensure `_snapshot` and `_checkpoint` are created via the existing namespace/table creation path in `service/vc/dbinit.go`, using `committerpb.SnapshotNamespaceID` and `committerpb.CheckpointNamespaceID`, following the `ns_<id>` convention (same mechanism as `_meta`/`_config`). No bespoke schema, no `snapshot_state` table, no `snapshot_data` table.

## TDD (test lands in this PR)

- **T010** — `service/vc/dbinit_test.go`: table-driven test asserting the `_snapshot` and `_checkpoint` namespace tables exist after init (same shape/assertions as existing `_meta`/`_config` bootstrap tests).

## Verification

```bash
make test-all-db   # DB-layer change → DB target required
make lint
```

## Done criteria

- Fresh DB init creates both namespace tables via the standard namespace bootstrap path. No new SQL DDL, no timestamp columns.
- Test green under `make test-all-db`.
- Docs update for `_snapshot`/`_checkpoint` namespace storage deferred to PR-18 (T051).

---

cendhu:
#### Type of change

- New feature
 
#### Description

Add committerpb.SnapshotNamespaceID and CheckpointNamespaceID to the systemNamespaces list so their ns_<id> tables are created via the existing namespace bootstrap path, alongside _meta and _config. 

Recognize _snapshot and _checkpoint as first-class system namespaces
across policy validation, the dependency graph, and the VC preparer,
mirroring existing _meta/_config handling.

- Add utils.IsSystemNamespace covering _meta, _config, _snapshot, and
  _checkpoint, and use it in policy validation and dependency-graph
  read-key construction to avoid duplicating the 4-way guard.
- policy: reject _snapshot/_checkpoint as ordinary application namespace
  policies.
- dependencygraph: skip the normal _meta namespace-version read
  dependency for system namespaces.
- vc preparer: skip the _meta dependency for both; forward the _snapshot
  marker TX to the committer as a single new write into the _snapshot
  namespace (key = tx_id, value = SnapshotState{TxRef}); retain the
  single ReadWrite for _checkpoint. Fail the preparer worker rather than
  silently dropping a _snapshot write, since that would fork state across
  deterministic replicas.

#### Related issues

 - resolves #657 and #658

### DIFF
--- a/service/vc/preparer.go
+++ b/service/vc/preparer.go
@@ -186,11 +186,10 @@ func (p *transactionPreparer) prepare(ctx context.Context) error { //nolint:goco
 				// duplicate snapshot requests.
 				if nsOperations.NsId == committerpb.SnapshotNamespaceID {
 					// proto.Marshal cannot fail here: SnapshotState{TxRef} is a small, acyclic
-					// message. We still return the error instead of skipping the TX: the preparer
-					// runs deterministically on every committer, so silently dropping this write
-					// would omit the _snapshot row on one replica while others keep it, forking
-					// state. Failing the worker (via errgroup) halts this committer uniformly
-					// rather than diverging.
+					// message. We return the error only for completeness. proto.Marshal is
+					// deterministic, so any failure would occur identically on every committer
+					// and halt them uniformly (via errgroup); there is no risk of one replica
+					// dropping the _snapshot write while others keep it.
 					state, marshalErr := proto.Marshal(&committerpb.SnapshotState{TxRef: tx.Ref})
 					if marshalErr != nil {
 						return errors.Wrapf(marshalErr, "failed to marshal snapshot state for TX %s", tx.Ref.TxId)
--- a/service/vc/preparer.go
+++ b/service/vc/preparer.go
@@ -10,9 +10,11 @@ import (
 	"context"
 	"time"
 
+	"github.com/cockroachdb/errors"
 	"github.com/hyperledger/fabric-x-common/api/applicationpb"
 	"github.com/hyperledger/fabric-x-common/api/committerpb"
 	"golang.org/x/sync/errgroup"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/hyperledger/fabric-x-committer/api/servicepb"
 	"github.com/hyperledger/fabric-x-committer/utils/channel"
@@ -116,8 +118,7 @@ func (p *transactionPreparer) run(ctx context.Context, numWorkers int) error {
 
 	for range numWorkers {
 		g.Go(func() error {
-			p.prepare(eCtx)
-			return nil
+			return p.prepare(eCtx)
 		})
 	}
 
@@ -129,13 +130,13 @@ func (p *transactionPreparer) run(ctx context.Context, numWorkers int) error {
 // the index of invalid transactions when a read is invalid.
 // It sends the prepared transactions to the preparedTxs channel which is an input
 // to the validator.
-func (p *transactionPreparer) prepare(ctx context.Context) { //nolint:gocognit
+func (p *transactionPreparer) prepare(ctx context.Context) error { //nolint:gocognit
 	incomingTransactionBatch := channel.NewReader(ctx, p.incomingTransactionBatch)
 	outgoingPreparedTransactions := channel.NewWriter(ctx, p.outgoingPreparedTransactions)
 	for {
 		txBatch, ok := incomingTransactionBatch.Read()
 		if !ok {
-			return
+			return nil
 		}
 		logger.Debugf("New batch with %d in the preparer.", len(txBatch.Transactions))
 		start := time.Now()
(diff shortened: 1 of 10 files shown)


service/vc/preparer.go
```diff
@@ -175,6 +176,35 @@ func (p *transactionPreparer) prepare(ctx context.Context) { //nolint:gocognit
...
+				// committer/hash worker fills in the lifecycle status and clone details. The
+				// nil version makes this a new key, and MVCC on tx_id gives idempotency for
+				// duplicate snapshot requests.
+				if nsOperations.NsId == committerpb.SnapshotNamespaceID {
+					// proto.Marshal cannot fail here: SnapshotState{TxRef} is a small, acyclic
+					// message. We still return the error instead of skipping the TX: the preparer
+					// runs deterministically on every committer, so silently dropping this write
+					// would omit the _snapshot row on one replica while others keep it, forking
+					// state. Failing the worker (via errgroup) halts this committer uniformly
+					// rather than diverging.
```
liran-funaro: This comment is incorrect. `proto.Marshal` is deterministic. So it will fail on all parties (no fork).
We can state that will return error for completeness,

utils/utils.go
```diff
@@ -117,6 +117,25 @@ func IsConfigTx(namespaces []*applicationpb.TxNamespace) bool {
...
+	case committerpb.MetaNamespaceID,
+		committerpb.ConfigNamespaceID,
+		committerpb.SnapshotNamespaceID,
+		committerpb.CheckpointNamespaceID:
+		return true
+	default:
+		return false
+	}
+}
+
```
liran-funaro, line 138: I suggest moving this method to the common repository under `committerpb`.

service/vc/preparer_test.go
```diff
@@ -654,6 +655,114 @@ func TestPrepareTx(t *testing.T) { //nolint:maintidx // cannot improve.
...
+			},
+		},
+		invalidTxIDStatus: map[TxID]committerpb.Status{},
+		txIDToHeight: transactionIDToHeight{
+			txs[0]: servicepb.NewHeight(8, 1),
+		},
+	}
+
+	env.txBatch <- tx
+	preparedTxs := <-env.preparedTxs
```
liran-funaro: Use context when reading from channels to avoid hanging in case of a bug.

service/vc/preparer_test.go
```diff
@@ -654,6 +655,114 @@ func TestPrepareTx(t *testing.T) { //nolint:maintidx // cannot improve.
...
+			},
+		},
+		invalidTxIDStatus: map[TxID]committerpb.Status{},
+		txIDToHeight: transactionIDToHeight{
+			txs[0]: servicepb.NewHeight(9, 1),
+		},
+	}
+
+	env.txBatch <- tx
+	preparedTxs := <-env.preparedTxs
```
liran-funaro: Use context when reading from channels to avoid hanging in case of a bug.

utils/utils.go
cendhu: I have a TODO in the comment. Will do in subsequent PRs.

service/vc/preparer.go
cendhu: Reworded. The comment now states proto.Marshal is deterministic (fails identically on all parties, no fork) and that we return the error only for completeness.

service/vc/preparer_test.go
cendhu: Added env.readPrepared(t) which reads via channel.NewReader(t.Context(), ...) and requires ok, so a bug that stalls the preparer fails the test instead of hanging.

service/vc/preparer_test.go
cendhu: Done via env.readPrepared(t) (context-aware read).

service/vc/preparer.go
cendhu: I see. Done.

liran-funaro (approved):

Issue #46, opened 2025-07-06
[loadgen] redability


liran-funaro:
- There are a few places where we can use explicit type instead of an interface. I.e., it is defined as an interface but only ever used with one type

PR #688, opened 2026-07-20
[loadgen] Remove the workload distribution

### LINKED ISSUE #682: [loadgen] removal of the distribution in the loadgen
Please open an issue requesting the removal of the distribution in the loadgen.
We can discuss this there.

_Originally posted by @liran-funaro in https://github.com/hyperledger/fabric-x-committer/pull/676#discussion_r3613135579_

liran-funaro:
#### Type of change

- Improvement
 
#### Description

- The load generator's `Distribution` descriptor supported five variants (`const`, `uniform`, `normal`, `bernoulli`, `discrete`), but only `const` was ever used. Load generation holds each parameter at a fixed value so results are deterministic and comparable across runs, so the unused variants only added config surface, sample-YAML documentation, generator types, and tests with no realized benefit. This collapses `Distribution` down to plain constant values.

- Delete `distributions.go` and the orphaned `Const`/`FloatToInt`/`FloatToPositiveInt`/`FloatToBoolean` generators; `MultiGenerator.Count` is now a plain `int`.
- Replace the `*Distribution` profile fields with plain scalars: read/read-write/blind-write counts and query size become `uint32`, the minimal invalid-keys portion becomes `Probability`, and the dependency gap becomes `uint64` (min 1).
- Keep per-transaction conflict injection (invalid signatures and dependency probability) via a small `bernoulliGenerator` (`rnd.Float64() < p`) with identical semantics to the previous Bernoulli-based path.
- Update the sample `loadgen.yaml` documentation and the affected tests.

#### Related issues

- resolves #682
- address #46


loadgen/workload/config.go
```diff
@@ -25,6 +25,16 @@ const (
...
 
+// Probability is a float in the closed interval [0,1].
+type Probability = float64
+
+const (
+	// Always is 100%.
+	Always Probability = 1
+	// Never is 0%.
+	Never Probability = 0
+)
```
cendhu: We aren't using these constants anywhere.

cendhu (approved):
LGTM. I have left a comment.

loadgen/workload/config.go
liran-funaro: Yeah. I'll remove them. GoLand flags these immediately. VSCode doesn't =/

PR #693, opened 2026-07-20
[loadgen] Validate conflict probabilities are in [0,1]

liran-funaro:
#### Type of change

- Improvement
 
#### Description

- The loadgen config exposes probability fields — `conflicts.invalid-signatures` and each `conflicts.dependencies[].probability` — that must lie in the closed interval [0,1]. Out-of-range values previously decoded silently and produced nonsensical per-transaction Bernoulli behavior. Reject them at config-load time.
- Annotate `ConflictProfile.InvalidSignatures` and `DependencyDescription.Probability` with `validate:"gte=0,lte=1"`, and the `ConflictProfile.Dependencies` slice with `validate:"dive"` so each element's probability is validated.
- Add `TestLoadGenProbabilityValidation` covering in-range acceptance and out-of-range rejection for both fields (including the sliced dependency).

### DIFF
--- a/loadgen/workload/config.go
+++ b/loadgen/workload/config.go
@@ -91,15 +91,15 @@ type TransactionProfile struct {
 // Note that each of the conflicts' probabilities are independent bernoulli distributions.
 type ConflictProfile struct {
 	// Probability of invalid signatures [0,1] (default: 0)
-	InvalidSignatures Probability `mapstructure:"invalid-signatures" yaml:"invalid-signatures"`
+	InvalidSignatures Probability `mapstructure:"invalid-signatures" yaml:"invalid-signatures" validate:"gte=0,lte=1"`
 	// Dependencies list of dependencies
-	Dependencies []DependencyDescription `mapstructure:"dependencies" yaml:"dependencies"`
+	Dependencies []DependencyDescription `mapstructure:"dependencies" yaml:"dependencies" validate:"dive"`
 }
 
 // DependencyDescription describes a dependency type.
 type DependencyDescription struct {
 	// Probability of the dependency type [0,1] (default: 0)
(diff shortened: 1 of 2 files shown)


loadgen/workload/config.go
```diff
@@ -91,15 +91,15 @@ type TransactionProfile struct {
 // Note that each of the conflicts' probabilities are independent bernoulli distributions.
 type ConflictProfile struct {
 	// Probability of invalid signatures [0,1] (default: 0)
-	InvalidSignatures Probability `mapstructure:"invalid-signatures" yaml:"invalid-signatures"`
+	InvalidSignatures Probability `mapstructure:"invalid-signatures" yaml:"invalid-signatures" validate:"gte=0,lte=1"`
 	// Dependencies list of dependencies
-	Dependencies []DependencyDescription `mapstructure:"dependencies" yaml:"dependencies"`
+	Dependencies []DependencyDescription `mapstructure:"dependencies" yaml:"dependencies" validate:"dive"`
```
cendhu, line 96: what does the "dive" mean?

liran-funaro: It tells the validator to descend into the elements of a slice/array/map and apply the rules that follow it — plus each element's own validate field tags — to every element, instead of validating the collection as a whole

cendhu: I see. So without the `dive`, validate within the DependencyDescription won't be exercised? I didn't know this when reviewing other PR added this validate tag. I assume we don't have such a slide, array or map in configurations. 

We can add a comment for our future self.

cendhu (approved):
LGTM

loadgen/workload/config.go
liran-funaro: I plan to remove this code anyway in my next PR.

PR #700, opened 2026-07-23
[sidecar] Validate and split _snapshot/_checkpoint system transactions

### LINKED ISSUE #660: I-10: [sidecar] Snapshot detection and transaction drain
# PR-10 — committer: sidecar `_snapshot`/`_checkpoint` form validation + block-split/drain (US1a)

**Tasks**: T017, T019, T020, T020a, T021 (see `the feature task breakdown`)
**Depends on**: PR-08 (system-namespace recognition), PR-09 (unified sidecar service in place).
**Blocks**: PR-11 (VC processes the snapshot TX the sidecar submits), PR-15 (VC checkpoint verify-before-commit relies on the key already being validated here).

## Scope

Sidecar-only. Detect the `_snapshot` marker TX, and split/drain the block so the snapshot TX is submitted alone after all regular TXs in the block drain, then wait for the VC to commit the snapshot before resuming. Also add the `_checkpoint` read-write-key form check (mirrors `_meta`'s `checkMetaNamespace` pattern).

- **T020** — `service/sidecar/mapping.go`: detect `committerpb.SnapshotNamespaceID` in `mapMessage`, set `isSnapshot=true` on `blockMappingResult`, add the targeted exception so a marker-only `_snapshot` TX (no reads/writes) maps as valid. Ordinary namespaces with no writes must still return `Status_MALFORMED_NO_WRITES`. **Explicit invalidation rule**: a `_snapshot` transaction's read-write set (reads, read-writes, AND blind-writes) MUST be entirely empty to be considered valid marker-only. If the read-write set is NOT empty, the sidecar MUST invalidate the transaction with a dedicated MALFORMED status (distinct from `MALFORMED_NO_WRITES`, e.g. `MALFORMED_SNAPSHOT_NOT_MARKER_ONLY`) rather than silently accepting it as a snapshot request — per `the applicable API contract` §1 and `the feature data model` "Snapshot Transaction Validation".
- **T020a** — `service/sidecar/mapping.go`: add a `checkCheckpointNamespace` form check for `committerpb.CheckpointNamespaceID`, mirroring the existing `checkMetaNamespace` pattern used for `_meta`. **Explicit invalidation rule**: a `_checkpoint` transaction's single read-write key MUST decode as a valid TxHeight (block number + tx number) via the existing order-preserving height encoder (`servicepb.NewHeightFromBytes`). If the key is NOT the TxHeight encoding, the sidecar MUST invalidate the transaction with a dedicated MALFORMED status (e.g. `MALFORMED_CHECKPOINT_INVALID_KEY`) rather than accepting/forwarding it for commit — per the checkpoint transaction contract §1/§4 and `the feature data model` "Checkpoint Transaction Validation". This runs at form-validation time, before the TX ever reaches the VC's verify-before-commit logic in PR-15.
- **T021** — `service/sidecar/relay.go`: block-split + drain in `preProcessBlock` — when `isSnapshot` is set, per `the applicable API contract` §1 step 3:
  1. Submit the regular TXs (if any) to the coordinator.
  2. `waitingTxsSlots.WaitTillEmpty(ctx)` — **first drain**: all in-flight regular TXs complete.
  3. Submit the snapshot TX separately.
  4. `waitingTxsSlots.WaitTillEmpty(ctx)` — **second drain**: wait until the VC returns COMMITTED for the snapshot TX (drain ends on VC COMMITTED, **not** on hash completion).
  5. Resume normal submission.

  Both `WaitTillEmpty` calls are required. On ctx cancellation, return the error.

> **Graphify**: run `graphify_explain` on `mapMessage`/`preProcessBlock` first. After changes, run `graphify update . --force`.

- **T017** [P] — `service/sidecar/mapping_test.go`: marker-only `_snapshot` TX (empty read-write set) → valid with `isSnapshot=true`; `_snapshot` mixed with reads → invalid; `_snapshot` mixed with read-writes → invalid; `_snapshot` mixed with blind-writes → invalid (all three non-empty cases use the dedicated MALFORMED status, never silently accepted); ordinary empty namespace stays `MALFORMED_NO_WRITES`; `_checkpoint` TX with a valid TxHeight key → valid; `_checkpoint` TX with a key that does NOT decode as a TxHeight → invalid with the dedicated MALFORMED status.
- **T019** [P] — `service/sidecar/relay_test.go` (create if absent): a block with regular TXs + a snapshot TX is split so regulars are **submitted and drained**, then the snapshot is **submitted and drained** (assert both drain points via `waitingTxsSlots`). Use `require.Eventually`, never `time.Sleep`.

## Verification

```bash
make test-no-db   # sidecar mapping/relay tests are DB-free
make lint
```

## Done criteria

- Marker-only snapshot TX (entirely empty read-write set) validated; any non-empty read-write set on `_snapshot` rejected with a dedicated MALFORMED status; ordinary no-write namespace still malformed via `MALFORMED_NO_WRITES`.
- `_checkpoint` TX with a non-TxHeight-decodable key rejected with a dedicated MALFORMED status; valid-TxHeight-key `_checkpoint` TX passes form validation.
- Block split correctly; regulars drain, snapshot submits, snapshot drains to VC COMMITTED, then resume — **two** `WaitTillEmpty` calls.

---

cendhu:
#### Type of change

- New feature
- Test update
- Documentation update

#### Description

- Detect the `_snapshot` marker TX in `mapMessage` and validate `_snapshot`/`_checkpoint`
  system-namespace forms in the sidecar:
  - `_snapshot`/`_checkpoint` must be standalone (exactly one namespace); a mixed
    system+user TX returns `MALFORMED_SYSTEM_TX_NOT_STANDALONE`.
  - `_snapshot` must be marker-only (no reads/read-writes/blind-writes), else
    `MALFORMED_SNAPSHOT_NOT_MARKER_ONLY`; an ordinary no-write namespace still returns
    `MALFORMED_NO_WRITES`.
  - `_checkpoint` must have exactly one read-write whose key decodes as a full
    `servicepb.Height` via `NewHeightFromBytes`, else `MALFORMED_CHECKPOINT_INVALID_KEY`.
  - System namespace IDs are recognized during namespace-ID validation while application
    `_config` transactions are still rejected.
- Split-and-drain snapshot blocks in the relay: treat the snapshot TX as a submission
  barrier - submit the regular TXs before it, `WaitTillEmpty`, submit the snapshot alone,
  `WaitTillEmpty` again (drain ends on VC COMMITTED, not on hash completion), then resume.
- Accept only the first `_snapshot` TX in a block; reject any further `_snapshot` in the
  same block with `REJECTED_DUPLICATE_SNAPSHOT_IN_BLOCK` (a stored status), bounding the
  drain to one barrier per block.

- `mapBlock` tracks `hasSnapshot`/`snapshotTxIndex` so the relay splits around the snapshot
  without re-inspecting TX bodies.
- Rejected statuses (which have no TX body) are partitioned around the snapshot by their
  original block position, so a stored post-snapshot reject (e.g. a duplicate snapshot)
  commits after the barrier and does not leak a `tx_status` row into the snapshot clone.
- Each split block is registered and counted active only once, avoiding an
  `activeBlocksCount` over-count across snapshot segments.
- Bumps `fabric-x-common` to the commit introducing the new statuses; behavior documented
  in `docs/sidecar.md`.
- Tested via `service/sidecar` mapping/relay unit tests (form validation, duplicate-in-block,
  segment partitioning for first/middle/last/only positions, and both drain points);
  `make lint` clean.

#### Related issues

- resolves #660

### DIFF
--- a/docs/sidecar.md
+++ b/docs/sidecar.md
@@ -256,8 +256,33 @@ enum Status {
    MALFORMED_MISSING_SIGNATURE = 113;                  // Number of signatures does not match the number of namespaces.
    MALFORMED_NAMESPACE_POLICY_INVALID = 114;           // Invalid namespace policy.
    MALFORMED_CONFIG_TX_INVALID = 115;                  // Invalid configuration transaction.
+   MALFORMED_SNAPSHOT_NOT_MARKER_ONLY = 116;           // `_snapshot` namespace has reads/writes; it must be marker only.
+   MALFORMED_CHECKPOINT_INVALID_KEY = 117;             // `_checkpoint` namespace form or key is invalid.
+   MALFORMED_SYSTEM_TX_NOT_STANDALONE = 118;           // System namespace TX (`_snapshot`/`_checkpoint`) is not standalone.
+   REJECTED_DUPLICATE_SNAPSHOT_IN_BLOCK = 121;         // More than one `_snapshot` TX in a block; only the first is processed, the rest rejected.
 }
 ```
+
+#### System namespace transaction forms
+
+The sidecar applies additional form checks for system namespaces before forwarding transactions to the coordinator.
+
+- `_snapshot` transactions must be standalone: exactly one namespace, and that namespace must be `_snapshot`. The
+  `_snapshot` namespace is a marker only, so it must have no reads, no read-writes, and no blind-writes. A non-empty
+  read-write set is rejected with `MALFORMED_SNAPSHOT_NOT_MARKER_ONLY`; a mixed `_snapshot`+user transaction is
+  rejected with `MALFORMED_SYSTEM_TX_NOT_STANDALONE`. Ordinary user namespaces with no writes still use
+  `MALFORMED_NO_WRITES`.
+- `_checkpoint` transactions must be standalone: exactly one namespace, and that namespace must be `_checkpoint`. The
+  namespace must contain exactly one read-write, no reads-only entries, and no blind writes. The read-write key must
+  decode as a full `servicepb.Height` via `servicepb.NewHeightFromBytes`; valid height prefixes with trailing bytes
+  are rejected. Invalid checkpoint form is rejected with `MALFORMED_CHECKPOINT_INVALID_KEY`; a mixed
+  `_checkpoint`+user transaction is rejected with `MALFORMED_SYSTEM_TX_NOT_STANDALONE`.
+- `_meta` keeps the existing namespace-policy update checks. `_config` is still rejected if submitted as an
+  application transaction namespace.
+- Only the first `_snapshot` transaction in a block is accepted. Any additional `_snapshot` transaction in the same
+  block is rejected with `REJECTED_DUPLICATE_SNAPSHOT_IN_BLOCK` (a stored status, so the outcome is recorded in the
+  state database), regardless of the first snapshot's outcome. This bounds the snapshot drain to at most one barrier
+  per block.
 
 **c. In-Flight Duplicate Transaction Detection**: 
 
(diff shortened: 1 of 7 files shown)


cendhu:
Superseded by #702 (same commit, pushed from fork cendhu/fabric-x-committer instead of an upstream branch).

PR #701, opened 2026-07-23
[vc] Create snapshot clone before committing the _snapshot marker

### LINKED ISSUE #661: I-11: [vc] Snapshot row and database clone
# PR-11 — committer: VC snapshot creation — PENDING row + clone + commit wiring (US1b-i)

**Tasks**: T018, T022, T023, T026 (see `the feature task breakdown`)
**Depends on**: PR-10 (sidecar submits the `_snapshot` TX this consumes).
**Blocks**: PR-12 (hash worker extends `snapshot.go`), PR-15 (checkpoint verifies against the computed hash).

> Read `docs/core-concurrency-pattern.md` before writing any goroutine/clone-lifecycle code.

## Scope

VC-only. First slice of `snapshot.go`: record a PENDING row on `_snapshot` commit, create a native zero-copy DB clone (both DB backends), and wire it into the commit path. **Hash worker + COMPLETED → PR-12. Crash recovery → PR-13. Retention config + cleanup → PR-17.**

- **T022** — create `service/vc/snapshot.go`: on `_snapshot` commit, write a versioned `SnapshotState` row (status=PENDING) into the `_snapshot` namespace keyed by tx_id — `tx_ref = TxRef{block_num: TxRef.BlockNum, tx_num: TxRef.TxNum, tx_id: TxRef.TxID}`, written **before** the clone begins. There are no timestamp fields on `SnapshotState`. Duplicate `tx_id` is ignored idempotently (write to an already-occupied key resolves as an MVCC no-op/conflict, not a duplicate insert). Use `cockroachdb/errors` at origin, `%w` for wrapping.
- **T023** — native zero-copy clone. Before issuing `CREATE DATABASE`, update the `_snapshot` row (status=IN_PROGRESS, clone_database=<name>, e.g. `snapshot_<block_height>`) via a versioned namespace write, so recovery can drop a partial clone by its stored name. Implement **both** backends per `the feature data model` §3 and `the applicable API contract` §4c:
  - **YugabyteDB**: `CREATE DATABASE snapshot_<h> TEMPLATE <src> AS OF <unix_micros>` via a dedicated connection. Source stays live — no termination. Requires `enable_db_clone=true` + snapshot schedule.
  - **PostgreSQL**: hold a **dedicated admin connection outside the pgxpool**; `ALTER DATABASE src ALLOW_CONNECTIONS false` → `SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname='src'` → `CREATE DATABASE snapshot_<h> TEMPLATE <src> STRATEGY=FILE_COPY` → `ALTER DATABASE src ALLOW_CONNECTIONS true`. Reconnect the pool afterward (existing `retry.Profile` backoff bounds the reconnect storm).
- **T026** — wire `_snapshot` into the VC commit path in `service/vc/validator_committer_service.go` (call into `snapshot.go` when `committerpb.SnapshotNamespaceID` is committed), keeping the commit path inline and simple. Return COMMITTED immediately.

> **Graphify**: run `graphify_explain` on the VC commit-path entry (`validator_committer_service.go` committer) before wiring. After adding `snapshot.go` functions, run `graphify update . --force`.

## TDD (test lands in this PR)

- **T018** [P] — create `service/vc/snapshot_test.go`: PENDING-row write (versioned write into `_snapshot` namespace keyed by tx_id, with `tx_ref` set), `clone_database` field set via a versioned write **before** `CREATE DATABASE`, idempotent duplicate `tx_id`. Cover the PostgreSQL clone path where the DB target supports it. (Hash/COMPLETED assertions → PR-12; recovery → PR-13.) Use `require.Eventually` for any async observation, never `time.Sleep`. `make test-all-db`.

## Verification

```bash
make test-all-db   # DB-layer + clone creation
make lint
```

## Done criteria

- `_snapshot` commit writes a versioned PENDING row (via a namespace write, no timestamp fields exist) and creates a named clone; `clone_database` field set via a versioned namespace write before cloning; duplicate tx_id ignored.
- Both YugabyteDB and PostgreSQL clone paths implemented (PostgreSQL: admin conn outside pool, terminate, FILE_COPY, re-allow, pool reconnect).
- Test green under `make test-all-db` (no `time.Sleep`).

---

### LINKED ISSUE #691: move IsSystemNamespace to common repository under `committerpb`.
I suggest moving this method to the common repository under `committerpb`.

_Originally posted by @liran-funaro in https://github.com/hyperledger/fabric-x-committer/pull/685#discussion_r3614855634_

cendhu:
#### Type of change

- New feature
- Documentation update

#### Description

- On `_snapshot` commit, create a native zero-copy DB clone BEFORE the marker's
  txID commits, then rewrite the marker to a PENDING `SnapshotState` carrying the
  clone name — persisted atomically with the txID by the normal commit path
  (invariant: txID committed <=> clone exists <=> PENDING row).
- Clone as of current time via plain `CREATE DATABASE ... TEMPLATE` (no `AS OF`):
  the snapshot TX is drained/standalone, so "now" is the exact cut.
- Both backends: YugabyteDB `TEMPLATE` clone (source stays live; requires
  `enable_db_clone` + a PITR snapshot schedule); PostgreSQL `TEMPLATE ...
  STRATEGY=FILE_COPY` behind the ALLOW_CONNECTIONS/terminate dance.
- Skip cloning when the snapshot txID already exists in `tx_status` (resubmission
  or duplicate) so no orphan clone is left behind; a duplicate clone name is
  treated as reuse. Clones are never dropped here.
- Wire the clone into the committer before the commit retry loop; on clone
  failure the batch is not committed and the coordinator re-drives it.
- Reuse `committerpb.IsSystemNamespace`/`SystemNamespaces` from fabric-x-common
  (bump the dep) and drop the duplicated local helpers.

- `adminExec` runs clone/admin DDL on a dedicated out-of-pool connection with a
  bounded dial timeout and is deliberately not retried (deterministic DDL errors;
  42P04 "already exists" is mapped to success).
- Test-only `EnsureSnapshotScheduleForCloning` provisions the required YugabyteDB
  snapshot schedule per test DB (yb-admin; no SQL API); the container enables
  `enable_db_clone`.
- Docs: added a snapshot subsection to `docs/validator-committer.md` and a
  provisioning/tuning step to `docs/deployment-guide.md` noting the schedule is
  mandatory and retention should be kept short.
- Verified with `make test-all-db` (YugabyteDB snapshot suite) and `make lint`.

#### Related issues

- resolves #661
- resolves #691

### DIFF
--- a/service/vc/snapshot.go
+++ b/service/vc/snapshot.go
@@ -1,0 +1,279 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package vc
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"strings"
+	"time"
+
+	"github.com/cockroachdb/errors"
+	"github.com/hyperledger/fabric-x-common/api/committerpb"
+	"github.com/jackc/pgerrcode"
+	"github.com/yugabyte/pgx/v5"
+	"github.com/yugabyte/pgx/v5/pgconn"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/hyperledger/fabric-x-committer/utils/statedb"
+)
+
+// maintenanceDBName is the neutral admin database the clone runs against.
+// A session cannot CREATE/DROP the database it is connected to on EITHER
+// backend, so both PostgreSQL and YugabyteDB need a separate always-present DB
+// to issue CREATE DATABASE from; that is what this connects to.
+//
+// "postgres" is chosen because it is created by default on both backends
+// (YugabyteDB ships a "postgres" database for PG compatibility) and is never
+// the clone source, so it is never locked out by the PostgreSQL
+// ALLOW_CONNECTIONS dance. The extra PostgreSQL requirement that the TEMPLATE
+// source be free of other sessions is handled in createPostgresSnapshotDatabase, not here,
+// and does not apply to YugabyteDB (DocDB cloning keeps the source live).
+const maintenanceDBName = "postgres"
+
+// createSnapshotIfPresent detects a _snapshot record in the batch's
+// per-transaction new-writes and, BEFORE the batch is committed, creates the
+// snapshot database and rewrites the record's value to a PENDING SnapshotState
+// carrying that database name. The rewritten value is then persisted atomically
+// with the snapshot txID by the normal db.commit path, giving the invariant
+// txID committed <=> snapshot database exists <=> PENDING record.
+//
+// The sidecar drains before and after a snapshot TX, so it is submitted
+// standalone: its batch holds exactly one transaction (one new-write entry) and
+// the preparer adds exactly one _snapshot record (key = tx_id) for it. We short-
+// circuit unless newWrites has exactly one entry and act on that single record
+// rather than scanning every write.
+//
+// The incoming record value carries only TxRef with status UNSPECIFIED (the
+// preparer sets no status). This function is called exactly once per batch,
+// before the committer's retry loop, so it neither reads a status back from the
+// _snapshot table nor re-observes its own PENDING rewrite.
+//
+// Snapshot database creation MUST succeed before txID is committed. On failure
+// this returns an error, batch is not committed, txID stays uncommitted, and
+// coordinator retries snapshot (this PR does not self-recover).
+func (db *database) createSnapshotIfPresent(ctx context.Context, newWrites transactionToWrites) error {
+	// A snapshot TX is submitted standalone: the sidecar drains before and after it,
+	// so its batch contains exactly one transaction and hence one new-write entry.
+	// Any other count means there is no _snapshot record to act on.
+	if len(newWrites) != 1 {
+		return nil
+	}
+	for _, nsWrites := range newWrites {
+		w := nsWrites[committerpb.SnapshotNamespaceID]
+		if w.empty() {
+			return nil
+		}
+		// Exactly one key: the preparer adds a single _snapshot record per snapshot TX.
+		snapshotState, err := db.createSnapshotDatabaseAndRewriteRecord(ctx, w.keys[0], w.values[0])
+		if err != nil {
+			return err
+		}
+		if snapshotState != nil {
+			w.values[0] = snapshotState
+		}
+	}
+	return nil
+}
+
+// createSnapshotDatabaseAndRewriteRecord decodes one _snapshot record, creates
+// or reuses its snapshot database when needed, and returns a rewritten PENDING
+// record. A nil result means leave recordValue unchanged.
+//
+// The tx_status lookup and CREATE-only database operation handle these cases:
+//
+//	snapshot case                                txID in tx_status  database action            record result
+//	fresh snapshot                               no                 create                    rewritten PENDING
+//	retry after failed database creation         no                 create                    rewritten PENDING
+//	retry after database creation, before commit no                 CREATE; duplicate = reuse rewritten PENDING
+//	resubmission or duplicate txID               yes                do not create or reuse     unchanged
+//
+// A txID already in tx_status may be a same-height resubmission or a
+// different-height duplicate. The normal commit path resolves that distinction;
+// this function must not create a snapshot database for either case.
+func (db *database) createSnapshotDatabaseAndRewriteRecord(
+	ctx context.Context, key, recordValue []byte,
+) ([]byte, error) {
+	var state committerpb.SnapshotState
+	if err := proto.Unmarshal(recordValue, &state); err != nil {
+		return nil, errors.Wrapf(err, "failed to decode _snapshot record for key %s", key)
+	}
+	ref := state.TxRef
+	if ref == nil {
+		return nil, errors.Newf("_snapshot record for key %s has no TxRef", key)
+	}
+
+	// Skip database creation unless this is first-ever submission (txID absent
+	// from tx_status). If txID exists at SAME height it is a resubmission whose
+	// snapshot database was created in a prior life; if it exists at DIFFERENT
+	// height snapshot TX is rejected as duplicate and must not leave an orphan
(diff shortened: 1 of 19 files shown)


service/vc/snapshot.go
```diff
@@ -0,0 +1,279 @@
...
+
+// maintenanceDBName is the neutral admin database the clone runs against.
+// A session cannot CREATE/DROP the database it is connected to on EITHER
+// backend, so both PostgreSQL and YugabyteDB need a separate always-present DB
+// to issue CREATE DATABASE from; that is what this connects to.
+//
+// "postgres" is chosen because it is created by default on both backends
+// (YugabyteDB ships a "postgres" database for PG compatibility) and is never
+// the clone source, so it is never locked out by the PostgreSQL
+// ALLOW_CONNECTIONS dance. The extra PostgreSQL requirement that the TEMPLATE
```
liran-funaro: dance?

service/vc/snapshot.go
```diff
@@ -0,0 +1,279 @@
...
+		return err
+	}
+
+	sql := fmt.Sprintf("CREATE DATABASE %s TEMPLATE %s STRATEGY=FILE_COPY", clone, src)
+	return ignoreDuplicateDatabase(db.adminExec(ctx, sql))
+}
+
+// adminExec opens a short-lived dedicated connection to the maintenance DB
+// (outside the pgxpool) and runs a single statement. Used for CREATE DATABASE
+// and the PostgreSQL ALTER DATABASE dance, which cannot run on the source pool.
```
liran-funaro, line 227: dance?

service/vc/snapshot.go
```diff
@@ -0,0 +1,279 @@
...
+		return errors.Wrap(err, "failed to parse maintenance-db DSN")
+	}
+	dialer := &net.Dialer{Timeout: 10 * time.Second}
+	connConfig.DialFunc = dialer.DialContext
+
+	conn, err := pgx.ConnectConfig(ctx, connConfig)
+	if err != nil {
+		return errors.Wrap(err, "failed to open maintenance-db admin connection")
+	}
+	defer func() { _ = conn.Close(ctx) }()
```
liran-funaro, line 265:
```suggestion
	defer connection.CloseConnectionsLog(conn)
```

service/vc/snapshot.go
```diff
@@ -0,0 +1,279 @@
...
+
+	if _, err := conn.Exec(ctx, sql); err != nil {
+		return errors.Wrapf(err, "failed to execute admin statement [%s]", sql)
+	}
+	return nil
```
liran-funaro: nit: Wrapf returns `nil` for `nil` errors.
```suggestion
	_, err := conn.Exec(ctx, sql)
	return errors.Wrapf(err, "failed to execute admin statement [%s]", sql)
```

service/vc/snapshot_test.go
```diff
@@ -0,0 +1,253 @@
...
+	// First creation succeeds and snapshot database exists.
+	require.NoError(t, env.DB.createSnapshotDatabase(ctx, name))
+	require.True(t, cloneExists(t, env.DB, name))
+
+	// Second creation over existing database is a no-op success (reuse), not drop+recreate.
+	require.NoError(t, env.DB.createSnapshotDatabase(ctx, name))
+	require.True(t, cloneExists(t, env.DB, name))
+}
+
+func cloneExists(t *testing.T, db *database, name string) bool {
```
liran-funaro, line 108: Do we also want to verify that the clone contains the same data as the original?

liran-funaro (approved):
Note: It is hard to review these small PRs without a wider context. It looks OK code wise, but I have no clarify into the full flow. I might find issues later when the full picture unravel.

cendhu:
> LGTM. Just some minor comments. Note: It is hard to review these small PRs without a wider context. It looks OK code wise, but I have no clarify into the full flow. I might find issues later when the full picture unravel.

I will push other important PRs too. This will give clarity into the full flow.

service/vc/snapshot.go
cendhu: Spelled it out in 72433f2b: the comment now names the sequence explicitly (`ALLOW_CONNECTIONS false / terminate-backends / ALLOW_CONNECTIONS true`) instead of calling it a "dance", and explains why `maintenanceDBName` avoids being caught by it.

service/vc/snapshot.go
cendhu: Same fix in 72433f2b: `createPostgresSnapshotDatabase`'s comment now spells out the three-step ALLOW_CONNECTIONS/terminate-backends/ALLOW_CONNECTIONS sequence instead of "dance".

service/vc/snapshot.go
cendhu: Checked this: `connection.CloseConnectionsLog[T io.Closer]` requires `Close() error`, but `*pgx.Conn.Close` here is `Close(context.Context) error`, so it doesn't satisfy `io.Closer` and the suggestion doesn't compile:

```
T (type *pgx.Conn) does not satisfy io.Closer (wrong type for method Close)
    have Close(context.Context) error
    want Close() error
```

Left the current `defer func() { _ = conn.Close(ctx) }()` in place. Open to a small inline log-on-error wrapper here if you'd like parity with `CloseConnectionsLog`'s logging behavior, but it can't reuse that helper directly for a `pgx.Conn`.

service/vc/snapshot.go
cendhu: Applied in 72433f2b.

service/vc/snapshot_test.go
cendhu: Added in 72433f2b: `TestCreateSnapshotDatabase` now writes distinguishable rows into two source namespaces before cloning and reads them back from a pool connected directly to the clone database, for both the fresh-create and reuse paths — so it verifies actual data, not just that the database exists.

PR #702, opened 2026-07-23
[sidecar] Validate and split _snapshot/_checkpoint system transactions

### LINKED ISSUE #660: I-10: [sidecar] Snapshot detection and transaction drain

### LINKED ISSUE #691: move IsSystemNamespace to common repository under `committerpb`.

cendhu:
#### Type of change

- New feature
- Documentation update

#### Description

- On `_snapshot` commit, create a native zero-copy DB clone BEFORE the marker's
  txID commits, then rewrite the marker to a PENDING `SnapshotState` carrying the
  clone name — persisted atomically with the txID by the normal commit path
  (invariant: txID committed <=> clone exists <=> PENDING row).
- Clone as of current time via plain `CREATE DATABASE ... TEMPLATE` (no `AS OF`):
  the snapshot TX is drained/standalone, so "now" is the exact cut.
- Both backends: YugabyteDB `TEMPLATE` clone (source stays live; requires
  `enable_db_clone` + a PITR snapshot schedule); PostgreSQL `TEMPLATE ...
  STRATEGY=FILE_COPY` behind the ALLOW_CONNECTIONS/terminate dance.
- Skip cloning when the snapshot txID already exists in `tx_status` (resubmission
  or duplicate) so no orphan clone is left behind; a duplicate clone name is
  treated as reuse. Clones are never dropped here.
- Wire the clone into the committer before the commit retry loop; on clone
  failure the batch is not committed and the coordinator re-drives it.
- Reuse `committerpb.IsSystemNamespace`/`SystemNamespaces` from fabric-x-common
  (bump the dep) and drop the duplicated local helpers.

- `adminExec` runs clone/admin DDL on a dedicated out-of-pool connection with a
  bounded dial timeout and is deliberately not retried (deterministic DDL errors;
  42P04 "already exists" is mapped to success).
- Test-only `EnsureSnapshotScheduleForCloning` provisions the required YugabyteDB
  snapshot schedule per test DB (yb-admin; no SQL API); the container enables
  `enable_db_clone`.
- Docs: added a snapshot subsection to `docs/validator-committer.md` and a
  provisioning/tuning step to `docs/deployment-guide.md` noting the schedule is
  mandatory and retention should be kept short.
- Verified with `make test-all-db` (YugabyteDB snapshot suite) and `make lint`.

#### Related issues

- resolves #660 
- resolves #691

### DIFF
--- a/docs/sidecar.md
+++ b/docs/sidecar.md
@@ -256,8 +256,33 @@ enum Status {
    MALFORMED_MISSING_SIGNATURE = 113;                  // Number of signatures does not match the number of namespaces.
    MALFORMED_NAMESPACE_POLICY_INVALID = 114;           // Invalid namespace policy.
    MALFORMED_CONFIG_TX_INVALID = 115;                  // Invalid configuration transaction.
+   MALFORMED_SNAPSHOT_NOT_MARKER_ONLY = 116;           // `_snapshot` namespace has reads/writes; it must be marker only.
+   MALFORMED_CHECKPOINT_INVALID_KEY = 117;             // `_checkpoint` namespace form or key is invalid.
+   MALFORMED_SYSTEM_TX_NOT_STANDALONE = 118;           // System namespace TX (`_snapshot`/`_checkpoint`) is not standalone.
+   REJECTED_DUPLICATE_SNAPSHOT_IN_BLOCK = 121;         // More than one `_snapshot` TX in a block; only the first is processed, the rest rejected.
 }
 ```
+
+#### System namespace transaction forms
+
+The sidecar applies additional form checks for system namespaces before forwarding transactions to the coordinator.
+
+- `_snapshot` transactions must be standalone: exactly one namespace, and that namespace must be `_snapshot`. The
+  `_snapshot` namespace is a marker only, so it must have no reads, no read-writes, and no blind-writes. A non-empty
+  read-write set is rejected with `MALFORMED_SNAPSHOT_NOT_MARKER_ONLY`; a mixed `_snapshot`+user transaction is
+  rejected with `MALFORMED_SYSTEM_TX_NOT_STANDALONE`. Ordinary user namespaces with no writes still use
+  `MALFORMED_NO_WRITES`.
+- `_checkpoint` transactions must be standalone: exactly one namespace, and that namespace must be `_checkpoint`. The
+  namespace must contain exactly one read-write, no reads-only entries, and no blind writes. The read-write key must
+  decode as a full `servicepb.Height` via `servicepb.NewHeightFromBytes`; valid height prefixes with trailing bytes
+  are rejected. Invalid checkpoint form is rejected with `MALFORMED_CHECKPOINT_INVALID_KEY`; a mixed
+  `_checkpoint`+user transaction is rejected with `MALFORMED_SYSTEM_TX_NOT_STANDALONE`.
+- `_meta` keeps the existing namespace-policy update checks. `_config` is still rejected if submitted as an
+  application transaction namespace.
+- Only the first `_snapshot` transaction in a block is accepted. Any additional `_snapshot` transaction in the same
+  block is rejected with `REJECTED_DUPLICATE_SNAPSHOT_IN_BLOCK` (a stored status, so the outcome is recorded in the
+  state database), regardless of the first snapshot's outcome. This bounds the snapshot drain to at most one barrier
+  per block.
 
 **c. In-Flight Duplicate Transaction Detection**: 
 
@@ -315,6 +340,26 @@ When a block satisfies the criteria outlined in step f, the relay component perf
 
 2. Creates a `committedBlockWithTxs` structure containing the block number, transaction list, and
    status information, and sends it to the notification service for distribution to subscribers.
+
+#### Snapshot split and drain
+
+When a valid `_snapshot` marker transaction appears in a block, the sidecar treats it as a submission barrier:
+
+1. Submit regular transactions that appear before the snapshot to the coordinator.
+2. Wait for `waitingTxsSlots.WaitTillEmpty(ctx)` so all in-flight regular transactions complete.
+3. Submit the snapshot transaction as a one-transaction coordinator batch, preserving its original block number and
+   transaction number.
+4. Wait for `waitingTxsSlots.WaitTillEmpty(ctx)` again so the snapshot transaction reaches a final status from the VC.
+5. Resume submission of later transactions and later blocks.
+
+The drain ends when the sidecar receives the snapshot transaction status, normally `COMMITTED`. It does not wait for
+background snapshot hash computation in the validator-committer.
+
+At most one snapshot barrier is applied per block: only the first `_snapshot` transaction in a block is accepted, so
+the block is split into at most three segments — the regular transactions before the snapshot, the snapshot
+transaction alone, and the regular transactions after it. When the snapshot is the first or last transaction (and
+there are no rejected statuses to carry on the empty side), the empty leading or trailing segment is omitted, leaving
+two segments, or a single snapshot-only segment when the block contains nothing else.
 
 ### Task 3. Persisting Committed Block in the File System
 
--- a/docs/sidecar.md
+++ b/docs/sidecar.md
@@ -343,9 +343,11 @@ When a block satisfies the criteria outlined in step f, the relay component perf
 
 #### Snapshot split and drain
 
-When a valid `_snapshot` marker transaction appears in a block, the sidecar treats it as a submission barrier:
+When a valid `_snapshot` marker transaction appears in a block, the sidecar treats it as a submission barrier and
+always sends the snapshot transaction last, regardless of where it originally appeared in the block:
 
-1. Submit regular transactions that appear before the snapshot to the coordinator.
+1. Submit every other transaction in the block (regardless of whether it originally preceded or followed the
+   snapshot) to the coordinator.
 2. Wait for `waitingTxsSlots.WaitTillEmpty(ctx)` so all in-flight regular transactions complete.
 3. Submit the snapshot transaction as a one-transaction coordinator batch, preserving its original block number and
    transaction number.
@@ -355,11 +357,15 @@ When a valid `_snapshot` marker transaction appears in a block, the sidecar trea
 The drain ends when the sidecar receives the snapshot transaction status, normally `COMMITTED`. It does not wait for
 background snapshot hash computation in the validator-committer.
 
+Sending the snapshot transaction last means the snapshot always reflects the full effect of its block, since it is
+committed only after every other transaction in that block has already committed. This is why the snapshot is always
+submitted last, not necessarily in its original position: whether it originally appeared first, in the middle, or
+last among the block's transactions, its outcome must still capture the whole block's writes.
+
 At most one snapshot barrier is applied per block: only the first `_snapshot` transaction in a block is accepted, so
-the block is split into at most three segments — the regular transactions before the snapshot, the snapshot
-transaction alone, and the regular transactions after it. When the snapshot is the first or last transaction (and
-there are no rejected statuses to carry on the empty side), the empty leading or trailing segment is omitted, leaving
-two segments, or a single snapshot-only segment when the block contains nothing else.
+the block is split into at most two segments — one segment carrying every other transaction and rejected status in
+the block (in original order), followed by the snapshot transaction alone. When the block contains nothing but the
+snapshot, the first segment is omitted, leaving a single snapshot-only segment.
 
 ### Task 3. Persisting Committed Block in the File System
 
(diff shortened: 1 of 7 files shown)


service/sidecar/relay.go
```diff
@@ -146,28 +145,184 @@ func (r *relay) preProcessBlock(
...
+	if !mappedBlock.hasSnapshot && !mappedBlock.isConfig {
+		// Common case: an ordinary user block with no submission barrier.
+		r.queueMappedBlock(ctx, queue, mappedBlock)
+		return nil
+	}
+
+	if mappedBlock.isConfig {
+		return r.submitConfigBlock(ctx, queue, block, mappedBlock)
+	}
+	return r.submitSnapshotBlock(ctx, queue, mappedBlock)
```
liran-funaro: It can be clearer using a switch
```suggestion
	switch {
	case mappedBlock.isConfig:
		return r.submitConfigBlock(ctx, queue, block, mappedBlock)
	case mappedBlock.hasSnapshot:
		return r.submitSnapshotBlock(ctx, queue, mappedBlock)
	default: // Common case: an ordinary user block with no submission barrier.
		r.queueMappedBlock(ctx, queue, mappedBlock)
		return nil
	}
```

service/sidecar/relay.go
```diff
@@ -146,28 +145,184 @@ func (r *relay) preProcessBlock(
...
+	promutil.AddToCounter(r.metrics.transactionInThroughput, txsCount)
+	r.waitingTxsSlots.Acquire(ctx, int64(txsCount))
+	promutil.AddToGauge(r.metrics.waitingTransactionsQueueSize, txsCount)
+	queue.Write(mappedBlock)
+}
+
+func splitSnapshotMappedBlock(mappedBlock *blockMappingResult) []*blockMappingResult {
+	if !mappedBlock.hasSnapshot {
+		return []*blockMappingResult{mappedBlock}
+	}
```
liran-funaro: this is dead code. we should remove this check.

service/sidecar/relay.go
```diff
@@ -146,28 +145,184 @@ func (r *relay) preProcessBlock(
...
+
+// submitSnapshotBlock splits a snapshot block into segments and submits them in order. The
+// single snapshot segment is a submission barrier: earlier regular transactions are drained
+// before it, and its status is drained after it, before later transactions are submitted.
+func (r *relay) submitSnapshotBlock(
+	ctx context.Context,
+	queue channel.Writer[*blockMappingResult],
+	mappedBlock *blockMappingResult,
+) error {
+	for _, segment := range splitSnapshotMappedBlock(mappedBlock) {
```
liran-funaro: This is a design comment. I'm sorry to bring this up so late in the process.
We could have decided that a snapshot TX splits at the end of the block that the TX resides on.
Meaning, even if it is in the start or the middle of the block, it is committed last.
We queue the entire block except this TX, then submit the snapshot alone, then continue to process the next block.
This will simplify the split mechanis.
What do you thinkg?

cendhu: Good idea. We can do that. 

> I'm sorry to bring this up so late in the process.

Approved design is only tentative. We can adjust it if there is a simplification or better solution.

cendhu: Done.

docs/sidecar.md
```diff
@@ -256,9 +256,34 @@ enum Status {
...
+  `_snapshot` namespace is a marker only, so it must have no reads, no read-writes, and no blind-writes. A non-empty
+  read-write set is rejected with `MALFORMED_SNAPSHOT_NOT_MARKER_ONLY`; a mixed `_snapshot`+user transaction is
+  rejected with `MALFORMED_SYSTEM_TX_NOT_STANDALONE`. Ordinary user namespaces with no writes still use
+  `MALFORMED_NO_WRITES`.
+- `_checkpoint` transactions must be standalone: exactly one namespace, and that namespace must be `_checkpoint`. The
+  namespace must contain exactly one read-write, no reads-only entries, and no blind writes. The read-write key must
+  decode as a full `servicepb.Height` via `servicepb.NewHeightFromBytes`; valid height prefixes with trailing bytes
+  are rejected. Invalid checkpoint form is rejected with `MALFORMED_CHECKPOINT_INVALID_KEY`; a mixed
+  `_checkpoint`+user transaction is rejected with `MALFORMED_SYSTEM_TX_NOT_STANDALONE`.
+- `_meta` keeps the existing namespace-policy update checks. `_config` is still rejected if submitted as an
```
liran-funaro:
```suggestion
- `_meta` keeps the existing namespace-policy update checks. `_config` is rejected if submitted as an
```

service/sidecar/relay.go
```diff
@@ -184,8 +285,19 @@ func (r *relay) sendBlocksToCoordinator(
 		}
 
 		startTime := time.Now()
-		r.blkNumToBlkWithStatus.Store(mappedBlock.blockNumber, mappedBlock.withStatus)
-		r.activeBlocksCount.Add(1)
+		// A snapshot block is split into multiple segments that share the same block number and
```
liran-funaro:
```suggestion
		// A snapshot block is split into two segments that share the same block number and
```

service/sidecar/relay.go
```diff
@@ -146,28 +145,130 @@ func (r *relay) preProcessBlock(
...
+	segments := make([]*blockMappingResult, 0, 2)
+	if len(mappedBlock.block.Txs) > 0 || len(mappedBlock.block.Rejected) > 0 {
+		segments = append(segments, &blockMappingResult{
+			blockNumber: mappedBlock.blockNumber,
+			block: &servicepb.CoordinatorBatch{
+				Txs:      mappedBlock.block.Txs,
+				Rejected: mappedBlock.block.Rejected,
+			},
+			withStatus:   mappedBlock.withStatus,
+			txIDToHeight: mappedBlock.txIDToHeight,
```
liran-funaro: Here and below: `txIDToHeight` field is only used when constructing `blockMappingResult`. It is then ignored.
There is no reason to assign it here.
For future reference, I suggest adding a comment above this field that explains this.

service/sidecar/relay.go
```diff
@@ -146,28 +145,130 @@ func (r *relay) preProcessBlock(
...
+		r.queueMappedBlock(ctx, queue, segment)
+
+		if segment.snapshotTx != nil {
+			if err := r.drain(ctx); err != nil {
+				return err
+			}
 		}
 	}
+
+	return nil
```
liran-funaro: The split and for loop is redundant and confusing.
```suggestion
	if len(mappedBlock.block.Txs) > 0 || len(mappedBlock.block.Rejected) > 0 {
		r.queueMappedBlock(ctx, queue, &blockMappingResult{
			blockNumber: mappedBlock.blockNumber,
			block: &servicepb.CoordinatorBatch{
				Txs:      mappedBlock.block.Txs,
				Rejected: mappedBlock.block.Rejected,
			},
			withStatus:   mappedBlock.withStatus,
		})
	}

	if err := r.drain(ctx); err != nil {
		return err
	}
	r.queueMappedBlock(ctx, queue, &blockMappingResult{
		blockNumber: mappedBlock.blockNumber,
		block: &servicepb.CoordinatorBatch{
			Txs: []*servicepb.TxWithRef{mappedBlock.snapshotTx},
		},
		withStatus:   mappedBlock.withStatus,
	})
	return r.drain(ctx)
}

docs/sidecar.md
cendhu: Fixed.

service/sidecar/relay.go
cendhu: Fixed.

service/sidecar/relay.go
cendhu: Dropped it from the segment construction in splitSnapshotMappedBlock (now inlined into submitSnapshotBlock), and added a note on the txIDToHeight field itself explaining it's only used while building the original blockMappingResult via mapBlock/addTxIDMapping, never read back off the struct afterwards.

service/sidecar/relay.go
cendhu: Inlined per your suggestion; splitSnapshotMappedBlock is removed and submitSnapshotBlock now does the two-segment queue/drain directly.

liran-funaro (approved):
LGTM

PR #706, opened 2026-07-23
[sidecar, query] Add metric for active client connections

### LINKED ISSUE #705: [sidecar, query] Add metric for active client connections
Currently, we have no way to observe how many clients are connected to a given service.

This information is useful to operators: to confirm that clients connected successfully, and to spot a service under heavy connection load that may need to be scaled out.

To address this, we will add a gRPC server handler that uses Prometheus to count the active client connections, exposed on the service's existing metrics endpoint.

#### Approach: A gRPC stats.Handler backed by a gauge

Register a gRPC `stats.Handler` on the server that keeps a Prometheus gauge in sync with the connection lifecycle: increment on `ConnBegin` (a client connects) and decrement on `ConnEnd` (the connection is torn down). The gauge therefore always reflects the number of connections currently open on the server, and is scraped from the service's existing `/metrics` endpoint (e.g. `sidecar_grpc_active_connections`).

dean-amar:
#### Type of change

- New feature
- Improvement (improvement to code)

#### Description

- Add a gRPC `stats.Handler` (`serve.ConnStatsHandler`) that tracks the number of
  open server-side connections in a Prometheus gauge - incrementing on `ConnBegin`
  and decrementing on `ConnEnd`. It is attached to every gRPC server.
- A service opts in by registering a gauge from its own provider via
  `serve.RegisterConnStatHandler` in `RegisterService`. Sidecar and query do so,
  exposing `sidecar_grpc_active_connections` and `queryservice_grpc_active_connections`
  on their existing metrics endpoints.
- The gauge is held in an `atomic.Pointer`, so registering it is safe even if it
  races the server handling connections.

#### Related issues

- resolves #705

### DIFF
--- a/service/query/metrics.go
+++ b/service/query/metrics.go
@@ -42,6 +42,7 @@ type perfMetrics struct {
 	batchResponseSize               prometheus.Histogram
 	requestAssignmentLatencySeconds prometheus.Histogram
 	queryLatencySeconds             prometheus.Histogram
+	serverConnections               prometheus.Gauge
 }
 
 func newQueryServiceMetrics() *perfMetrics {
@@ -67,6 +68,12 @@ func newQueryServiceMetrics() *perfMetrics {
 			Subsystem: "grpc",
 			Name:      "key_requested_total",
 			Help:      "Number of keys requested by the service",
+		}),
+		serverConnections: p.NewGauge(prometheus.GaugeOpts{
+			Namespace: "queryservice",
+			Subsystem: "grpc",
+			Name:      "active_connections",
+			Help:      "Number of client connections currently open on the server",
 		}),
 		keysResponded: p.NewCounter(prometheus.CounterOpts{
 			Namespace: "queryservice",
--- a/service/query/metrics.go
+++ b/service/query/metrics.go
@@ -69,11 +69,9 @@ func newQueryServiceMetrics() *perfMetrics {
 			Name:      "key_requested_total",
 			Help:      "Number of keys requested by the service",
 		}),
-		serverConnections: p.NewGauge(prometheus.GaugeOpts{
+		serverConnections: monitoring.NewConnectionStatsMetrics(p, monitoring.MetricsParameters{
 			Namespace: "queryservice",
 			Subsystem: "grpc",
-			Name:      "active_connections",
-			Help:      "Number of client connections currently open on the server",
 		}),
 		keysResponded: p.NewCounter(prometheus.CounterOpts{
 			Namespace: "queryservice",
--- a/utils/serve/conn_stats.go
+++ b/utils/serve/conn_stats.go
@@ -1,0 +1,71 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package serve
+
+import (
+	"context"
+	"sync/atomic"
+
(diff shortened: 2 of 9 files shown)


utils/serve/conn_stats.go
```diff
@@ -0,0 +1,71 @@
...
+//
+// The gauge is held in an atomic pointer, so RegisterConnStatHandler is safe to
+// call while or after the server starts serving connections.
+type ConnStatsHandler struct {
+	activeConnections atomic.Pointer[prometheus.Gauge]
+}
+
+func newConnStatsHandler() *ConnStatsHandler {
+	return &ConnStatsHandler{}
+}
```
liran-funaro: minor: redundant constructor.

service/query/metrics.go
```diff
@@ -68,6 +69,12 @@ func newQueryServiceMetrics() *perfMetrics {
 			Name:      "key_requested_total",
 			Help:      "Number of keys requested by the service",
 		}),
+		serverConnections: p.NewGauge(prometheus.GaugeOpts{
+			Namespace: "queryservice",
+			Subsystem: "grpc",
+			Name:      "active_connections",
+			Help:      "Number of client connections currently open on the server",
+		}),
```
liran-funaro: This part can be reused in the sidecar. Please add a common metric method.

utils/serve/conn_stats.go
```diff
@@ -0,0 +1,71 @@
...
+	return ctx
+}
+
+// TagRPC is required by stats.Handler; it is a no-op.
+func (*ConnStatsHandler) TagRPC(ctx context.Context, _ *stats.RPCTagInfo) context.Context {
+	return ctx
+}
+
+// HandleRPC is required by stats.Handler; RPC-level stats are not tracked.
+func (*ConnStatsHandler) HandleRPC(context.Context, stats.RPCStats) {}
```
liran-funaro, line 67: What other metrics we can track using these calls? It is worth investigating.

utils/serve/start_serve.go
```diff
@@ -133,7 +137,9 @@ func NewServers(ctx context.Context, conf *Config) (s Servers, err error) {
 		return s, errors.Wrap(err, "failed to create TLS provider")
 	}
 
-	s.GRPC, err = newGRPCServer(&conf.GRPC, s.GrpcTLSProvider)
+	s.ConnStatsHandler = newConnStatsHandler()
```
liran-funaro:
```suggestion
	s.ConnStatsHandler = &ConnStatsHandler{}
```

liran-funaro (commented):
Looks good. Only minor comments.
I suggest exploring all the metrics we can track using this handler.

utils/serve/start_serve.go
dean-amar: Done.

utils/serve/conn_stats.go
dean-amar: Done.

service/query/metrics.go
dean-amar: Done.

utils/serve/conn_stats.go
dean-amar: I will elaborate on this in a dedicated GitHub issue.

liran-funaro (approved):
LGTM

PR #708, opened 2026-07-28
[vc] Reject snapshot request while prior snapshot not CHECKPOINTED

### LINKED ISSUE #707: I-11b: [vc] Reject snapshot request while a prior snapshot is not CHECKPOINTED
# I-11b: [vc] Reject snapshot request while a prior snapshot is not CHECKPOINTED

Follow-up to #661 (I-11: Snapshot row and database clone). Part of EPIC #571 —
Snapshot and Checkpointing.

## The gap

I-11 lets the VC create a snapshot database and commit a PENDING `_snapshot`
record for any incoming snapshot request, with no check on prior snapshots.
Nothing currently stops a second snapshot request from being accepted while an
earlier one is still PENDING/IN_PROGRESS/COMPLETED (i.e. not yet
CHECKPOINTED). That leaves more than one snapshot lifecycle active at once,
which #662 (I-12: background snapshot hash worker) is not designed to
tolerate — it assumes the prior snapshot is CHECKPOINTED before a new one
starts hashing.

## Proposal

Gate a new `_snapshot` request in `service/vc/snapshot.go`, before it creates
a snapshot database or commits a `_snapshot` record:

- Track the most recently accepted snapshot via a lightweight pointer, and
  look up its current state.
- Accept the incoming request only when the latest snapshot record is
  `CHECKPOINTED`, or no snapshot has ever been accepted.
- Otherwise reject the incoming snapshot transaction — without creating a
  snapshot database or writing a new `_snapshot` record:
  - Latest record `COMPLETED` but never checkpointed → a distinct
    "no checkpoint" rejection status.
  - Any other non-terminal/unknown state (or an undecodable record) → an
    "in progress" rejection status.
- A same-height resubmission of an existing snapshot transaction must keep
  going through the existing tx-status deduplication path unchanged — it
  should not be treated as a new snapshot request.

Requires two new `committerpb.Status` values (`REJECTED_SNAPSHOT_IN_PROGRESS`,
`REJECTED_SNAPSHOT_NO_CHECKPOINT`) from `fabric-x-common`.

## Open questions / notes

- A missing row for a recorded pointer, or a record that fails to decode, is
  a data-integrity problem, not a normal outcome — should surface as a hard
  error (batch fails/retries), not a silent rejection status.
- At most one snapshot lifecycle should be active system-wide at any time.

---

cendhu:
[vc] Reject snapshot request while prior snapshot not CHECKPOINTED

#### Type of change

- New feature
- Test update

#### Description

- Gate a new `_snapshot` request in `commitTransactions`, right after
  blind-write categorization and before `createSnapshotIfPresent`: look up
  the latest `_snapshot` record via a metadata pointer and accept only when
  it is `CHECKPOINTED` or absent.
- The pointer (metadata key `latest snapshot key`) is pre-seeded (NULL) at
  DB init alongside `last committed block number`, and updated with the
  existing `setMetadataPrepSQLStmt`/`getMetadataPrepSQLStmt` helpers. It is
  set in `database.commit`, right after `writeStatesByGroup` succeeds and
  before `tx.Commit`, so it lands in the same DB transaction as the
  `_snapshot` row it targets and is always consistent with it.
- On rejection, remove the incoming `_snapshot` write from `newWrites` and
  set `invalidTxStatus`, so the normal status path reports the rejection
  without creating a snapshot database or record:
  - `COMPLETED`-but-not-checkpointed -> `REJECTED_SNAPSHOT_NO_CHECKPOINT` (120)
  - any other non-CHECKPOINTED state -> `REJECTED_SNAPSHOT_IN_PROGRESS` (119)
- A pointer that names a missing row, or a row whose value fails to decode,
  is an invariant violation or storage corruption, not a normal rejection
  outcome -- both are returned as hard errors so the batch fails/retries
  instead of silently masking the anomaly.
- A resubmission (txID already in `tx_status`) bypasses the gate and keeps
  its committed status via existing duplicate-tx handling.

Bumps `fabric-x-common` to the pseudo-version carrying
`Status_REJECTED_SNAPSHOT_IN_PROGRESS` / `Status_REJECTED_SNAPSHOT_NO_CHECKPOINT`.
Tested with `make test-all-db` (841 tests) and `make lint`.

#### Related issues

- resolves #707

### DIFF
--- a/service/vc/committer.go
+++ b/service/vc/committer.go
@@ -101,14 +101,6 @@ func (c *transactionCommitter) commitTransactions(
 		return nil, err
 	}
 
-	// Gate a new snapshot request: reject it unless the latest _snapshot record
-	// is CHECKPOINTED. On rejection this removes the incoming _snapshot write and
-	// sets its invalid status, so createSnapshotIfPresent below no-ops and no
-	// snapshot database or record is created.
-	if err := db.rejectSnapshotIfPriorNotCheckpointed(ctx, vTx); err != nil {
-		return nil, fmt.Errorf("failed to gate snapshot before commit: %w", err)
-	}
-
 	// Ideally, we should add BlindWrite entries with a null version to the
 	// `readToTxIDs` map. This is because transactions can be resubmitted to multiple
 	// `vcservice` instances during connection issue between the coordinator and
--- a/service/vc/committer.go
+++ b/service/vc/committer.go
@@ -101,6 +101,14 @@ func (c *transactionCommitter) commitTransactions(
 		return nil, err
 	}
 
+	// Gate a new snapshot request: reject it unless the latest _snapshot record
+	// is CHECKPOINTED. On rejection this removes the incoming _snapshot write and
+	// sets its invalid status, so createSnapshotIfPresent below no-ops and no
+	// snapshot database or record is created.
+	if err := db.rejectSnapshotIfPriorNotCheckpointed(ctx, vTx); err != nil {
+		return nil, fmt.Errorf("failed to gate snapshot before commit: %w", err)
+	}
+
 	// Ideally, we should add BlindWrite entries with a null version to the
 	// `readToTxIDs` map. This is because transactions can be resubmitted to multiple
 	// `vcservice` instances during connection issue between the coordinator and
--- a/service/vc/database.go
+++ b/service/vc/database.go
@@ -33,6 +33,8 @@ const (
 	insertNsStatesSQLTempl = "SELECT * FROM insert_ns_${NAMESPACE_ID}($1::BYTEA[], $2::BYTEA[]);"
 	// queryVersionsSQLTempl template for the querying versions for given keys for each namespace.
 	queryVersionsSQLTempl = "SELECT key, version FROM ns_${NAMESPACE_ID} WHERE key = ANY($1);"
+	// queryValuesByKeysSQLTempl template for querying values for given keys for each namespace.
+	queryValuesByKeysSQLTempl = "SELECT key, value FROM ns_${NAMESPACE_ID} WHERE key = ANY($1);"
 
 	// insertTxStatusSQLStmt commits transaction's status for each TX.
 	insertTxStatusSQLStmt = "SELECT * FROM insert_tx_status($1::BYTEA[], $2::INTEGER[], $3::BYTEA[]);"
@@ -46,7 +48,17 @@ const (
 	queryTxIDsStatusPrepSQLStmt = "SELECT tx_id, status, height FROM tx_status WHERE tx_id = ANY($1);"
 )
 
-var lastCommittedBlockNumberKey = []byte("last committed block number")
+var (
+	lastCommittedBlockNumberKey = []byte("last committed block number")
+
+	// latestSnapshotKeyMetadataKey stores the tx_id of the most recently accepted
+	// _snapshot record, so the snapshot gate (see snapshot.go) can look up its
+	// current status with a single key lookup instead of scanning the (small but
+	// growing) ns__snapshot table. Pre-seeded (NULL) at DB init alongside
+	// lastCommittedBlockNumberKey and written atomically, in the same DB
+	// transaction as the _snapshot row it points to, by commit.
+	latestSnapshotKeyMetadataKey = []byte("latest snapshot key")
+)
 
 type (
 	// database handles the database operations.
@@ -219,6 +231,13 @@ func (d *database) commit(ctx context.Context, states *statesToBeCommitted) (*co
 		return res, nil
 	}
 
+	// If this batch carries a _snapshot write, record its key as the latest
+	// snapshot pointer in the SAME transaction, so the pointer and the _snapshot
+	// row it targets are always consistent (see rejectSnapshotIfPriorNotCheckpointed).
+	if err = setLatestSnapshotKeyIfPresent(ctx, tx, states.newWrites[committerpb.SnapshotNamespaceID]); err != nil {
+		return nil, fmt.Errorf("failed to set latest snapshot key: %w", err)
+	}
+
 	err = tx.Commit(ctx)
 	promutil.Observe(d.metrics.databaseTxBatchCommitLatencySeconds, time.Since(start))
 	if err != nil {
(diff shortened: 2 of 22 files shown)


service/vc/snapshot.go
```diff
@@ -40,6 +41,139 @@ import (
...
+// after this gate runs).
+func (db *database) rejectSnapshotIfPriorNotCheckpointed(
+	ctx context.Context, vTx *validatedTransactions,
+) error {
+	// A snapshot TX is submitted standalone (one transaction, one new-write entry).
+	if len(vTx.newWrites) != 1 {
+		return nil
+	}
+	var incomingTxID TxID
+	var found bool
```
liran-funaro: nit: `found` is redundant. `incomingTxID` not empty is sufficient check.

service/vc/snapshot.go
```diff
@@ -40,6 +41,139 @@ import (
...
+// incoming request should receive, or STATUS_UNSPECIFIED when the request may
+// proceed (no prior snapshot ever accepted, or the latest one is CHECKPOINTED).
+//
+// A pointer that names a missing row, or a row whose value fails to decode, is
+// an invariant violation (the pointer is written atomically with its row; see
+// setLatestSnapshotKeyIfPresent) or storage corruption, not a normal rejection
+// outcome. Both are returned as errors -- never silently mapped to a
+// conservative rejection status -- so the batch fails/retries instead of
+// masking the anomaly.
+func (db *database) latestSnapshotRejectionStatus(ctx context.Context) (committerpb.Status, error) {
```
liran-funaro: nit: the name is a bit confusing. It sounds like we're reading a rejection status from the state DB, while the method actually infers/derives it from the current state.
Also, `latestSnapshotRejectionStatus` suggests that the result is always a rejection status. When reading the caller above, I initially didn't understand why execution terminates at this point.

Maybe: `determineSnapshotStatus`

service/vc/snapshot.go
```diff
@@ -40,6 +41,139 @@ import (
...
+		// silently making the gate assume no active snapshot.
+		return committerpb.Status_STATUS_UNSPECIFIED,
+			errors.Newf("latest snapshot key %s has no matching _snapshot record", key)
+	}
+
+	var state committerpb.SnapshotState
+	if decodeErr := proto.Unmarshal(rows[0], &state); decodeErr != nil {
+		return committerpb.Status_STATUS_UNSPECIFIED,
+			errors.Wrapf(decodeErr, "failed to decode latest _snapshot record for key %s", key)
+	}
```
liran-funaro: nit: I suggest moving this into `readRawSnapshotRecords()`, making the method a complete cycle, rather than a step.

service/vc/committer.go
```diff
@@ -101,6 +101,14 @@ func (c *transactionCommitter) commitTransactions(
 		return nil, err
 	}
 
+	// Gate a new snapshot request: reject it unless the latest _snapshot record
+	// is CHECKPOINTED. On rejection this removes the incoming _snapshot write and
+	// sets its invalid status, so createSnapshotIfPresent below no-ops and no
+	// snapshot database or record is created.
+	if err := db.rejectSnapshotIfPriorNotCheckpointed(ctx, vTx); err != nil {
```
liran-funaro: Shouldn't this step happen in the preparer? We only reject TXs at this point for new keys - this is only due to a insert/fail optimization.

If I understand the flow correctly, we verify that a snapshot is valid (no other active one) before the db TX start.
Then, the TX commits a snapshot value (`setLatestSnapshotKeyIfPresent()`).
Correct?

If so, this exactly mirrors any other TX (preparer checks, then committer commits).

service/vc/database.go
```diff
@@ -274,6 +293,18 @@ func (d *database) writeStatesByGroup(
 	return nil, nil
 }
 
+// setLatestSnapshotKeyIfPresent points latestSnapshotKeyMetadataKey at w's sole
+// key, or no-ops when w is empty (normal non-snapshot batch). The key is
+// pre-seeded (NULL) at DB init, so the existing setMetadataPrepSQLStmt UPDATE
+// is reused as-is.
+func setLatestSnapshotKeyIfPresent(ctx context.Context, tx pgx.Tx, w *namespaceWrites) error {
```
liran-funaro, line 300: Why do we use the metadata here? Isn't it worth using the actual `_snapshot` table to store this?
Then, this can be a regular RW set added to the TX, without having to add special handling for this kind of TX.
Furthermore, I'm not sure this qualifies as metadata.
So far we used the metadata to store recovery data, not correctness.
This key is imperative to the correctness of the execution, thus require a fitting namespace.

service/vc/committer.go
cendhu: Initially, I was wondering whether this should happen at the sidecar but it required introduing gRPC APIs. I didn't prefer that as such occurrences are rare. Hence, I moved it to the lowest part as it accesses the database while prepare does not. We can also move it to the validator.

service/vc/database.go
cendhu: This entry will be used by the recovery mechanism too. It is more of an optimization that avoid scanning all rows in the _snapshot namespace. We will also introduce a latest checkpoint key in the metadata field. I agree that it is more of a pointer to optimize query. I am okay to move this to _snapshot table if you prefer that as I am okay with either.

service/vc/committer.go
liran-funaro: Yeah. I was just coming here to write that I got confused. This should be in the validator.

service/vc/database.go
liran-funaro: Maybe I don't understand correctly the flow. My understanding that this was not optimization but a requirement.
Can you clarify?

cendhu: For recovery and to reject snapshot request when prior is not checkpointed, we need to query the _snapshot table to check the status of the most recent snapshot request. This check can be done by counting the status or querying all rows, sort by block height, and pick the most recent or etc... To avoid that, we simply create a pointer to the latest snapshot request so that snapshot request can be easily rejected and coordinator can recover hash workers if vc fails during the hash computation.

liran-funaro: I see. Then this is fine to keep this in the metadata, as it does not add new information to the State DB, just index it.

cendhu:
@liran-funaro All comments have been addressed.

liran-funaro (approved):
LGTM

PR #710, opened 2026-07-28
[coordinator] Fix VC manager in-flight transaction tracking

### LINKED ISSUE #608: [coordinator] VC manager stream end handling
VC manager does not re-add jobs to `txBeingValidated` on failed submit to queue (e.g., when the stream ends).

liran-funaro:
#### Type of change

- Bug fix
- Test update

#### Description

- Return the transactions to `txBeingValidated` when forwarding their results fails, so that
  `recoverPendingTransactions` can re-queue them instead of dropping them. The signature verifier
  manager already guards this invariant, the VC manager did not.
- Move the map insertion into a new `addTxsBeingValidated`, mirroring the verifier manager, so the
  key is computed in one place.
- Key `validatorCommitter.txBeingValidated` by `servicepb.Height` rather than `TxRef.TxId`, again
  mirroring the verifier manager. The height is the transaction's own identity within the block, so
  matching a status back to its node no longer depends on transaction IDs being unique among the
  transactions currently in flight.

The failed-forward re-add is currently reachable only while the manager itself is tearing down,
because both output writes are bound to the manager's context rather than the stream's. It is
guarded here so the invariant holds regardless of how those contexts are wired later.

`TestValidatorCommitterAddAndRecoverPendingTxs` covers the add/recover round trip that the guard
relies on.

#### Related issues

- resolves #608

### DIFF
--- a/service/coordinator/validator_committer_manager.go
+++ b/service/coordinator/validator_committer_manager.go
@@ -52,9 +52,9 @@ type (
 		metrics   *perfMetrics
 		policyMgr *policyManager
 
-		// vc service returns only the txID and the status of the transaction. To find the
-		// transaction node associated with the txID, we use txBeingValidated map.
-		txBeingValidated utils.SyncMap[string, *dependencygraph.TransactionNode]
+		// txBeingValidated stores the transactions currently being validated by this vcservice, so the
+		// status returned by the vcservice can be matched back to its transaction node.
+		txBeingValidated utils.SyncMap[servicepb.Height, *dependencygraph.TransactionNode]
 	}
 
 	validatorCommitterManagerConfig struct {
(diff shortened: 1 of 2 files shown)


service/coordinator/validator_committer_manager.go
```diff
@@ -333,14 +333,21 @@ func (vc *validatorCommitter) receiveStatusAndForwardToOutput(
 		//       this is not an issue. If the sidecar becomes bottlenecked and cannot receive
 		//       the statuses quickly, the gRPC flow control will activate and slow down the
 		//       whole system, allowing the sidecar to catch up.
+		// NOTE: getTxsAndUpdatePolicies removes the transactions from txBeingValidated before their
+		//       results are queued, so a failed write must return them to the map. Otherwise
+		//       recoverPendingTransactions has nothing to re-queue and the transactions are lost:
+		//       their status never reaches the sidecar, and their nodes never free their dependents
+		//       in the dependency graph. The signature verifier manager guards the same invariant.
 		if ok := outputTxsStatus.write(ctx, txsStatus); !ok {
+			vc.addTxsBeingValidated(txsNode)
```
cendhu, line 342: I understood this change but not the change of map key from txID to txHeight. We will not receive duplicate txIDs across active txs. 

Noticed that verifier has height as map key instead of txID. Hence, this change make sense.

cendhu (approved):
LGTM

service/coordinator/validator_committer_manager.go
liran-funaro: I wanted to align the implementations. Either moving them both to TX ID or both to TX height was fine.

PR #713, opened 2026-07-28
[coordinator] Split the VC request/response API from the VC manager

liran-funaro:
#### Type of change

#### Description

- Move the load-balanced vcservice connection, and the request/response calls it serves
  (`SetLastCommittedBlockNumber`, `GetNextBlockNumberToCommit`, `GetTransactionsStatus`,
  `GetNamespacePolicies`, `GetConfigTransaction`), out of `validatorCommitterManager` and into a new
  `validatorCommitterAPI`. These calls are served by whichever vcservice the load-balanced
  connection picks and are issued outside the transaction pipeline, so the manager is now
  responsible only for streaming transactions to the vcservices.
- The coordinator owns the API and creates it in `Run` before starting the managers. The
  `channel.Ready` handshake that existed only to tell the coordinator when the manager had assigned
  its `commonClient` is therefore removed, along with the `WaitForReady` call in `Run`.
- `newVcMgrTestEnv` waits for the vcservice connections instead of that handshake, matching what
  `newSvMgrTestEnv` already does for the verifiers.

#### Related issues

- resolves partly #611

### DIFF
--- a/service/coordinator/coordinator.go
+++ b/service/coordinator/coordinator.go
@@ -41,6 +41,7 @@ type (
 		dependencyMgr         *dependencygraph.Manager
 		signatureVerifierMgr  *signatureVerifierManager
 		validatorCommitterMgr *validatorCommitterManager
+		validatorCommitterAPI *validatorCommitterAPI
 		policyMgr             *policyManager
 		queues                *channels
 		config                *Config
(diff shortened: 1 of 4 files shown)


cendhu (approved):
LGTM.

Nice change.

Issue #671, opened 2026-07-16
[loadgen] Loading the query and committer in parallel


liran-funaro:
### Motivation

Our load generator drives the commit path well, but two production-critical behaviors are effectively untested under load:

1. **The query service.** In real deployments, endorsing clients read current key versions from the query service before submitting transactions. The loadgen never issues these reads, so we have no way to load-test the query service alongside a live commit workload, or to measure how querying and committing interact when they share the same database.
2. **Reads that carry a real version — including under contention.** Today every generated read uses a nil version (valid only because the keys are freshly created). We can inject conflicts, but those transactions are engineered to *fail*. We have never load-tested the far more common real-world case: transactions that read an existing key at its committed version, contend with concurrent writers to that key, get *ordered* by the dependency graph, and **still commit**.

Previously this required a separate benchmarking tool and a cumbersome two-phase procedure — fill the database in one run, then re-run against the same generated keys to query and read them back — with no way to control the timing relationship between a key's creation and its later read.

### What we want to evaluate

A single load run that concurrently commits new transactions **and** reads back previously-created keys through the query service, with a configurable gap between when a key is created and when it is read back:

- **Query-service throughput and latency** while a full commit workload runs against the same database.
- **The read-validation / commit path for versioned reads**, at realistic scale.
- **Contention that commits.** With a small create-to-read gap, creates and read-modify-writes of the same key collide in-flight; the dependency graph must order them and most should still commit. Sweeping the gap from small to large should move the system from a high-contention regime (some aborts) to a clean regime (few/no aborts), letting us characterize the coordinator's ordering behavior and the query-service load in between.
- A mix of **hits and misses** — a configurable share of reads targeting keys that don't exist — to stress the query service's not-found path and produce nil-version reads.

### Success criteria

- A single loadgen configuration produces both commit traffic and query traffic; no second run or manual key bookkeeping.
- Query load (read counts, hit/miss mix) and the create-to-read gap are configurable.
- Loadgen metrics report query volume, latency, and hit/miss counts alongside the existing commit metrics, so a gap sweep can be plotted against query load and abort rate.

PR #718, opened 2026-07-30
[loadgen] Index-addressable workload with commit-time contention

### LINKED ISSUE #678: [loadgen] Index-addressable (random-access) key regeneration
### Motivation

The load generator derives its keys from a sequential pseudo-random stream: each worker pulls key bytes from a seeded PRNG in generation order. A key is therefore defined by its *position* in the stream, and the only way to obtain the key at position *N* is to consume the stream up to *N*. There is no way to regenerate a specific previously-generated key — or a past transaction — directly.

This blocks any workload that needs to reference keys created earlier in the run — most importantly, reading or querying keys the loadgen already committed (see #671). Today such scenarios require either holding large key sets in memory, or a two-phase procedure (fill in one run, then replay in a second run with the same seed) with no control over how far back a referenced key was created. It also makes the generated key bytes sensitive to the exact order in which the generator consumes its seed, which is fragile as the generator evolves.

### What we want

Make each generated key a pure function of its **address** — `(worker, transaction-index, slot)` — rather than its position in a stream, so that:

- **Any transaction is regenerable by its index** — its full content (layout / set sizes, keys, values, and id) is recomputable directly from the index, without replaying the transactions before it.
- **Any key is regenerable by its address in O(1)** — no replay, no in-memory key store, no second run.
- A workload can reference previously-created keys at a **configurable lag** within a single run, by regenerating a past transaction *t = j − L* by its index and recalling its keys.
- The per-transaction **layout** stays a deterministic function of the transaction index, kept **separate from key bytes**: the layout decides *how many* keys a transaction has; the key bytes are a pure function of the address.
- The determinism contract is stronger and simpler: the same seed (with worker count and key size) reproduces the same key for a given address, independent of how the generator consumes its seed.

The ECDSA **signature** and the envelope **timestamp** remain non-deterministic and are outside the contract — they are not part of a transaction's identity or content.

### Relationship to #671

This is the enabler for #671 (loading the query and committer in parallel): querying previously-created keys during a live commit load means, for a live transaction *j*, regenerating a past transaction *t = j − L* by its index and recalling its keys — an O(1) operation with no buffering and no two-phase run. The current sequential stream cannot do this. It also stands on its own as a general determinism/reproducibility improvement for the load generator.

### Success criteria

- Keys are a deterministic function of their `(worker, transaction-index, slot)` address, reproducible from seed + worker count + key size (the contract is preserved and strengthened, though key bytes are not identical to the current implementation).
- Any transaction (layout, keys, values, id) and any individual key are regenerable by index/address in O(1), without consuming or replaying prior state.
- The per-transaction layout (set sizes) is a deterministic function of the transaction index, decoupled from key identity.
- The existing statistical / distribution properties of the generated workload are preserved.

liran-funaro:
#### Type of change

- New feature

#### Description

- Make every transaction a pure function of a single global index: keys, values, nonce, and metadata are derived from a domain-separated PRF, so any transaction is regenerable in O(1) with no replay buffer or key store.
- Feed all workers from one shared atomic transaction-index counter, so the generated multiset is worker-count invariant (workers are pure parallelism) and reproducibility needs only the seed.
- Add a new-vs-existing key split (`new-keys-rate`, `tx-reference-gap`, `key-lookback-window`): slots either create fresh keys or reference keys from earlier transactions, producing commit-time contention deterministically, with no key map and no second run.
- Consolidate the transaction knobs into `TransactionProfile` and drop `KeyProfile`/`ConflictProfile`, the Modifier mechanism, the per-worker seeders, and `ByteArrayGenerator`.
- Rewrite `BenchmarkDependencyGraph` to exercise the backward dependency shapes (no-dep / RAW / WAW / RAW+WAW) via the split.

Reads carry nil versions for now; a later PR adds a querier that fills the real committed versions (the query seam is nil throughout). The reference knobs name their units — `tx-reference-gap` in transactions (previously key indices), `key-lookback-window` in keys; since this is unreleased, no existing config depends on them.

The key-generation metrics (`created_keys_total`, `referenced_read_keys_total`, `referenced_write_keys_total`) are split into a follow-up PR to keep this change focused on the workload model.

#### Related issues

- resolves #678
- related to #671


service/coordinator/dependencygraph/manager_test.go
```diff
@@ -39,121 +45,133 @@ func BenchmarkDependencyGraph(b *testing.B) {
...
+	// noSplit leaves new-keys-rate unset: every slot is a fresh unique key, so there are no dependencies.
+	noSplit := func(ro, rw, bw uint32) func(*workload.Profile) {
+		return func(p *workload.Profile) {
+			p.Transaction.ReadOnlyCount = ro
+			p.Transaction.ReadWriteCount = rw
+			p.Transaction.BlindWriteCount = bw
 		}
 	}
+	// withSplit enables the split: references are drawn from a 64-key working set one transaction behind.
+	withSplit := func(newKeysRate float64, ro, rw, bw uint32) func(*workload.Profile) {
```
cendhu: nit. the newKeysRate is always 1. Why not set the value 1 in line 72 instead of accepting it as an argument.

cmd/config/samples/loadgen.yaml
```diff
@@ -141,32 +141,62 @@ load-profile:
...
+    # unique key). A value of 0 means NO new keys at all (every slot references the static pre-genesis
+    # working set). A positive value must not exceed the TOTAL slot count (read-only + read-write-count +
+    # write-count). To keep read-writes doing checked inserts rather than forcing blind-write overflow,
+    # keep new-keys-rate <= read-write-count.
+    # new-keys-rate: 1
+    # Minimum transaction distance behind the current transaction from which existing references are
+    # drawn: at transaction j the reference window ends at the frontier as of j-gap. 0 (default)
+    # references the newest keys (in-flight, highest contention); larger values reference older, already-
+    # committed keys. Expressed in TRANSACTIONS (a distance), not keys. Ignored when new-keys-rate unset.
+    # reference-gap: 0
```
cendhu: Sorry, i gave up on this. Earlier gap, src, dest were easy to grasp though it did not cover all scenarios. I am not sure whether such a complexity is required or a simple alternative is possible. Can you add some example or maybe a doc to explain new keys rate, reference-gap, and lookback-window with an example? I am sure it will be helpful for anyone who uses the loadgen and configure these values.

loadgen/metrics/metrics.go
```diff
@@ -100,6 +112,25 @@ func NewLoadgenServiceMetrics(c *Config) *PerfMetrics {
...
+		referencedReadKeysTotal: p.NewCounterFunc(prometheus.CounterOpts{
+			Namespace: "loadgen",
+			Name:      "referenced_read_keys_total",
+			Help:      "Total number of existing (backward) read-only key references generated",
+		}, func() float64 { return float64(counter.KeyStats().ReferencedReadKeys) }),
+		referencedWriteKeysTotal: p.NewCounterFunc(prometheus.CounterOpts{
+			Namespace: "loadgen",
+			Name:      "referenced_write_keys_total",
+			Help:      "Total number of existing write-slot key references (read-write + blind-write) generated",
+		}, func() float64 { return float64(counter.KeyStats().ReferencedWriteKeys) }),
```
cendhu: This can be part of a separate PR too. Suddenly changing the focus area is hard for me.

loadgen/workload/config.go
```diff
@@ -81,22 +82,77 @@ type TransactionProfile struct {
...
+	// contention); larger values reference older, already-committed keys. Expressed in TRANSACTIONS (a
+	// distance), not keys. Ignored when new-keys-rate is unset.
+	ReferenceGap uint64 `mapstructure:"reference-gap" yaml:"reference-gap"`
+	// LookbackWindow is the size, IN KEYS, of the moving working set an existing reference is drawn from:
+	// a reference lands in [top-W, top), spread across the window by a per-transaction random offset so
+	// references are not all pinned to a single key. The window slides forward as the frontier grows, so
+	// a larger window spreads references over more keys (less contention) and a smaller one concentrates
+	// them (more contention). Required to be at least the total slot count (read-only + read-write +
+	// write) when new-keys-rate is set, so all references within one transaction stay distinct. Expressed
+	// in KEYS (a working-set size), not transactions. Ignored when new-keys-rate is unset.
```
cendhu: what is the problem with previous dependency creation approach with gap, src, and dest?

loadgen/workload/stream.go
```diff
@@ -49,22 +74,14 @@ func (s *TxStream) Run(ctx context.Context) error {
 	g, gCtx := errgroup.WithContext(ctx)
 	for _, gen := range s.gens {
 		g.Go(func() error {
-			return s.generateBatches(gCtx, gen)
+			// The query stage is wired per worker (its own connection) once a query client exists;
+			// until then it is nil and reads keep their nil versions.
+			return s.generateBatches(gCtx, gen, nil)
```
cendhu, line 79: we are always passing nil. Why do we need the batchQuerier then? If this is going to be used in the subsequent PR, can we introduce the batchQuerier later?

loadgen/workload/stream.go
```diff
@@ -10,29 +10,54 @@ import (
...
+	// the query decision to the tx index. Instead queries-rate is an AVERAGE number of reads versioned per
+	// TX that a concrete querier realizes with its own rate mechanism (e.g. a stateful accumulator), NOT a
+	// cumulative floor over the tx index. Within each TX it versions the highest-priority reads first by
+	// iterating the read set in REVERSE: reads are laid out new-keys-first, so back-references (existing,
+	// committed keys) come before newly-introduced ones.
+	batchQuerier interface {
+		// FillVersions fills in the committed version of the reads (ReadsOnly and ReadWrites) it selects
+		// across the whole batch of unsigned transactions, in place, before the batch is signed.
+		FillVersions(ctx context.Context, batch []*applicationpb.Tx) error
+	}
```
cendhu, line 51: It is used only in the test. Based on the comment, i understand this method would query version from database and use it in the read-only and read-write set. Is that correct?

loadgen/workload/stream_test.go
```diff
@@ -214,6 +224,79 @@ func TestGenValidTx(t *testing.T) {
...
+
+	counter := NewTxCounter(p.Transaction)
+	s, err := NewTxStream(p, defaultStreamOptions(), counter)
+	require.NoError(t, err)
+	const n = 100
+	s.gens[0].buildBatch(n)
+	w := uint64(p.Transaction.ReadWriteCount + p.Transaction.BlindWriteCount)
+	slotsPerTx := w + uint64(p.Transaction.ReadOnlyCount)
+	require.Equal(t, KeyStats{KeyFrontier: n * slotsPerTx}, counter.KeyStats())
+}
```
cendhu: some level of duplication. If a table test is not possible, I am fine with this.

loadgen/workload/tx_rand.go
```diff
@@ -0,0 +1,246 @@
+/*
```
cendhu, line 1: This has the core changes. I need some meta comment or doc to understand 
1. how keys are generated and used across reads-only, reads-write, and blind writes.
2. how versions are added
3. how do we allocate keys per tx
4. how do we reuse some old keys to generate conflicts on readonly to other, read-write to others and so on.

liran-funaro: TBH, I think this mechanism is much simpler from the previous one. I think the issue is shifting the mental model between old and new. But for a new user, this is much simpler.


>how keys are generated and used across reads-only, reads-write, and blind writes.

TX layout is static (same count fields as before).

>how do we allocate keys per tx

new-keys-rate says how many new-keys are generated per TX. otherwise, it will reference previous keys.
new keys populate first read-write, then blind-write, then read-only.

>how versions are added

next PR will introduce querier that will fill the versions. for now, nil versions are used.

 >how do we reuse some old keys to generate conflicts on readonly to other, read-write to others and so on.

Use a lower new-keys-rate than the sum of keys slots (reads-only-count + reads-write-count + blind-writes-count).
If you use a low reference-gap, the conflicting TXs may land in the same block which will create a live dependency.

cmd/config/samples/loadgen.yaml
liran-funaro: I updated the description. Please let me know if it is simple enough now.

loadgen/workload/config.go
liran-funaro: The previous method changes the layout decided by the user config by adding keys.
The new method is using a fixed layout, and changing which key is new or reference.
IMHO, it is more explicit.
It explicitly say how many new keys are created - explicitly deciding the grow rate.
A later knob will added to control exactly how many queries will be made per TX, which will decide the load the query service.

loadgen/workload/stream.go
liran-funaro: Correct

loadgen/metrics/metrics.go
liran-funaro: I moved the metrics to a new PR

cmd/config/samples/loadgen.yaml
```diff
@@ -141,32 +141,44 @@ load-profile:
...
+    # reference-gap, and lookback-window control only which keys fill those slots — fresh keys, or
+    # references to keys from earlier transactions — which is what creates (or avoids) commit-time
+    # contention. Reads carry nil versions for now; a later PR adds a querier that fills the real versions.
+    #
+    # new-keys-rate: average number of new keys generated per transaction; the remaining slots reference
+    # keys from earlier transactions, so keys are reused and the coordinator sees commit-time contention.
+    # New keys fill slots in layout order: read-write, then blind-write, then read-only. UNSET (default)
+    # gives every slot a fresh unique key — no reuse, no contention. To create conflicts, set it below the
+    # total slot count (read-only + read-write + write); it must not exceed that count. 0 means no new keys
+    # at all (every slot references a fixed static set).
```
cendhu: If the slot size is 4 (ro: 2, rw: 1, bw: 1) and new key rate is 2, then rw and bw have new keys while ro refers to some old keys?

cmd/config/samples/loadgen.yaml
```diff
@@ -141,32 +141,44 @@ load-profile:
...
+    # new-keys-rate: 1
+    # reference-gap: how far back, in transactions, references reach — a reference targets a key at least
+    # reference-gap transactions old. 0 (default) draws the newest keys, which may still be in flight, so
+    # conflicting transactions can land in the same block (a live dependency); larger values draw older,
+    # already-committed keys. Ignored when new-keys-rate is unset.
+    # reference-gap: 0
```
cendhu: reference gap is used to decide which old keys to use? In the following example, the slot size is 4 (ro: 2, rw: 1, bw: 1) and new key rate is 2, then rw and bw have new keys while ro refers to some old keys, how various reference-gap impacts the ro key set? 0, 10, 1000?

cmd/config/samples/loadgen.yaml
```diff
@@ -141,32 +141,44 @@ load-profile:
...
+    # lookback-window: how many recent keys a reference is drawn from. Larger spreads references over more
+    # keys (less contention); smaller concentrates them (more contention). Must be at least the total slot
+    # count so a transaction's references stay distinct. Ignored when new-keys-rate is unset.
+    # lookback-window: 1024
```
cendhu: lookback-window and reference gap are confusing. How are they related or different?

cmd/config/samples/loadgen.yaml
```diff
@@ -141,32 +141,44 @@ load-profile:
...
-    # dependencies:
-    #   - probability: 0.05   # 5% of transactions will have this dependency
-    #     gap: 1              # Dependent transaction is the immediately next one
-    #     src: read-write     # Source transaction reads and writes a key
-    #     dst: read-write     # Destination transaction reads and writes the same key
+    #
+    # The transaction layout above is fixed (same slot counts every transaction). new-keys-rate,
+    # reference-gap, and lookback-window control only which keys fill those slots — fresh keys, or
+    # references to keys from earlier transactions — which is what creates (or avoids) commit-time
+    # contention. Reads carry nil versions for now; a later PR adds a querier that fills the real versions.
```
cendhu: Are we planning to generate load for query service via this fill version utility? Or will we have a separate load for query service along with fill version?

cmd/config/samples/loadgen.yaml
liran-funaro: Correct

cmd/config/samples/loadgen.yaml
liran-funaro: `reference-gap` (TX granularity) and `lookback-window` (keys granularity) work together.
If we are in TD index 1000, and `reference-gap` is 100, and `lookback-window` is 10. Then `ro` will sample (deterministically) from keys generated starting from TX index 900 down to 10 keys before this TX.

cmd/config/samples/loadgen.yaml
liran-funaro: Explained above.

cmd/config/samples/loadgen.yaml
liran-funaro: The next PR will call the query service on each batch to fill the versions according to a `query-rate` parameter.

cmd/config/samples/loadgen.yaml
cendhu: Please add this example in the comment.

cmd/config/samples/loadgen.yaml
cendhu: can we rename reference-gap to tx-reference-gap and lookback-window to key-lookback-window or something that reflects the key lookup widow at tx-reference?

cmd/config/samples/loadgen.yaml
liran-funaro: done. Also modified the fields names

loadgen/workload/config.go
```diff
@@ -81,22 +82,53 @@ type TransactionProfile struct {
...
+	// ReferenceGap is how far back, in transactions, references reach: a reference targets a key at least
+	// ReferenceGap transactions old. 0 (default) draws the newest keys, which may still be in flight — so
+	// conflicting transactions can land in the same block (a live dependency); larger values draw older,
+	// already-committed keys. Ignored when new-keys-rate is unset.
+	ReferenceGap uint64 `mapstructure:"reference-gap" yaml:"reference-gap"`
+	// LookbackWindow is how many recent keys a reference is drawn from. A larger window spreads references
+	// over more keys (less contention); a smaller one concentrates them (more contention). Must be at
+	// least the total slot count so a transaction's references stay distinct. Ignored when new-keys-rate
+	// is unset.
+	LookbackWindow uint64 `mapstructure:"lookback-window" yaml:"lookback-window"`
```
cendhu: no defaults as they are enabled only when new keys rate is set?

loadgen/workload/tx_counter.go
```diff
@@ -0,0 +1,30 @@
...
+
+import "sync/atomic"
+
+// TxCounter is the shared global transaction-index counter. Every worker generator reserves its
+// consecutive range of indices from this one counter (a single atomic increment per batch); sharing the
+// counter is what makes the generated multiset worker-count invariant (workers are pure parallelism). It
+// is created before the stream and shared with it; it depends only on the TransactionProfile, not on the
+// policy or the crypto artifacts.
+type TxCounter struct {
+	profile TransactionProfile
```
cendhu: where do we use this profile? I do not see it used anywhere.

loadgen/workload/tx_rand.go
```diff
@@ -0,0 +1,246 @@
...
+//     NONEXISTENT READ — it never creates, so the frontier index it consumes is wasted (never becomes a
+//     committed key). Only writes advance real committed state, but the frontier counts every introduced
+//     index, write or read.
+//   - every remaining slot BACKWARD-references an existing key drawn from the moving working set
+//     [top-W, top), where top = C(max(0, txIdx-gap)) is the frontier `gap` transactions behind and
+//     W = LookbackWindow. The offsets are a UNIFORM SAMPLE WITHOUT REPLACEMENT of [0, W) drawn by Floyd's
+//     algorithm (seeded from the tx index), so a transaction's slots reference a distinct, scattered
+//     subset of the window rather than a single contiguous run. Floyd's fixes the subset uniformly but
+//     does not shuffle it, so which drawn key lands on which slot is not itself uniform — immaterial
+//     here, where contention depends only on which keys are touched and that they are distinct.
```
cendhu: using formulas here does not make understanding the code easier. We should try to explain the logic in simple english as code already use formulas.

cendhu (commented):
The change seems fine to me. 

I still have a dilemma that whether this complexity is needed or some of the testing can be moved to coordinator/graph only benchmark. Hence, I leave this decision to you.

loadgen/workload/tx_counter.go
liran-funaro: This was used by the metrics which was moved to a followup PR.
I remove this here and re add it when needed.

loadgen/workload/tx_rand.go
liran-funaro: done

loadgen/workload/config.go
liran-funaro: The defaults are zero by default, so no defaults. I added a comment on the default behaviour when unset.

loadgen/workload/stream_test.go
liran-funaro: table test was too complicated so I reverted to duplication.

cendhu (approved):

Issue #681, opened 2026-07-20
Document breaking changes


cendhu:

liran-funaro:
## Breaking changes: loadgen YAML

Scope: the loadgen config surface (`cmd/config/samples/loadgen.yaml`, and
`cmd/config/templates/loadgen_shared.yaml.tmpl` which is integration-test only), for the
changes merged since this issue was opened — #689, #688, #699.

> [!IMPORTANT]
> **Most of these fail silently.** Config decoding does not set mapstructure's
> `ErrorUnused` (`cmd/config/app_config.go:132`), so a key that was *renamed or moved* is
> simply ignored: the file still loads, and the field falls back to its zero value. Only
> the changes that alter a value's *type* (#688) fail loudly. Please grep your own YAMLs
> rather than relying on startup to tell you.

### #699 — `key` and `conflicts` folded into `transaction`

`load-profile.key` and `load-profile.conflicts` no longer exist; their fields moved under
`load-profile.transaction`:

| Old key | New key |
| --- | --- |
| `load-profile.key.size` | `load-profile.transaction.key-size` |
| `load-profile.conflicts.invalid-signatures` | `load-profile.transaction.invalid-signatures` |
| `load-profile.conflicts.dependencies` | `load-profile.transaction.dependencies` |

**Impact — silent.** Feeding a pre-#699 YAML to current `main` loads with no error and
yields `KeySize=0`, `InvalidSignatures=0`, `Dependencies=[]`. Consequences:

- `key-size: 0` makes `ByteArrayGenerator.Next()` return `nil`
  (`loadgen/workload/primitives.go:28`), so **every generated key is the empty key** — the
  whole run collapses onto a single key instead of the intended keyspace.
- Invalid-signature injection and dependency-conflict injection are **silently disabled**,
  so a run meant to exercise the verifier's reject path or MVCC conflicts exercises
  neither.

### #688 — distribution wrappers removed (plain scalars)

`Distribution` is gone; the `const:`/`uniform:`/`normal:`/`bernoulli:`/`discrete:` variants
are no longer accepted anywhere. Affected keys under `load-profile.transaction`:

| Old | New |
| --- | --- |
| `read-only-count: {const: N}` | `read-only-count: N` |
| `read-write-count: {const: N}` | `read-write-count: N` |
| `write-count: {const: N}` | `write-count: N` |
| `dependencies[].gap: {const: N}` | `dependencies[].gap: N` |

[...]

PR #724, opened 2026-08-02
[coordinator] Fix swapped manager metrics, report queues on scrape

### LINKED ISSUE #720: [coordinator] Verifier and vcservice retry counters are exported under swapped subsystems
The two per-manager retry counters in `service/coordinator/metrics.go` are registered with each
other's `Subsystem`:

| field | incremented by | registered subsystem | exported series |
|---|---|---|---|
| `verifiersRetriedTransactionTotal` (`metrics.go:98`) | verifier manager | `vcservice` | `coordinator_vcservice_retired_transaction_total` |
| `vcservicesRetriedTransactionTotal` (`metrics.go:108`) | VC manager | `verifier` | `coordinator_verifier_retired_transaction_total` |

Every verifier retry is therefore exported as a vcservice retry, and vice versa.
`docs/metrics_reference.md` documents each series by its `Help` string, so the documented
meaning and the emitted series disagree for both rows.

The two tests that cover these counters — `requireRetriedTxsTotal` in
`signature_verifier_manager_test.go` and in `validator_committer_manager_test.go` — assert on the
Go field rather than on the exported name, so they pass while the exported series is wrong.

Three adjacent problems live in the same code:

* The metric name says `retired` where `retried` is meant; the `Help` string on both counters
  says "retried".
* The per-manager metric set (connection, processed total, retried total, queue sizes) is defined
  twice with only the subsystem differing. That duplication is how the subsystems got swapped,
  and it is the same drift described in #611.
* The four queue-size gauges are sampled by two dedicated goroutines on a
  `queue-monitor-sampling-time` ticker (`Service.monitorQueues`). Sampling a value that is already
  available on demand costs a goroutine per manager and makes every reading up to one interval
  stale.

### Proposal

1. Swap the subsystems back, so each counter is exported under the manager that increments it.
2. Rename `retired_transaction_total` to `retried_transaction_total`. This breaks existing
   dashboard queries, so it is worth doing together with the label fix — one break instead of two.
3. Define the metrics common to both managers once, parameterised by
   `monitoring.MetricsParameters`, and instantiate it per manager. Keep every existing metric.
   `coordinator_vcservice_output_tx_status_batch_queue_size` has no verifier counterpart, so it
   stays a coordinator-level metric — it must not be dropped, since it tracks the one queue whose
   saturation causes the idle-handshake deadlock documented at `coordinator.go:319`.
4. Report the queue depths with `prometheus.NewGaugeFunc` instead of a gauge that a goroutine
   sets on a timer. The value is then exact at scrape time, and the sampling goroutines,
   their tickers, and the sampling interval plumbed through the manager constructors all go away.
   This needs a `Provider.NewGaugeFunc` wrapper in `utils/monitoring`.
5. `scripts/metrics_doc_extract.py` needs two fixes for `make generate-metrics-doc` to keep
   working: follow a metric helper defined in the file being parsed (it currently resolves only
   `package.Func` against `utils/<package>/metrics.go`), and recognise `NewGaugeFunc` (the metric
   pattern allows only a `Vec` suffix today, so a gauge func is silently omitted from the table).
   A definition whose namespace is still `params.Namespace` must be skipped, since it is emitted
   once per call site with the namespace substituted.
6. Assert on the exported metric name per manager, so a future subsystem swap fails a test
   instead of shipping silently.

liran-funaro:
#### Type of change

- Bug fix
- Breaking change
- Test update

#### Description

- Register the metric set common to both service managers from a single `newManagerMetrics`,
  instantiated once per manager, so the verifier and vcservice series can no longer be given
  each other's subsystem.
- Rename `retired_transaction_total` to `transaction_retried_total` and shorten the queue-size
  gauge names. Dashboards querying the old names need updating; see the mapping below.
- Report the queue depths with `prometheus.NewGaugeFunc`, through a new
  `monitoring.Provider.NewGaugeFunc`, and drop `Service.monitorQueues` along with its goroutine
  and ticker. The reported value is now exact at scrape time instead of up to one sampling
  interval stale. `queue-monitor-sampling-time` is unchanged and still used by the dependency
  graph, which keeps its own sampler.
- `scripts/metrics_doc_extract.py`: follow a metric helper defined in the file being parsed (it
  resolved only `package.Func` against `utils/<package>/metrics.go`), recognise `NewGaugeFunc`,
  and skip a definition whose namespace is still `params.Namespace` — it is emitted once per call
  site with the namespace substituted.
- `TestManagerMetricsSubsystem` asserts on the exported series names. The existing tests asserted
  on the `perfMetrics` fields, which is why they passed while the series were swapped.

Metric name changes, for dashboard migration. The first two rows also change meaning, since the
counters were reporting each other's manager:

| before | after |
|---|---|
| `coordinator_vcservice_retired_transaction_total` (held verifier retries) | `coordinator_verifier_transaction_retried_total` |
| `coordinator_verifier_retired_transaction_total` (held vcservice retries) | `coordinator_vcservice_transaction_retried_total` |
| `coordinator_verifier_input_tx_batch_queue_size` | `coordinator_verifier_input_batch_queue_size` |
| `coordinator_verifier_output_validated_tx_batch_queue_size` | `coordinator_verifier_output_batch_queue_size` |
| `coordinator_vcservice_output_validated_tx_batch_queue_size` | `coordinator_vcservice_output_batch_queue_size` |

`coordinator_vcservice_input_batch_queue_size` is new. It observes the same channel as
`coordinator_verifier_output_batch_queue_size`, which is the queue between the two managers.

`coordinator_vcservice_output_tx_status_batch_queue_size` keeps its name. It has no verifier
counterpart, so it stays a coordinator-level metric rather than part of the shared set.

#### Related issues

- resolves #720

### DIFF
--- a/utils/monitoring/provider.go
+++ b/utils/monitoring/provider.go
@@ -86,6 +86,16 @@ func (p *Provider) NewGauge(opts prometheus.GaugeOpts) prometheus.Gauge {
 	return g
 }
 
+// NewGaugeFunc creates a new prometheus gauge whose value is reported by calling fn on every
+// scrape. Use it instead of a NewGauge that a dedicated goroutine samples on a timer: the value
+// is exact rather than as fresh as the sampling interval, and no goroutine is needed. fn is
+// called by the scrape handler, so it must be cheap and safe for concurrent use.
+func (p *Provider) NewGaugeFunc(opts prometheus.GaugeOpts, fn func() float64) prometheus.GaugeFunc {
+	g := prometheus.NewGaugeFunc(opts, fn)
+	p.registry.MustRegister(g)
+	return g
+}
+
 // NewGaugeVec creates a new prometheus gauge vector.
 func (p *Provider) NewGaugeVec(opts prometheus.GaugeOpts, labels []string) *prometheus.GaugeVec {
 	gv := prometheus.NewGaugeVec(opts, labels)
--- a/docs/metrics_reference.md
+++ b/docs/metrics_reference.md
@@ -52,22 +52,23 @@ The following Sidecar metrics are exported for consumption by Prometheus.
 
 The following Coordinator metrics are exported for consumption by Prometheus.
 
-| Name                                                       | Type    | Labels      | Description                                                                                                |
-|------------------------------------------------------------|---------|-------------|------------------------------------------------------------------------------------------------------------|
-| coordinator_grpc_received_transaction_total                | counter |             | Total number of transactions received by the coordinator service from the client.                          |
-| coordinator_grpc_committed_transaction_total               | counter | status      | Total number of transactions committed status sent by the coordinator service to the client.               |
-| coordinator_verifier_input_tx_batch_queue_size             | gauge   |             | Size of the input transaction batch queue of the signature verifier manager.                               |
-| coordinator_verifier_output_validated_tx_batch_queue_size  | gauge   |             | Size of the output validated transaction batch queue of the signature verifier manager.                    |
-| coordinator_vcservice_output_tx_status_batch_queue_size    | gauge   |             | Size of the output transaction status batch queue of the validation and committer service manager.         |
-| coordinator_vcservice_output_validated_tx_batch_queue_size | gauge   |             | Size of the output validated transaction batch queue of the validation and committer service manager.      |
-| coordinator_verifier_transaction_processed_total           | counter |             | Total number of transactions processed by the signature verifier manager.                                  |
-| coordinator_vcservice_transaction_processed_total          | counter |             | Total number of transactions processed by the validation and committer service manager.                    |
-| coordinator_verifier_connection_status                     | gauge   | grpc_target | Connection status to verifier service by grpc target (1 = connected, 0 = disconnected).                    |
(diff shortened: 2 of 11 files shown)


utils/monitoring/provider.go
```diff
@@ -86,6 +86,16 @@ func (p *Provider) NewGauge(opts prometheus.GaugeOpts) prometheus.Gauge {
 	return g
 }
 
+// NewGaugeFunc creates a new prometheus gauge whose value is reported by calling fn on every
+// scrape. Use it instead of a NewGauge that a dedicated goroutine samples on a timer: the value
+// is exact rather than as fresh as the sampling interval, and no goroutine is needed. fn is
+// called by the scrape handler, so it must be cheap and safe for concurrent use.
+func (p *Provider) NewGaugeFunc(opts prometheus.GaugeOpts, fn func() float64) prometheus.GaugeFunc {
+	g := prometheus.NewGaugeFunc(opts, fn)
```
cendhu, line 94: good find.

cendhu (approved):

cendhu:
https://github.com/hyperledger/fabric-x-committer/issues/681 please update this issue too

PR #725, opened 2026-08-02
[coordinator] Align VC and verifier manager stream behavior

### LINKED ISSUE #721: [coordinator] Align VC and verifier manager stream and forward behavior
`signatureVerifierManager` and `validatorCommitterManager` run the same stream loop, but the two
copies have drifted apart in three places. Each is a small fix on its own; together they are the
entire semantic delta that any extraction of the shared loop (#611) has to preserve, so they are
worth settling in place first.

### 1. `InvalidArgument` is retried forever on the VC path

The verifier's receive loop maps `codes.InvalidArgument` to `retry.ErrNonRetryable`
(`signature_verifier_manager.go:281-285`), on the grounds that it indicates corrupted or
maliciously altered state, or a bug in the committer — none of which a retry can fix. The VC's
receive loop has no such check, so the identical condition from a vcservice is retried under the
sustain policy indefinitely. Same cause, same correct response; only one manager implements it.

### 2. The VC sends empty batches on the wire

`sendTransactionsToVCService` has no empty-batch guard, so an empty node batch is marshalled and
sent as an empty `VcBatch` (`validator_committer_manager.go:192`).

Empty batches do reach it. The verifier's receive loop forwards `validatedTxs` unconditionally
(`signature_verifier_manager.go:294`), and that slice is empty when every status in a received
batch is untracked — which happens after a verifier restart produces duplicate responses. The
VC's own dep-graph write has the guard the verifier lacks
(`validator_committer_manager.go:281`). Both managers should skip empty batches, in both
directions.

### 3. A partially forwarded VC batch can be re-queued after its statuses were delivered

`receiveStatusAndForwardToOutput` writes the statuses first and returns early if that write fails
(`validator_committer_manager.go:273-277`), so the nodes are never forwarded for a batch whose
statuses were dropped. The reverse order of failure is not guarded, and cannot be by ordering
alone: if the status write succeeds and the node write then fails
(`validator_committer_manager.go:281-284`), the batch is returned to `txBeingValidated` and
re-queued by `recoverPendingTransactions`. The vcservice then re-commits (idempotently) and
re-emits the same statuses, so `numTxsInProgress` is decremented twice for one transaction.

That counter drives `Service.NoPendingTransactionProcessing` (`coordinator.go:319`), whose
documented invariant is `numTxsInProgress >= readyCount >= 0`. An over-decrement makes it
negative, the idle equality can never hold again, and the sidecar can never re-establish its
stream.

What keeps this unreachable today is that both output writes are bound to the manager context
rather than the stream context, so the failure branch is taken only while the manager is tearing
down (noted in #710). `validator_committer_manager.go:155` records that requirement but not its
consequence, so a refactor can rebind the contexts without seeing what it breaks — and the symptom
would surface as a sidecar that never reconnects, far from the change. State the consequence at
that NOTE.

### Relation to #611

All three are properties of the one stream loop that both managers implement separately, so
extracting that loop fixes them in a single place rather than twice. If the extraction lands
first, it resolves this issue; if not, each item above is a small in-place fix to whichever
manager lacks it.

liran-funaro:
#### Type of change

- Bug fix
- Test update

#### Description

- Bind both managers to one receive-error classification, `classifyStreamRecvError`. The VC
  manager previously retried an `InvalidArgument` under the sustain policy forever; the verifier
  manager already treated it as non-retryable.
- Skip empty batches in the VC send loop instead of marshalling and sending an empty `VcBatch`.
- Do not forward an empty batch from the verifier's receive loop to the VC manager, which is
  where those empty batches came from. The VC manager already guards its own output this way.
- Record at the VC's receive goroutine what breaks if the two output writes are bound to the
  stream context rather than the manager's: the re-queue becomes reachable on every stream
  failure, the statuses of a re-queued batch are delivered twice, and
  `Service.numTxsInProgress` is decremented twice for one transaction. The existing NOTE stated
  the requirement but not its consequence, which is what makes it easy to undo.

`TestValidatorCommitterManagerX/an_empty_batch_is_not_sent_to_the_vcservice` sends a real batch
first, because the first batch of a stream goes through `splitAndSendToVC`, which already sends
nothing for an empty batch — the empty send is only reachable once that path is behind us. With
the guard removed the test observes three batches on the wire instead of two.

No behavior changed for the failed-forward re-queue itself. The status write is already attempted
first and returns early on failure, so the nodes are never forwarded for a batch whose statuses
were dropped; the reverse ordering cannot be fixed by ordering and is instead held unreachable by
the context wiring, which is what the NOTE now explains.

This touches the same receive and send loops as #724, so whichever merges second needs a trivial
rebase.

#### Related issues

- resolves #721

### DIFF
--- a/service/coordinator/signature_verifier_manager.go
+++ b/service/coordinator/signature_verifier_manager.go
@@ -278,20 +278,13 @@ func (sv *signatureVerifier) receiveStatusAndForwardToOutput(
 	for {
 		response, err := stream.Recv()
 		if err != nil {
-			if grpcerror.HasCode(err, codes.InvalidArgument) {
-				// While it is unlikely that svm would send an invalid policy, it could happen
-				// if the stored policy in the database is corrupted or maliciously altered, or
-				// if there is a bug in the committer that modifies the policy bytes.
-				return errors.Join(retry.ErrNonRetryable, err)
-			}
-			// The stream ended or the SVM was closed.
-			return errors.Wrap(err, "receive from stream ended with error")
+			return classifyStreamRecvError(err)
 		}
 
 		logger.Debugf("New batch came from sv to sv manager, contains %d items", len(response.Status))
 
 		validatedTxs := sv.fetchAndDeleteTxBeingValidated(response)
-		if !outputValidatedTxs.Write(validatedTxs) {
+		if len(validatedTxs) > 0 && !outputValidatedTxs.Write(validatedTxs) {
 			// Since transactions are loaded and deleted from txBeingValidated before their
 			// validation results are queued, we must re-queue the transaction to txBeingValidated
 			// if its result cannot be added to the outputValidatedTxs queue.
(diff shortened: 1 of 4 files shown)


cendhu (approved):
LGTM. Nice set of fixes.

Issue #730, opened 2026-08-03
[coordinator, sidecar, vc, verifier] Report every queue size on scrape


liran-funaro:
#724 replaced the coordinator managers' queue gauges with `prometheus.GaugeFunc`, so the channel
length is read at scrape time instead of being sampled by a goroutine on a ticker
(`service/coordinator/metrics.go:108-123` is the reference shape).

This is the umbrella for finishing that across the system. Three ticker-based queue monitors remain,
several pipeline queues have no gauge at all, and one package's metrics are missing from the
reference doc entirely. The work is split into the sub-issues below, in the order they should land.

### Sub-issues

| | Scope |
| --- | --- |
| #732 | Scan every metrics file, so the dependency graph's 16 metrics reach `docs/metrics_reference.md` at all. |
| #733 | **sidecar** — the dead sampler, the six notifier and relay queues with no gauge, and the three queues rebuilt per coordinator session. |
| #770 | **vc** — the 250ms sampler, and `vcservice_batcher_input_queue_size` for the first queue the service fills from the stream. |
| #771 | **coordinator** — the dependency graph sampler, and removing the `queue-monitor-sampling-time` config key. |
| #811 | **coordinator** — the managers' pending queues and the two in-flight counters, which nothing reports today. |
| #734 | Report the two genuinely per-stream queues as one series per stream, labelled, so an operator can `sum` or `max` across workers. |

Originally #733 covered the sidecar, `vc` and the coordinator together. It was split per service on
review, since one PR spanning three services is slow to review and the three share nothing but the
helper.

Order: #732 first, so the doc covers what changes. #733 next, because it introduces the
`monitoring.NewChannelLenGauge` and `NewAtomicChannelLenGauge` helpers that #770 and #771 reuse;
those two are independent of each other afterwards. #734 last. Every one of them regenerates
`docs/metrics_reference.md`, so whichever lands later needs a rebase and a re-run of
`make generate-metrics-doc`.

### Why this is worth doing beyond tidiness

Sampling a value that is available on demand costs a goroutine, a ticker and a config key, and makes
the reading up to one interval stale — but the real cost is a wiring step that can be lost. The
sidecar's sampler was passed to `StartPrometheusServer` as a callback; 2b0e144f (#544, 2026-05-06)
removed that plumbing and orphaned it, so two gauges the tuning guide tells operators to watch have
read a flat `0` for three months. A flat zero reads as "healthy", not "not wired".

[...]

PR #735, opened 2026-08-03
[docs] Scan every metrics file when generating the reference

### LINKED ISSUE #732: [docs] Scan every metrics file so the dependency graph metrics are documented
Part of #730.

`scripts/metrics_doc.sh:46-51` passes exactly one metrics file per service to the extractor:

```bash
generate_service_doc "Coordinator" "service/coordinator/metrics.go"
```

The coordinator's metrics are split across two packages, so
`service/coordinator/dependencygraph/metrics.go` is never read and **none** of its 16 metrics
appear in `docs/metrics_reference.md`. That includes the two queue gauges the rest of #730
converts, plus the local/global dependency graph counters and the ten histograms that time graph
construction and validated-batch processing — the metrics an operator would reach for to explain a
dependency-graph stall.

The gap is silent: `make check-metrics-doc` compares the generated file against itself, so a
metrics file that is never scanned always passes.

### Proposal

Make `generate_service_doc` accept more than one metrics file per service section, and pass both
coordinator files. The dependency graph's metrics belong in the existing Coordinator section rather
than a section of their own — they are reported by the coordinator process on the same endpoint.

Worth landing before the rest of #730, so the queue gauges that issue touches are documented when
they change.

liran-funaro:
#### Type of change

- Bug fix
- Documentation update

#### Description

- `generate_service_doc` took a single metrics file, so a service whose metrics are split across
  packages was only partly documented. It now takes one or more, and the Coordinator section covers
  `service/coordinator/dependencygraph/metrics.go` as well.
- That adds the dependency graph's 16 metrics to `docs/metrics_reference.md`. No series is renamed
  or removed; the diff is larger than 16 rows only because the table is re-padded.

The gap was invisible to CI: `make check-metrics-doc` regenerates the doc and compares it to the
committed one, so a metrics file that is never scanned agrees with itself. Nothing else in the
tree is affected — `metrics_doc_extract.py` already parsed this file correctly, it was simply never
handed to it.

#### Related issues

- resolves #732
- part of #730

Signed-off-by: Liran Funaro <liran.funaro@gmail.com>

### DIFF
--- a/scripts/metrics_doc.sh
+++ b/scripts/metrics_doc.sh
@@ -13,9 +13,11 @@ extract_metrics_script="${repo_root_dir}/scripts/metrics_doc_extract.py"
 metrics_doc="${repo_root_dir}/docs/metrics_reference.md"
 python_cmd="${python3:-${PYTHON_CMD}}"
 
+# generate_service_doc <service name> <metrics file>... - one section per service, over one or
+# more metrics files, for a service whose metrics are split across packages.
 generate_service_doc() {
   local service_name="$1"
-  local metrics_file="$2"
+  shift
 
   cat <<EOF
 ## ${service_name} Metrics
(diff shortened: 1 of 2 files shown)


scripts/metrics_doc.sh
```diff
@@ -44,7 +49,9 @@ SPDX-License-Identifier: Apache-2.0
 EOF
 
   generate_service_doc "Sidecar" "service/sidecar/metrics.go"
-  generate_service_doc "Coordinator" "service/coordinator/metrics.go"
+  generate_service_doc "Coordinator" \
+    "service/coordinator/metrics.go" \
+    "service/coordinator/dependencygraph/metrics.go"
```
cendhu, line 54: Good catch.

cendhu (approved):

Issue #734, opened 2026-08-03
[verifier, sidecar] Report per-stream queues as one series per stream


liran-funaro:
Part of #730, covering its Part 3.

Two queues exist once per stream, with several live at a time, so a gauge that closes over a single
channel cannot address them:

- `service/verifier/parallel_executor.go:39-41` `inputCh`, `outputSingleCh`, `outputCh` — a new
  executor per stream (`verifier_server.go:77`). The verifier exports **no** queue gauge at all
  today, only `active_streams` and `active_requests`, so verification backpressure is invisible.
- `service/sidecar/notify.go:241` `blockQueue` — the block buffer of one all-transactions
  subscription. `registerAllTxStream` / `unregisterAllTxStream` are the natural hooks.

### Proposal

Report each queue as its own series, labelled by the stream, and let the operator aggregate:
`sum(<metric>)` for the total, `max(<metric>)` to catch one saturated worker while the total still
looks healthy.

`prometheus` has no `GaugeFuncVec`, so this needs a small custom `Collector` in `utils/monitoring`
that holds the registered channels and emits one gauge per channel on `Collect`. Two properties
matter:

- **No sampler.** The length is read on the scrape path, same as the rest of #730.
- **A series exists only while its stream does.** Registering on stream start and unregistering on
  stream end means cardinality follows the number of *live* streams rather than growing without
  bound. It also avoids the worse failure: a series left behind keeps reporting the depth of a queue
  nothing drains, which reads as a stuck worker rather than a finished one.

Re-registering the same label values must replace the channel rather than emit the label twice —
prometheus rejects a duplicate series mid-scrape.

### Also in scope

- `docs/performance-tuning.md` should say how to read a per-stream queue metric, since `sum` over a
  labelled gauge is not obvious from the metric name alone.
- `scripts/metrics_doc_extract.py` needs to recognise an explicitly instantiated generic helper
  (`NewChannelLenGaugeVec[[]*pkg.Type](...)`, including the nested brackets of a slice element) and
  a label list passed as a variable rather than a literal. Both are dropped from the generated doc
  silently today.

### Not in scope

The per-stream queues that #730 lists as needing a metric-shape decision are only these two. The
coordinator managers' `txBatchQueue` turned out to be **one per `run` shared by all per-endpoint
senders**, not one per stream, so it needs an `atomic.Pointer`, not a label — it is handled in the
conversion work instead.

PR #736, opened 2026-08-03
[sidecar] Report queue sizes on scrape

### LINKED ISSUE #733: [sidecar] Report queue sizes on scrape instead of sampling them
Part of #730, narrowed to the sidecar. The equivalent work for `vc` and the coordinator is tracked
separately, so each service can be reviewed on its own.

### The sidecar's monitor is dead code

`(*Service).monitorQueues` (`service/sidecar/sidecar.go:433`) is never called. It used to be passed
to `StartPrometheusServer` as a callback, and 2b0e144f (#544, 2026-05-06) removed that plumbing and
orphaned the sampler. So `sidecar_relay_input_block_queue_size` and
`sidecar_relay_output_committed_block_queue_size` have reported a flat `0` since May, while
`docs/performance-tuning.md:47` and `:49` tell operators to watch both. A flat zero reads as
"healthy", not "not wired".

`golangci-lint` does flag the dead method, but `make lint-go` runs with `--new-from-rev=main`, so a
pre-existing finding never surfaces.

With a gauge evaluated on scrape the closure is attached at registration, so there is no second call
site left to lose — the bug becomes unreachable rather than merely fixed.

### Sidecar queues with no gauge

| Channel | Suggested series |
| --- | --- |
| `service/sidecar/sidecar.go:107` `committedBlockWithTxs` | `sidecar_notifier_input_block_queue_size` |
| `service/sidecar/sidecar.go:108` `statusQueue` | `sidecar_notifier_input_status_queue_size` |
| `service/sidecar/notify.go:109` `requestQueue` | `sidecar_notifier_request_queue_size` |
| `service/sidecar/notify.go:123` `timeoutQueue` | `sidecar_notifier_timeout_queue_size` |
| `service/sidecar/relay.go:100` `mappedBlockQueue` | `sidecar_relay_mapped_block_queue_size` |
| `service/sidecar/relay.go:108` `statusBatch` | `sidecar_relay_input_status_batch_queue_size` |

### Queues replaced every session

Three of these live for one coordinator session rather than for the service, and rebuilding them is
how a failed session's enqueued blocks are dropped — they belong to a stream that no longer exists.
Promoting them to a plain field would leak those leftovers into the next session, so they need an
`atomic.Pointer` the session installs: the gauge follows whichever channel is current, and an unset
pointer reports zero rather than the previous session's last value. A plain field would also race,
since the session writes it while a scrape reads it.

This applies to the sidecar's `blocksToBeCommitted` and the relay's two internal queues.

### Constraints

- The queues must exist before the metrics that observe them are registered, so the service
  constructor has to build them first and pass them in.

### Metric-surface impact

No rename. The converted gauges become exact at scrape time instead of up to one interval stale, so
brief spikes the sampler missed will start showing, and the two gauges above begin reporting real
values for the first time since #544 — which will look like a regression on a dashboard that has
learned they are always zero.

liran-funaro:
#### Type of change

- Bug fix
- New feature

#### Description

- Delete the dead `(*Service).monitorQueues` and report the relay's queue sizes as gauges
  evaluated on scrape instead. The closure is attached at registration, so there is no second
  call site left to lose.
- Add `Provider.NewChannelLenGauge` for a channel that lives as long as the service, and
  `Provider.NewAtomicChannelLenGauge` for one replaced at runtime. Go 1.27 lets a method carry
  its own type parameter, so both sit on `Provider` beside `NewGauge` and `NewGaugeFunc` instead
  of being free functions that take one.
- Gather the sidecar's queues into a `queues` struct built before the metrics that observe them,
  so a queue gauge cannot be registered against a channel that does not exist yet.
- Add gauges for the six queues that had none: `sidecar_relay_mapped_block_queue_size`,
  `sidecar_relay_input_status_batch_queue_size` and the four `sidecar_notifier_*` queues.

The `queues` struct separates the two lifetimes, since the difference is what decides how each
queue is reported. The fixed group is created once, so a gauge can close over the channel itself.
The other three are replaced by `queues.reset()` at the start of every coordinator session --
dropping the old channel is how the blocks a failed session had enqueued are discarded, as they
belong to a stream that no longer exists -- so a gauge cannot close over them and reads through an
`atomic.Pointer` instead, reporting whichever channel is current, or zero between sessions.
`reset` installs all three and the session reads them back through the pointers, and it sizes them
from the same buffer size as the fixed queues so the two cannot drift.

Those pointers live on `queues` rather than on `perfMetrics`, so the metrics struct holds gauges
only and never a queue, and the relay holds no reference to `queues` at all -- it receives the two
queues connecting its own stages through `relayRunConfig`, alongside the session queues it was
already passed.

Note that `relayRunConfig` now carries channels of two kinds: the outgoing ones, where a `nil` is
a supported "not wired" state that `channel.Write` short-circuits, and `mappedBlockQueue` /
`statusBatch`, which are required -- a `nil` there would silently drop every block rather than
fail.

No metric is renamed. `sidecar_relay_input_block_queue_size` and
`sidecar_relay_output_committed_block_queue_size` begin reporting real values for the first time
since #544, so a dashboard that has learned they are always zero will show what looks like a
regression.

Scoped to the sidecar alone. The equivalent conversions for `vc` (#770) and the coordinator (#771)
follow separately and reuse the helpers added here, and the sidecar's one genuinely per-stream queue
is #734.

#### Related issues

- resolves #733
- related to #730


cendhu:
@liran-funaro This PR seems to address multiple things at a time. Limited scoped PRs are easier and quicker to review.

liran-funaro:
@cendhu it is a single topic - use gague func instead of monitor.

cendhu:
> @cendhu it is a single topic - use gague func instead of monitor.

Is this PR not adding any new metrics? Further, this new gauge func can be added to each service in an individual PR.

liran-funaro:
@cendhu Adding the metrics is complementary since we gather the queues to a struct. 
I can split into per service PR if it helps.

service/coordinator/metrics.go
```diff
@@ -71,27 +85,39 @@ func newPerformanceMetrics(q *channels) *perfMetrics {
...
 // newManagerMetrics creates the metric set of a single service manager. The namespace and
-// subsystem identify which manager reports them, and the two queues are reported on scrape.
+// subsystem identify which manager reports them. Both queues are reported on scrape and are
+// only ever read from here, so both are typed receive-only.
 func newManagerMetrics(
 	p *monitoring.Provider,
 	params monitoring.MetricsParameters,
-	inputQueue <-chan dependencygraph.TxNodeBatch,
-	outputQueue chan<- dependencygraph.TxNodeBatch,
+	inputQueue, outputQueue <-chan dependencygraph.TxNodeBatch,
```
cendhu: Do we need to specify the chan read/write with <-chan? I am confused because on the left we have both <-chan and chain<-.

service/coordinator/metrics.go
```diff
@@ -105,21 +131,25 @@ func newManagerMetrics(
...
-		}),
+		}, outputQueue),
 	}
+	m.pendingQueueSize = monitoring.NewAtomicChannelLenGauge(p, prometheus.GaugeOpts{
+		Namespace: params.Namespace,
+		Subsystem: params.Subsystem,
+		Name:      "pending_batch_queue_size",
+		Help: "Size of the queue the manager's per-endpoint senders draw from, including the " +
+			"batches re-queued after a stream failure.",
+	}, &m.pendingQueue)
```
cendhu: The link between pendingQueueSize and pendingQueue is confusing even after reading the comments in managerMetrics. I am not sure whether the metrics can hold a queue.

service/sidecar/metrics.go
```diff
@@ -30,8 +37,18 @@ type perfMetrics struct {
...
+	statusBatchQueueSize            prometheus.GaugeFunc
+	committedBlocksQueueSize        prometheus.GaugeFunc
+
+	// The relay's queues are recreated for every coordinator session, so the gauges above report
+	// them through these pointers, which run installs. Dropping the old channel with the session
+	// is deliberate: its leftovers belong to a stream that failed. An unset pointer reports zero,
+	// which is what a scrape between sessions should see rather than the last session's value.
+	relayInputBlockQueue  atomic.Pointer[chan *common.Block]
+	relayMappedBlockQueue atomic.Pointer[chan *blockMappingResult]
+	relayStatusBatchQueue atomic.Pointer[chan *committerpb.TxStatusBatch]
```
cendhu: This pattern is repeating. I don't know whether it is correct to have pointer to internal data structure within the metrics object. Reads unnatural to me. Passing of queue as an argument in newPerformanceMetrics() reads fine to me.

cendhu (commented):
My observation was that the earlier method was very simple though it was an approximation on monitoring the queue size. Now passing of queues and also keeping pointers to queue make it hard to maintain a mental model. If you insist on this change, at least avoid keeping pointers to the queue in the metrics object.

service/coordinator/dependencygraph/metrics.go
```diff
@@ -45,21 +53,21 @@ type perfMetrics struct {
 	dependentTransactionsQueueSize prometheus.Gauge
```
cendhu, line 55: why do we ignore this queue size and gdgWaitingTxQueueSize?

liran-funaro:
@cendhu I fixed the issues. I split this PR for per service PR to simplify the review.

docs/metrics_reference.md
```diff
@@ -18,8 +18,11 @@ The following Sidecar metrics are exported for consumption by Prometheus.
...
+| sidecar_notifier_input_block_queue_size                    | gauge     |                                  | Size of the committed block queue delivered from the relay to the notifier.                                                                              |
+| sidecar_notifier_input_status_queue_size                   | gauge     |                                  | Size of the transaction status queue delivered from the relay to the notifier.                                                                           |
+| sidecar_notifier_request_queue_size                        | gauge     |                                  | Size of the queue of notification requests received from clients.                                                                                        |
+| sidecar_notifier_timeout_queue_size                        | gauge     |                                  | Size of the queue of notification requests waiting to be timed out.                                                                                      |
```
cendhu:
```suggestion
| sidecar_notifier_timeout_queue_size                        | gauge     |                                  | Size of the queue of notification requests timed out.                                                                                      |
```

cendhu (approved):
LGTM

Issue #752, opened 2026-08-06
Correctly process a config TX w/o getting TX-ID from outer envelope


tock-ibm:
### The problem

Committer rejects config-tx with `committerpb.Status_MALFORMED_MISSING_TX_ID, "no TX ID"`

See PR from fabric-x-orderer, where a contributor tried to fix it in the wrong place: https://github.com/hyperledger/fabric-x-orderer/pull/1083

### Why

When the committer processes TXs, here:

https://github.com/hyperledger/fabric-x-committer/blob/260027471523dc2383127b9e47fe1b1356038f9a/service/sidecar/mapping.go#L129

It first extracts the TX-ID, and only then classifies whether it is a `common.HeaderType_CONFIG`, here:

https://github.com/hyperledger/fabric-x-committer/blob/260027471523dc2383127b9e47fe1b1356038f9a/service/sidecar/mapping.go#L135

The config TX outer envelope does not carry a TxID. It is created and signed by the consensus leader, so the `creator` and `nonce` are different from the clients `creator` and `nonce`. The clients TxID is in the transaction that is within the `ConfigEnvelope.LastUpdate` field. 

You also do not need to do policy.ValidateConfigTX( ) on a config TX, the orderer does that. A config TX cannot be invalidated at the committer. Any failure to process a new config TX is basically fatal.

### Solution

First classify the TX, and if config, just append and apply.
If you need a TX-ID in order to notify the client, extract it from the `ConfigEnvelope.LastUpdate` field.

tock-ibm:
@elishambadi  please follow the issue here, or maybe push a PR here?

elishambadi:
Hey @liran-funaro.
I've pushed a PR for the fix here: https://github.com/hyperledger/fabric-x-committer/pull/753

tock-ibm:
See also: https://github.com/hyperledger/fabric-x-orderer/pull/1083#issuecomment-5205046288

tinystone007:
[...]

PR #753, opened 2026-08-07
fix config_tx_id validation

elishambadi:
#### Type of change
- Bug fix
 
#### Description

- no TxID check for CONFIG case, moved to MESSAGE case only. As config TX don't have outer TxID in channel header.
- CONFIG case no longer runs policy validation as it is already done by the orderer.
- add txID from config.LastUpdate to TxRef for notification.
- Updated tests for `service/sidecar/mapping_test.go`
- Docs update.

#### Related issues

Fixes issue: https://github.com/hyperledger/fabric-x-committer/issues/752

### DIFF
--- a/service/sidecar/mapping.go
+++ b/service/sidecar/mapping.go
@@ -21,6 +21,7 @@ import (
 	"github.com/hyperledger/fabric-x-committer/api/servicepb"
 	"github.com/hyperledger/fabric-x-committer/service/verifier/policy"
 	"github.com/hyperledger/fabric-x-committer/utils"
+	"github.com/hyperledger/fabric-x-committer/utils/retry"
 	"github.com/hyperledger/fabric-x-committer/utils/serialization"
 )
 
@@ -106,7 +107,8 @@ func mapBlock(block *common.Block, txIDToHeight *utils.SyncMap[string, servicepb
 		logger.Debugf("Mapping transaction [blk,tx] = [%d,%d]", blockNumber, msgIndex)
 		err := mappedBlock.mapMessage(uint32(msgIndex), msg) //nolint:gosec // int -> uint32.
 		if err != nil {
-			// This can never occur unless there is a bug in the relay.
+			// This can never occur unless there is a bug in the relay or a
+			// mapping error which indicates a bad block.
 			return nil, err
 		}
 	}
@@ -129,10 +131,17 @@ func (b *blockMappingResult) mapMessage(msgIndex uint32, msg []byte) error {
 	}
 	switch common.HeaderType(envLite.HeaderType) {
 	case common.HeaderType_CONFIG:
-		// extracts configId from config.LastUpdate.TxId, which is the client's TxID.
-		ref.TxId = envLite.TxID
+		// TxID is extracted from the ConfigEnvelope.LastUpdate field.
+		ref.TxId = extractConfigTxID(envLite.Data)
 		if ref.TxId == "" {
-			ref.TxId = extractConfigTxID(envLite.Data)
+			return b.rejectNonDBStatusTx(ref, committerpb.Status_MALFORMED_MISSING_TX_ID, "no TX ID")
+		}
+		// We must validate the config TX here. If the mapping fails, it will be retried with a
+		// backoff error, which restarts the block source feed from one block before and
+		// tries again, finally failing if the parsing keeps failing.
+		if err := policy.ValidateConfigTx(msg); err != nil {
+			return errors.Join(retry.ErrBackOff,
+				fmt.Errorf("invalid config TX [%s]: %w", ref.TxId, err))
 		}
 		b.isConfig = true
 		return b.appendTx(ref, configTx(msg))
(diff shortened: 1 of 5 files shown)


service/sidecar/mapping.go
```diff
@@ -126,19 +127,20 @@ func (b *blockMappingResult) mapMessage(msgIndex uint32, msg []byte) error {
...
 	}
-	if envLite.TxID == "" || !utf8.ValidString(envLite.TxID) {
-		return b.rejectNonDBStatusTx(ref, committerpb.Status_MALFORMED_MISSING_TX_ID, "no TX ID")
-	}
-	ref.TxId = envLite.TxID
-
 	switch common.HeaderType(envLite.HeaderType) {
 	case common.HeaderType_CONFIG:
-		if err := policy.ValidateConfigTx(msg); err != nil {
-			return b.rejectTx(ref, committerpb.Status_MALFORMED_CONFIG_TX_INVALID, err.Error())
```
liran-funaro, line 137: `committerpb.Status_MALFORMED_CONFIG_TX_INVALID` should be removed in fabric-x-common

service/sidecar/mapping.go
```diff
@@ -126,19 +127,20 @@ func (b *blockMappingResult) mapMessage(msgIndex uint32, msg []byte) error {
...
 		return b.rejectNonDBStatusTx(ref, committerpb.Status_MALFORMED_BAD_ENVELOPE, envErr.Error())
 	}
-	if envLite.TxID == "" || !utf8.ValidString(envLite.TxID) {
-		return b.rejectNonDBStatusTx(ref, committerpb.Status_MALFORMED_MISSING_TX_ID, "no TX ID")
-	}
-	ref.TxId = envLite.TxID
-
 	switch common.HeaderType(envLite.HeaderType) {
 	case common.HeaderType_CONFIG:
-		if err := policy.ValidateConfigTx(msg); err != nil {
```
liran-funaro, line 136: We must still validate the config TX here as we cannot handle config parsing failure later on the process.
At the delivery side we also parse the config block, but we parse different parts.
There, upon failure we try to fetch the block from other source, and finally fail if the parsing keeps failing.
We cannot try to fetch from different source at this point, so we can restart the the block source feed from one block before and try again.

service/sidecar/mapping.go
```diff
@@ -126,19 +127,20 @@ func (b *blockMappingResult) mapMessage(msgIndex uint32, msg []byte) error {
...
-			return b.rejectTx(ref, committerpb.Status_MALFORMED_CONFIG_TX_INVALID, err.Error())
+		// extracts configId from config.LastUpdate.TxId, which is the client's TxID.
+		ref.TxId = envLite.TxID
```
liran-funaro: We should not use the `envLite.TxID` as the TX ID as this was issued by the consenter.
We need to use the user issued TX ID exclusively, which is what extractConfigTxID.
Similarity to handling the config parsing, we should fail and retry with different source once we have empty TX ID.
The committer must have a valid TX ID to properly process any TX.

service/sidecar/mapping.go
```diff
@@ -281,6 +283,20 @@ func debugTx(ref *committerpb.TxRef, format string, a ...any) {
 	logger.Debugf("ID [%s]: %s", txID, fmt.Sprintf(format, a...))
 }
 
+// extractConfigTxID returns the client's TxID from a config envelope. The outer envelope of a
+// config TX carries no TxID (it is created and signed by the consensus leader).
+func extractConfigTxID(value []byte) string {
+	configEnv, err := protoutil.UnmarshalConfigEnvelope(value)
+	if err != nil {
+		return ""
```
liran-funaro, line 301: We should at least log the error here

service/sidecar/mapping.go
```diff
@@ -281,6 +283,20 @@ func debugTx(ref *committerpb.TxRef, format string, a ...any) {
...
+	}
+	chdr, err := protoutil.ChannelHeader(configEnv.LastUpdate)
+	if err != nil {
+		return ""
```
liran-funaro, line 306: We should at least log the error here

liran-funaro (commented):
Thanks for the PR.
We should decide on how to proceed in case we fail to parse a config block.
This can be discussed in the issue, not in the PR.
Once we have a decision we can proceed with this PR.

service/sidecar/mapping.go
elishambadi: Raised a conversation point on the issue. https://github.com/hyperledger/fabric-x-committer/issues/752#issuecomment-5265223148

service/sidecar/mapping.go
elishambadi: Okay. Restored config envelope validation here with `retry.ErrBackOff`.

service/sidecar/mapping.go
elishambadi: Resolved.

Either obtain txID from config.LastUpdate or return the noTXID error.

service/sidecar/mapping.go
elishambadi: Resolved.

service/sidecar/mapping.go
elishambadi: Resolved.

service/sidecar/mapping.go
elishambadi: Once that is settled then we can remove on fabric-x-common and also handle parsing failure.

elishambadi:
Closed by PR #768 .

PR #759, opened 2026-08-11
[loadgen] Query committed read versions during load (queries-rate)

liran-funaro:
#### Type of change

- New feature
- Test update

#### Description

Add a query stage to the load generator so a single run drives commit traffic and
query traffic together: a rate-controlled subset of each transaction's reads fetches
its committed version from the query service before the transaction is signed.

- `queries-rate` (`load-profile.transaction`): average reads per transaction whose
  committed version is fetched; validated `>= 0` and `<=` the version-bearing reads
  (`read-only-count + read-write-count`), independent of `key-backref-rate`.
- `queryFiller` selects the rate-controlled reads via a persistent fractional
  accumulator (accurate at any batch size, including 1), groups them by namespace, and
  fetches their versions with a single `GetRows` (one query-namespace per distinct
  namespace), filling the fetched version on each hit (a miss keeps the nil version).
- The query service maps a query for a non-existent namespace table to `NotFound`, and
  `queryFiller` tolerates `NotFound` as an all-miss, leaving those reads at their nil
  version.
- `query-client` (top-level `connection.MultiClientConfig`): the query-service
  connection, required when `queries-rate > 0`. `TxStream.Run` owns its lifetime — it
  dials one load-balanced connection per worker (round-robining across the configured
  endpoints), builds a `queryFiller` per worker, and closes the connections on return.
- Integration config: `SystemConfig.LoadGenQueriesRate` drives `queries-rate` in the
  shared template and, when positive, renders a `query-client` block pointing at the
  runtime's query service; rate 0 (default) leaves the stage off and renders no
  `query-client`, so existing configs are unchanged.

Reads that miss keep their nil version; with `key-backref-rate` at its default every
read is a fresh key, so queries miss by design. The tests exercise the wiring and
parallel query+commit load, not version-hit correctness under contention.

#### Related issues

- related to #671


loadgen/workload/query.go
```diff
@@ -0,0 +1,191 @@
...
+		acc    float64
+	}
+
+	// keyVersionQuerier is the minimal seam a queryFiller needs to fetch committed key versions. It is
+	// satisfied by committerpb.QueryServiceClient — a later task dials the real client and injects it, one
+	// per worker; tests inject a stub.
+	keyVersionQuerier interface {
+		GetRows(ctx context.Context, in *committerpb.Query, opts ...grpc.CallOption) (*committerpb.Rows, error)
+	}
+
```
cendhu, line 41: `yagni:` `keyVersionQuerier` is a second seam for a single path. `batchQuerier` in `stream.go` is already the injection point, and `stream_test.go` already exercises a real in-process `fakeQueryServiceServer` over a real dial. Take `committerpb.QueryServiceClient` concretely in `newQueryFiller` and delete this interface.

loadgen/workload/query.go
```diff
@@ -0,0 +1,191 @@
...
+	// nsSelection is the subset of a single namespace's version-bearing reads selected for querying, across
+	// the whole batch.
+	nsSelection struct {
+		reads      []*applicationpb.Read
+		readWrites []*applicationpb.ReadWrite
+	}
```
cendhu, line 47: `shrink:` `nsSelection{reads, readWrites}` + `dedupKeys` (below) re-dedup what a key-keyed map gives for free. Make the selection `map[nsID]map[key]*targets`; then `buildQuery` just iterates map keys and `dedupKeys` disappears (~20 lines).

loadgen/workload/query.go
```diff
@@ -0,0 +1,191 @@
...
+func selectPriorityReads(tx *applicationpb.Tx, k int, selected map[string]*nsSelection) {
+	for _, ns := range tx.GetNamespaces() {
+		if k <= 0 {
+			return
+		}
+		readsOnly := ns.GetReadsOnly()
+		readWrites := ns.GetReadWrites()
+		if len(readsOnly) == 0 && len(readWrites) == 0 {
+			continue
+		}
```
cendhu, line 127: `delete:` this empty-namespace guard exists only to avoid creating an empty `nsSelection`. Create `sel` lazily on first append instead.

loadgen/workload/query.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	for _, r := range reads {
+		add(r.Key)
+	}
+	for _, rw := range readWrites {
+		add(rw.Key)
+	}
+	return keys
+}
+
```
cendhu, line 176: `shrink:` `dedupKeys` hand-rolls dedup with its own `seen` map. Falls out for free if the selection is keyed by key (see the `nsSelection` comment).

cendhu (commented):
Ponytail review — over-engineering / complexity only (correctness, security and perf explicitly out of scope for this pass). The 42P01 -> NotFound mapping in `service/query` is lean; everything below is in loadgen/config/tests.

Net: roughly **-280 lines** possible.

loadgen/workload/query.go
```diff
@@ -0,0 +1,191 @@
...
+	for _, ns := range namespaces {
+		rs := ns.GetRows()
+		nsVersions := make(map[string]uint64, len(rs))
+		for _, row := range rs {
+			nsVersions[string(row.GetKey())] = row.GetVersion()
+		}
+		versions[ns.GetNsId()] = nsVersions
+	}
+	return versions
+}
```
cendhu, line 191: `shrink:` `versionsByNamespace` materializes a full nested `map[ns]map[key]uint64`, then `fillSelectedVersions` walks the selection to look back into it. One loop over `rows.GetNamespaces()` -> rows, setting the version on selection hits, collapses both functions into one (~30 lines).

loadgen/workload/stream.go
```diff
@@ -68,20 +71,59 @@ func NewTxStream(profile *Profile, options *StreamOptions, counter *TxCounter) (
 	}, nil
 }
 
-// Run starts the stream workers.
-func (s *TxStream) Run(ctx context.Context) error {
+// Run starts the stream workers. When queryClient enables the query stage (set and queries-rate > 0) Run
+// dials one load-balanced query-service connection per worker and each worker versions a rate-controlled
+// subset of its batches' reads before signing; otherwise every worker runs without a querier. The dialed
+// connections are closed when Run returns, coupling their lifetime to the run.
+func (s *TxStream) Run(ctx context.Context, queryClient *connection.MultiClientConfig) error {
```
cendhu, line 78: `shrink:` adding a param to `Run` breaks the `Run(ctx)` service shape and forces three closure wrappers in tests (`stream_test.go` x3) plus plumbing through `client.go`. Pass the query-client config to `NewTxStream` (or `StreamOptions`) and keep `Run(ctx)`.

loadgen/workload/stream.go
```diff
@@ -68,20 +71,59 @@ func NewTxStream(profile *Profile, options *StreamOptions, counter *TxCounter) (
...
+		}
+		conns = append(conns, conn)
+		queriers[i] = newQueryFiller(committerpb.NewQueryServiceClient(conn), s.gens[i].QueriesRate)
+	}
+	return queriers, conns, nil
+}
+
 // AppendBatch appends a batch to the stream.
 func (s *TxStream) AppendBatch(ctx context.Context, batch []*servicepb.LoadGenTx) {
 	channel.NewWriter(ctx, s.queue).Write(batch)
```
cendhu, line 129: `yagni:` one `grpc.ClientConn` per worker. A ClientConn multiplexes and is safe to share; the "independent flow-control / stream-concurrency headroom" rationale is speculative and unmeasured here. Share one load-balanced conn and this whole function collapses: no `conns` slice, no partial-dial cleanup, no error branch. Keep a `queryFiller` per worker (it owns the accumulator state), just share the client.

`shrink:` also, the 10-line doc preamble here plus the 20-line essay on the 3-line `batchQuerier` interface above are longer than the code they describe. Three lines each: what it does, and nil = no query stage.

loadgen/config.go
```diff
@@ -34,5 +43,11 @@ func (c *ClientConfig) Validate() error {
...
 		return nil
 	}
-	return c.LoadProfile.Transaction.Validate()
+	if err := c.LoadProfile.Transaction.Validate(); err != nil {
+		return err
+	}
+	if c.LoadProfile.Transaction.QueriesRate > 0 && c.QueryClient == nil {
+		return errors.New("queries-rate > 0 requires query-client")
+	}
+	return nil
```
cendhu, line 52: `shrink:` the cross-field check itself is fine, but the doc paragraph on `QueryClient` above is the third copy of the same text (`loadgen.yaml` and `TransactionProfile.QueriesRate` already carry it). One line + pointer.

loadgen/workload/config_test.go
```diff
@@ -45,3 +45,62 @@ func TestTransactionProfileValidate(t *testing.T) {
...
+			require.Error(t, tc.profile.Validate())
+		})
+	}
+}
+
+// withQueriesRate returns a copy of p with QueriesRate set.
+func withQueriesRate(p TransactionProfile, rate float64) TransactionProfile {
+	p.QueriesRate = rate
+	return p
+}
```
cendhu, line 100: `shrink:` `withQueriesRate` / `withKeyBackrefRate` are copy-helpers for setting one field. Inline the literals in the table (`TransactionProfile{ReadOnlyCount: 2, ReadWriteCount: 2, BlindWriteCount: 3, QueriesRate: 4}`) and drop both helpers.

loadgen/workload/query_test.go
```diff
@@ -0,0 +1,392 @@
...
+	tx := newTestTx([][]byte{[]byte("ro")}, [][]byte{[]byte("rw")})
+	stub := &stubKeyVersionQuerier{versions: singleNsVersions(map[string]uint64{"ro": 1, "rw": 2})}
+	f := newQueryFiller(stub, 0)
+
+	require.NoError(t, f.FillVersions(t.Context(), []*applicationpb.Tx{tx}))
+
+	require.Empty(t, stub.calls)
+	requireVersions(t, tx, nil)
+}
+
```
cendhu, line 141: `delete:` rate 0 is already covered by the "batches with a zero budget must not issue a GetRows call" assertion in `TestFillVersionsAverageConvergesAtBatchSizeOne`. If you want it explicit, add a `rate: 0` row to the priority table rather than a standalone test.

loadgen/client_test.go
```diff
@@ -119,6 +120,58 @@ func TestLoadGenForVCService(t *testing.T) {
...
+
+	queryMetricsURL, err := monitoring.MakeMetricsURL(querySrvConf.HTTP.Endpoint.Address(), &e.clientTLSConfig)
+	require.NoError(t, err)
+	queryMetricsTLSConfig := test.MustGetTLSConfig(t, &e.clientTLSConfig)
+	require.EventuallyWithT(t, func(ct *assert.CollectT) {
+		v := test.GetMetricValueFromURL(
+			ct, queryMetricsURL, "queryservice_grpc_key_requested_total", queryMetricsTLSConfig,
+		)
+		assert.Positive(ct, v)
+	}, 2*time.Minute, time.Second)
```
cendhu, line 172: `delete:` this duplicates `integration/test/loadgen_test.go:TestLoadGenWithQueryService` — same claim (the wire is connected, the query metric goes positive) — while dragging the loadgen unit suite onto the shared DB container and hand-tuning a full `query.Config`. Keep the integration test.

cmd/config/app_config_test.go
```diff
@@ -563,6 +563,48 @@ func TestLoadGenTransactionProfileValidation(t *testing.T) {
...
+			t.Parallel()
+			require.NoError(t, read(t, tc.yaml))
+		})
+	}
+
+	t.Run("queries-rate positive without query-client is rejected", func(t *testing.T) {
+		t.Parallel()
+		err := read(t, withOneRead)
+		require.ErrorContains(t, err, "query-client")
+	})
```
cendhu, line 605: `shrink:` a `read` closure + 2 consts + a 2-row all-`NoError` table + a separate reject subtest is more scaffolding than the check. One table with a `wantErr` field, or three flat asserts.

cmd/config/samples/loadgen.yaml
```diff
@@ -114,6 +114,25 @@ orderer-client:
...
+#   tls: *tlsConfig
+#   # Optional exponential backoff retry for the query connections. GetRows retries transient
+#   # gRPC failures (UNAVAILABLE / DEADLINE_EXCEEDED / RESOURCE_EXHAUSTED) with this backoff and
+#   # fails the run once the budget is spent. Default:
+#   # reconnect:
+#   #   initial-interval: 500ms
+#   #   randomization-factor: 0.5
+#   #   multiplier: 1.5
+#   #   max-interval: 10s
+#   #   # Total retry budget. Set to 0s for unlimited retries (never give up).
```
cendhu, line 133: `shrink:` 12 commented lines re-documenting the `reconnect` backoff defaults, duplicated from the other `*-client` blocks. `tls: *tlsConfig` plus one line pointing at the shared retry docs.

PR #765, opened 2026-08-12
[metrics] Extract duplicated namespace, subsystem and name literals in service metrics to consts

### LINKED ISSUE #764: [metrics] Metrics doc generator ignores const namespace/subsystem values
The metrics doc generator (`scripts/metrics_doc_extract.py`) only reads
`Namespace`/`Subsystem`/`Name` when they are written as quoted string literals.
When a metric declares them through a Go `const` instead (e.g.
`Namespace: namespace`), the field resolves to empty and the metric is dropped
from `docs/metrics_reference.md` **silently** — no error, no warning.

_Originally posted by @dean-amar in https://github.com/hyperledger/fabric-x-committer/pull/748#discussion_r3757053433_

dean-amar:
#### Type of change

#### Description

Building on the extractor's new const support, this replaces the repeated `Namespace`/`Subsystem` literals (and the `Name` values intentionally shared across subsystems in vc and dependencygraph) with local consts across every service's `metrics.go`. 

#### Related issues

- resolves #764

### DIFF
--- a/loadgen/metrics/metrics.go
+++ b/loadgen/metrics/metrics.go
@@ -18,6 +18,8 @@ import (
 	"github.com/hyperledger/fabric-x-committer/utils/monitoring/promutil"
 )
 
+const namespace = "loadgen"
+
 type (
 	// PerfMetrics is a struct that contains the metrics for the block generator.
 	PerfMetrics struct {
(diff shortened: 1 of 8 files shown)


liran-funaro (approved):
LGTM

PR #782, opened 2026-08-26
[vc] Schedule snapshot hashing through a fenced lease

### LINKED ISSUE #663: I-13: [vc] Snapshot crash recovery
# PR-13 — committer: VC snapshot crash recovery (US1b-iii)

**Tasks**: T028 (see `the feature task breakdown`)
**Depends on**: PR-12 (worker + status transitions exist).
**Blocks**: PR-14 (query surfaces the recovered rows/statuses).

> Read `docs/core-concurrency-pattern.md` before touching startup/goroutine lifecycle.

## Scope

VC-only. Startup recovery so an interrupted snapshot resumes cleanly.

- **T028** — in `service/vc/snapshot.go`, on VC startup scan `_snapshot` namespace rows for status PENDING/IN_PROGRESS/FAILED and re-enqueue them: decode each row's `SnapshotState`, read `clone_database`, run `DROP DATABASE IF EXISTS <clone_database>` **before** re-cloning, then re-enqueue into the hash worker (PR-12). No timestamp field is needed for this scan (there are none on `SnapshotState`). CHECKPOINTED rows are terminal and skipped, per `the applicable API contract` background step 6 and FR-016.

> **Graphify**: after adding the recovery function, run `graphify update . --force`.

Extend `service/vc/snapshot_test.go` with a recovery case: seed PENDING/IN_PROGRESS/FAILED rows + stale clones, start the VC, assert stale clones are dropped, rows re-enqueued, and eventually COMPLETED; CHECKPOINTED rows untouched. `require.Eventually`, no `time.Sleep`. `make test-all-db`.

## Verification

```bash
make test-all-db
make lint
```

## Done criteria

- On restart, in-flight/failed snapshots re-clone and recompute; stale clones dropped first; CHECKPOINTED skipped.
- Recovery test green (no `time.Sleep`).

---

cendhu:
- New feature
- Test update
- Documentation update

- Move snapshot hashing out of the commit path entirely: the committer only makes the `_snapshot` record durable, and a dedicated scheduler goroutine claims and starts every hash job by reading durable state. One path now covers a fresh snapshot, a resubmitted one, and one orphaned by a dead worker.
- Guard the scheduler with a database-clock lease carrying a per-acquisition ownership token, so several VCs may run the same loop without an elected leader and exactly one wins each job.
- Fence lease renewal, release, state transitions, and completion on that token, so a stale worker cannot mutate successor state or publish a hash after its job was taken over.
- Run hashing and lease renewal as one errgroup: a lost lease stops the hash, and a finished hash stops the renewal.
- Preserve deterministic snapshot clones across retries and duplicate races; a clone is never dropped, since a peer VC may reuse or commit it.
- Add lease-ownership, takeover, duplicate-race, clone-preservation, and corrupt-record coverage. Document the scheduling model and the new `snapshot-hash-lease-ttl` setting.

- The scheduler polls once per lease TTL, so hashing begins within one TTL of the snapshot committing. This is a deliberate trade: a single start path, with no commit-path coupling and no coordinator involvement, in exchange for a bounded delay.
- A snapshot record that is committed but carries no `clone_database` is unrecoverable — a committed txID must always have a matching clone, and re-cloning an already-committed txID is not safe. Such a record is marked FAILED with a diagnostic message and a halt is signalled, rather than being retried silently.

- resolves #663


cendhu:
Superseded by #803, which addresses the same issue (#663) with a different design.

Rather than fencing concurrent schedulers with a database-clock lease, #803 moves snapshot hashing into a dedicated single-instance service (`service/snapshothasher`) that polls the durable `_snapshot` record and hashes inline. With exactly one scheduler and no concurrency to arbitrate, the lease, its per-acquisition ownership token, the fenced state transitions, the renewal loop, and the `snapshot-hash-lease-ttl` setting are all unnecessary and do not appear there.

Closing this in favour of that approach.

PR #783, opened 2026-08-26
[vc] Report queue sizes on scrape

### LINKED ISSUE #770: [vc] Report queue sizes on scrape instead of sampling them
Part of #730, split out of #733 so each service can be reviewed on its own.

### The monitor

`(*ValidatorCommitterService).monitorQueues` (`service/vc/validator_committer_service.go:173`) wakes
every 250ms — hardcoded, behind a `// TODO: make sampling time configurable` — to publish four
`len(ch)` values that are available on demand: the preparer, validator, committer and txstatus
queues. The goroutine and the ticker buy nothing, and the reading is up to one interval stale.

Converting them means the value is exact at scrape time, and the closure is attached at
registration, so there is no separate wiring step that can be lost. That is not hypothetical: the
sidecar's equivalent sampler was orphaned for three months (#733).

### The queue with no gauge

| Channel | Suggested series |
| --- | --- |
| `service/vc/validator_committer_service.go:81` `receivedTxBatch` | `vcservice_batcher_input_queue_size` |

This is the one that matters most in the whole epic: `receivedTxBatch` is the first queue the VC
service fills from the coordinator's stream, so coordinator-side over-drive shows up there before
any later stage.

### Also in scope

The service's pipeline channels are currently locals in `NewValidatorCommitterService`
(`validator_committer_service.go:81` onward) and fields on the service. A gauge has to be registered
against a channel that already exists, so they should be gathered into a `queues` struct built
before `newVCServiceMetrics`, which is then passed in. Registering a gauge against a nil channel is
otherwise easy to do by accident, and it reports a silent zero.

### Constraints

- None of these queues is recreated during the service's lifetime, so each gauge can close over its
  channel directly — no indirection is needed here, unlike the sidecar in #733.

### Metric-surface impact

No rename, and no buffer size changes. The four converted gauges become exact at scrape time rather
than up to 250ms stale, so brief spikes the sampler missed will start showing.

liran-funaro:
#### Type of change

- Bug fix

#### Description

- Replace the validator-committer's ticker-based queue monitor with gauges evaluated on scrape, and
  delete the sampler goroutine.
- Report each of them with `Provider.NewChannelLenGauge`.
- Gather the service's pipeline channels into a `queues` struct built before the metrics that
  observe them.
- Add `vcservice_batcher_input_queue_size` for `receivedTxBatch`, the first queue the service fills
  from the coordinator's stream, so coordinator-side over-drive shows up where it starts rather
  than one stage later.

No metric is renamed and no buffer size changes. The four converted gauges are now exact at scrape
time rather than up to 250ms stale, so brief spikes the sampler used to miss will start showing.

#### Related issues

- resolves #770
- related to #730


service/vc/metrics_test.go
```diff
@@ -0,0 +1,74 @@
...
+		preparedTxs:     make(chan *preparedTransactions, 4),
+		validatedTxs:    make(chan *validatedTransactions, 4),
+		txsStatus:       make(chan *committerpb.TxStatusBatch, 4),
+	}
+	m := newVCServiceMetrics(q)
+
+	for _, tc := range []struct {
+		name  string
+		fill  func()
+		gauge func() int
```
dean-amar: Shouldn't that be a function that receives a `t *testing.T` parameter?

Alternatively, you can define the gauge as a `prometheus.Metric` and use:

```go
require.Equal(t, 0, test.GetIntMetricValue(t, tc.gauge))
```

.bob/skills/development/references/service-construction.md
```diff
@@ -242,29 +242,60 @@ incrementally rather than derivable on demand (e.g. a count of in-flight subscri
...
+func newVCServiceMetrics(q *queues) *perfMetrics {
 	p := monitoring.NewProvider()
 	return &perfMetrics{
 		Provider: p,
 		transactionReceivedTotal: p.NewCounter(prometheus.CounterOpts{
-			Namespace: "vcservice", Subsystem: "grpc",
+			Namespace: namespace, Subsystem: subsystemGRPC,
 			Name: "received_transaction_total", Help: "Number of transactions received.",
 		}),
+		preparerInputQueueSize: monitoring.NewChannelLenGauge(p, prometheus.GaugeOpts{
```
dean-amar: `NewChannelLenGauge` is a `*Provider`'s method, not a standalone function:
```
		preparerInputQueueSize: p.NewChannelLenGauge(prometheus.GaugeOpts{
			Namespace: namespace, Subsystem: subsystemPreparer,
			Name: nameInputQueueSize, Help: "The preparer input queue size",
		}, q.toPrepareTxs),
```

.bob/skills/development/references/service-construction.md
```diff
@@ -242,29 +242,60 @@ incrementally rather than derivable on demand (e.g. a count of in-flight subscri
...
 ```
 
+**Never sample a value on a ticker if it can be read on demand.** A goroutine that wakes up to
+call `promutil.SetGauge(len(ch))` costs a goroutine and makes the value up to one interval
+stale, and the wiring is easy to lose — the sidecar shipped one for three months after its only
+call site was deleted, so two gauges silently read zero. Instead:
+
+| Reporting | Use |
+|---|---|
+| Length of a channel that lives as long as the service | `monitoring.NewChannelLenGauge(p, opts, ch)` |
```
dean-amar: Should be `p.NewChannelLenGauge(opts, ch)`

.bob/skills/development/references/service-construction.md
```diff
@@ -242,29 +242,60 @@ incrementally rather than derivable on demand (e.g. a count of in-flight subscri
...
+| Anything else already computable (a map size, a struct field) | `p.NewGaugeFunc(opts, fn)` — `fn` runs on the scrape path, so keep it cheap, non-blocking and nil-safe |
```
dean-amar: We should also add `p.NewAtomicChannelLenGauge(opts, ch)`

dean-amar (approved):
LGTM

cendhu (approved):

PR #802, opened 2026-08-31
[sidecar] add blockhash and prevblockhash to allTxStream block event

### LINKED ISSUE #773: [sidecar] Add block_hash and prev_block_hash to StreamAllTransactions response
# Context

StreamAllTransactions streams committed transactions to clients as committerpb.TxEventBatch, which today only carries block_number and events (defined in the external fabric-x-common module's api/committerpb/notify.proto).

If StreamAllTransactions includes these two fields, it's output is equivalent to the Delivery endpoint. This allows clients to catch up via Delivery and then switch to the cheaper Notification endpoint. For example, in fabric-x-evm we need the hashes to be compatible with the ethereum API. Rather than mocking them, we want to have the real Fabric-X representation. It's also likely that the fabric-x-block-explorer can benefit from this change, which would lighten the load on the committer.

# Proposal

Add populated block_hash and prev_block_hash (bytes) to TxEventBatch.

# Data availability / cost

`prev_block_hash` is free. It's already in memory as blkWithStatus.block.Header.PreviousHash by the time committedBlockWithTxs is built at relay.go (processCommittedBlocksInOrder).

`block_hash` is currently not computed. It's one call to protoutil.BlockHeaderHash(header) (SHA-256 over the header's Number/PreviousHash/DataHash, ~100 bytes) per block — negligible next to the existing per-tx work in filterAndBuildEvents (namespace filtering, endorsements, read/write sets).

Note: fabric-x-common's utils/deliverorderer/verify.go already computes this same hash once per block for chain verification, but only keeps it in a private field (lastBlockHeaderHash), not attached to the block, and that verification path is skippable under some fault-tolerance configs — so it can't be relied on as already-available upstream. The sidecar would need to recompute it itself.

# Scope

This spans two repos:

*fabric-x-common*: add `bytes block_hash = 3;` / `bytes prev_block_hash = 4;` to TxEventBatch in api/committerpb/notify.proto, regenerate notify.pb.go.

*This repo*, after bumping go.mod to the new fabric-x-common version:
- committedBlockWithTxs (service/sidecar/notify.go)— add blockHash / prevBlockHash []byte fields.
- processCommittedBlocksInOrder (service/sidecar/relay.go) — compute protoutil.BlockHeaderHash(blkWithStatus.block.Header) and read Header.PreviousHash.
- streamWorker — copy the new fields into TxEventBatch.
- Update test fixtures.

arner:
#### Type of change

- New feature
 
#### Description

- upgrade fabric-x-common to latest `main`
- rename TxEventBatch to BlockEvent, as updated in fabric-x-common
- add blockHash and prevBlockHash to all BlockEvents. 

Impact on code: users of `Notifier_StreamAllTransactionsClient` that upgrade their committer dependency must rename TxEventBatch to BlockEvent.

~GRPC API and protos remain backwards compatible.~. This change is *not* backwards compatible on the wire. Clients of the upcoming fabric-x-committer version must also upgrade their fabric-x-common dependency; Notifier_StreamAllTransactionsClient itself is renamed to Notifier_StreamBlocksClient in fabric-x-common, with a new `StreamBlocks` GRPC endpoint replacing StreamAllTransactions.

#### Related issues

Closes https://github.com/hyperledger/fabric-x-committer/issues/773

### DIFF
--- a/service/sidecar/relay.go
+++ b/service/sidecar/relay.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/hyperledger/fabric-protos-go-apiv2/common"
 	"github.com/hyperledger/fabric-x-common/api/committerpb"
+	"github.com/hyperledger/fabric-x-common/protoutil"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/hyperledger/fabric-x-committer/api/servicepb"
@@ -414,9 +415,11 @@ func (r *relay) processCommittedBlocksInOrder(
 
 		// Create committedBlockWithTxs from blockWithStatus for notifier
 		outgoingCommittedBlockWithTxs.Write(&committedBlockWithTxs{
-			blockNumber: blkWithStatus.blockNumber,
-			txs:         blkWithStatus.txs,
-			statuses:    blkWithStatus.txStatus,
+			blockNumber:   blkWithStatus.blockNumber,
+			txs:           blkWithStatus.txs,
+			statuses:      blkWithStatus.txStatus,
+			blockHash:     protoutil.BlockHeaderHash(blkWithStatus.block.Header),
+			prevBlockHash: blkWithStatus.block.Header.PreviousHash,
 		})
 	}
 }
--- a/docs/notification-service.md
+++ b/docs/notification-service.md
@@ -197,10 +197,14 @@ message StreamAllRequest {
     repeated Status filter_status = 2;      // Optional: filter by status
     bool include_read_write_sets = 3;       // Optional: include read/write sets
     bool include_endorsements = 4;          // Optional: include endorsements
+    bool include_metadata = 5;              // Optional: include transaction metadata
 }
 
-message TxBatch {
-    repeated TxEvent events = 1;
+message BlockEvent {
+    uint64 block_number = 1;                // The block these events originated from
+    repeated TxEvent events = 2;
+    bytes block_hash = 3;                   // The hash of this block
+    bytes prev_block_hash = 4;              // The hash of the previous block
 }
 
 message TxEvent {
@@ -208,6 +212,7 @@ message TxEvent {
     Status status = 2;                      // Transaction status
     repeated TxNamespace namespaces = 3;    // Namespaces (if filtering enabled)
     repeated Endorsement endorsements = 4;  // Endorsements (if requested)
+    repeated bytes metadata = 5;            // Transaction metadata (if requested)
 }
 ```
 
@@ -240,10 +245,14 @@ stream, err := client.StreamAllTransactions(ctx, &committerpb.StreamAllRequest{
 - **`include_endorsements`** (optional): If true, the `endorsements` field in each `TxEvent`
   includes the transaction endorsements. Default: false.
 
+- **`include_metadata`** (optional): If true, the `metadata` field in each `TxEvent`
+  includes the transaction metadata. Default: false.
+
 ### 3.3. Receiving Transaction Events
 
-The server sends `TxBatch` messages containing one or more `TxEvent` entries. Transactions are batched by block. All
-transactions from the same block are delivered in a single `TxBatch`.
+The server sends `BlockEvent` messages containing one or more `TxEvent` entries, along with the block's hash and its
+predecessor's hash. Transactions are batched by block. All transactions from the same block are delivered in a single
+`BlockEvent`.
 
 ```go
 for {
--- a/docs/notification-service.md
+++ b/docs/notification-service.md
@@ -250,16 +250,16 @@ stream, err := client.StreamAllTransactions(ctx, &committerpb.StreamAllRequest{
 
 ### 3.3. Receiving Transaction Events
 
-The server sends `BlockEvent` messages containing one or more `TxEvent` entries, along with the block's hash and its
-predecessor's hash. Transactions are batched by block. All transactions from the same block are delivered in a single
-`BlockEvent`.
+The server sends one `BlockEvent` per committed block, containing zero or more `TxEvent` entries (zero if none matched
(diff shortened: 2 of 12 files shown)


cendhu (approved):
I think, we should also rename the AllTxStream to BlockStream or something similar? What do you think? @arner

arner:
Good point. But there's a small inaccuracy there that I didn't think of before: the AllTxStream endpoint can be filtered on namespace. That means that the blocks we push are not necessarily complete, real blocks. The focus is still on the transactions, so maybe the name change would obscure that fact.

It also raises the question if we should send "empty" blocks. Some clients might prefer a full uninterrupted chain (incremental block numbers and hashes with no gaps) -- I think we'd prefer that for fabric-x-evm. While others would prefer not to receive those blocks at all.

Maybe we should add an `include_empty_blocks` (or `skip_empty_blocks` - a breaking change) as a config flag when opening the stream, analogous to the other flags. That would make the name change more logical too. 

Relevant piece of code in the streamWorker:

```go
		// Filter and build transactions in this worker thread
		filteredEvents := s.filterAndBuildEvents(block)
		// Only send if there are events after filtering
		if len(filteredEvents) == 0 {
			continue
		}

```

cendhu:
> Good point. But there's a small inaccuracy there that I didn't think of before: the AllTxStream endpoint can be filtered on namespace. That means that the blocks we push are not necessarily complete, real blocks. The focus is still on the transactions, so maybe the name change would obscure that fact.
> 
> It also raises the question if we should send "empty" blocks. Some clients might prefer a full uninterrupted chain (incremental block numbers and hashes with no gaps) -- I think we'd prefer that for fabric-x-evm. While others would prefer not to receive those blocks at all.
> 
> Maybe we should add an `include_empty_blocks` (or `skip_empty_blocks` - a breaking change) as a config flag when opening the stream, analogous to the other flags. That would make the name change more logical too.
> 
> Relevant piece of code in the streamWorker:
> 
> ```go
> 		// Filter and build transactions in this worker thread
> 		filteredEvents := s.filterAndBuildEvents(block)
> 		// Only send if there are events after filtering
> 		if len(filteredEvents) == 0 {
> 			continue
> 		}
> ```

For the client to get blockHash and previousBlockHash, i think it is necessary to have the empty block event by default even if there is no txEvent. No need for a special flag.

arner:
It's a change in contract: clients currently only get batches with transactions. But considering how new it is, there are probably not so many users of this endpoint yet. I can make that change as part of this PR if you want.

service/sidecar/relay.go
```diff
@@ -414,9 +415,11 @@ func (r *relay) processCommittedBlocksInOrder(
 
 		// Create committedBlockWithTxs from blockWithStatus for notifier
 		outgoingCommittedBlockWithTxs.Write(&committedBlockWithTxs{
-			blockNumber: blkWithStatus.blockNumber,
-			txs:         blkWithStatus.txs,
-			statuses:    blkWithStatus.txStatus,
+			blockNumber:   blkWithStatus.blockNumber,
+			txs:           blkWithStatus.txs,
+			statuses:      blkWithStatus.txStatus,
+			blockHash:     protoutil.BlockHeaderHash(blkWithStatus.block.Header),
```
liran-funaro, line 421: Is it the block hash or the header hash?
The names implies it is the current block hash, but we put the header hash.

The comment from the protobuf says:

> The hash of the block from which these transactions originated.

arner: Oh...! That's confusing...

It seems that in the committer blockVerificationStateMachine, you're making sure that the`block.Header.PreviousHash` is the same as `protoutil.BlockHeaderHash(blk.Block.Header)`. So it's all about the header....

Is there a even a blockhash? There is `block.Header.DataHash`, but if we would use that, it would mean that the PrevBlockHash does not match the BlockHash of the previous block, that can't be right either. Does this mean the name in fabric-x-common is wrong? @cendhu do you have an opinion about this?

cendhu: Block header hash is considered as the block hash. There is no separate hash of the whole block. The chain is formed on the block headers only.

cendhu: We can also simply return the header to avoid any confusion as this.

liran-funaro: The block have two hashes: data-hash (block.Header.DataHash) and header-hash (block.Header.PreviousHash).
We just have to be explicit which one we emit.

arner: @liran-funaro and @cendhu, please let me know how to proceed. We can go a couple of ways:

1. Keep it like this.
2. Rename BlockHash to HeaderHash and PreviousBlockHash to PreviousHash in fabric-x-common's committerpb.BlockEvent.
3. Include the common.BlockHeader in committerpb.BlockEvent and remove BlockNumber, BlockHash, PreviousBlockHash from the top level. The BlockHeader includes DataHash, so a client can calculate the blockhash with protoutil.BlockHeaderHash.

The last option is how I understand @cendhu s suggestion and it seems the cleanest of the three. The header only contains three fields (num, prevHash, dataHash) so we're sending the same amount of data over the wire. We return a standard object that already existed in Fabric, so protoutil and other tooling works with it.

```
// BlockEvent contains a batch of transaction events from a single block.
message BlockEvent {
    reserved 1;

    // List of transaction events (may be filtered based on request).
    repeated TxEvent events = 2;

    // The header of the block from which these transactions originated.
    common.BlockHeader header = 3;
}
```

Shall I make a PR in fabric-x-common? If yes, do you prefer me to keep the original blocknum field (1) reserved or replace it with the header? (AFAIK we're already making breaking changes in the API this round, so might as well put the header on number 1?).

cendhu: Yes, the approach 3.

arner:
@liran-funaro @cendhu ready for review

cendhu (approved):

PR #803, opened 2026-08-31
[snapshot-hasher, vc] Hash snapshots in a dedicated service

### LINKED ISSUE #663: I-13: [vc] Snapshot crash recovery

cendhu:
#### Type of change

- New feature
- Test update
- Documentation update

#### Description

- Add `service/snapshothasher`, a new single-instance service that owns snapshot hashing end to end: it polls the latest durable `_snapshot` record every `poll-interval` and hashes the record's clone inline, so one path covers a fresh snapshot, a resubmitted one, and one orphaned mid-hash by a restart. Start it with `committer start snapshot-hasher` (gRPC 3001, monitoring 2120); it serves only health checks, since work arrives through the state database, so `WaitForReady` is unconditionally true -- the readiness handshake exists to hold back a gRPC server that would otherwise answer requests before it can. It is named for hashing rather than for snapshots because creating a snapshot and its clone stays in the VC.
- Move the hasher (table discovery, keyset-paginated scans, deterministic per-table and root digests) out of `service/vc` into that service, and remove snapshot hashing from the VC entirely. The VC's only snapshot duty is now to make the record durable atomically with its clone.
- Because exactly one scheduler exists and hashing runs inline, no cross-process exclusion is introduced at all: no lease, no ownership token, no fenced state transitions, no leader election.
- Extract the durable `_snapshot` record contract into `utils/statedb/snapshot_state.go` as a `SnapshotStateManager` (latest-record pointer, encoding, `FOR UPDATE` read/update) and cover it with unit tests. It lives in `statedb`, next to the schema it reads, rather than in a package of its own; `utils/statedb` therefore joins `REQUIRES_DB_PACKAGES` in the Makefile, since its tests now need a database but not a particular one. Both processes share it: the VC reads it to gate a new snapshot request and will write CHECKPOINTED through it when checkpointing lands, while the snapshot service drives PENDING -> IN_PROGRESS -> COMPLETED. `Update` requires a status, writes the error text unconditionally so a record cannot report both COMPLETED and a previous attempt's failure, and treats a missing or undecodable row as terminal instead of spending the retry budget on it; `retry.Execute` takes terminal errors for that, matching `ExecuteWithResult`.
- Encode `tx_status` as `int4send(coalesce(status, -1)) || height`, because `status` is nullable and `NULL || height` is `NULL`: without the coalesce a single NULL status drops that row's height from the digest, so two rows differing only in height hash identically.
- Open the clone pool before the record is marked IN_PROGRESS, so a committed record whose clone is not there to hash is left exactly as it was found. A clone is created before its snapshot transaction commits, so a committed record always names a clone that exists; neither a missing `clone_database` nor a missing database can be produced by this system, so both are reported as `ErrCorruptSnapshotState`, stop the service, and leave the record untouched as evidence.
- Make `statedb.NewPool` treat SQLSTATE 3D000 (`invalid_catalog_name`) as terminal instead of retrying it, export `ErrDatabaseNotFound` so a caller can tell an absent database from a connection failure, and close a pool whose ping failed rather than leaking one per attempt. The database is provisioned before any service starts, so 3D000 means the configured name is wrong or the cluster was never initialized; retrying spent the whole retry budget before reporting a condition that was decidable on the first attempt.
- Replace the VC's `max-workers-for-snapshot-hash` and `snapshot-hash-batch-size` with the new service's `poll-interval`, `resource-limits.max-workers-for-hash`, and `resource-limits.hash-batch-size`; add `cmd/config/samples/snapshot-hasher.yaml` and its test template, under the `SC_SNAPSHOT_HASHER_` env prefix.
- Export snapshot/state test fixtures from `service/vc` so the new service's tests seed records through the real commit path, and add `docs/snapshot-hasher.md` plus `snapshothasher_*` metrics for completed/failed jobs, hash duration, poll errors, when the running hash started, and the shared gRPC server metrics every other service reports.

- Running a second instance of the snapshot hasher is a deployment error rather than a correctness hazard: a clone is immutable and the digest deterministic, so the symptom is duplicate work writing the same digest. This is the trade that keeps the design free of exclusion machinery.
- A hash failure stays retryable and is recorded on the record, because the record is the only queue: nothing re-enqueues it, so a terminal status would strand the snapshot on one transient failure, and the VC would then reject every later snapshot. Retrying is cheap and safe, since a clone is immutable and the digest deterministic.
- `snapshothasher_poll_errors_total` counts ticks that could not determine whether there is work (state database unreachable, a pointer naming no row, a record that does not decode). It is incremented where the failure is classified rather than in the poll loop, since only there is a failed poll distinguishable from a hash that started and then failed; the hash-job counters cannot express the former, so without it a service whose database is down looks like an idle one.
- `snapshothasher_hash_started_timestamp_seconds` describes a hash while it runs, which nothing else here can: `hash_duration_seconds` is observed only once a hash returns, so for the many minutes a full clone scan takes, a busy service and an idle one publish identical numbers, and this service has no RPC or queue depth to read instead. Alert on the elapsed time rather than on the value being non-zero, since a stuck job holds it forever and never reaches the histogram.
- Each hash job opens its own pool on the clone with exactly `max-workers-for-hash` connections, since at most one connection per worker is in use while it pages through its table; `database.max-connections` sizes only the record poll, whose sample default drops to 2.
- Hashing latency is bounded by one `poll-interval` (default 1m): nothing notifies the service, so that interval is both the scheduling delay and the restart resume delay.
- Planned per-table digests and a Merkle Patricia trie for localizing divergence between organizations land in this service; both are finer-grained hashing, so they fit the name rather than outgrowing it.
- Upgrading a deployment that already took state snapshots: deploy the hasher as a new unit and delete the two removed keys from `vc.yaml`, which are ignored rather than rejected if left behind. In-flight records need no migration — the new service picks up whatever state the old VC left, including a record left IN_PROGRESS by the upgrade itself.
- Left as a TODO on `hashSnapshot`: no hash failure is classified as permanent. The hasher's per-page and table-discovery retries pass no terminal errors, so a permanent failure (a permission change, a dropped table) would be retried for the whole budget and then again on the next tick. No operational path reaches that today, so it is recorded rather than fixed here.

#### Related issues

- resolves #663


service/snapshothasher/snapshot_hasher_service.go
```diff
@@ -0,0 +1,89 @@
...
+
+	scheduler := newScheduler(&schedulerConfig{
+		state:        snapshotstate.NewStateManager(pool, s.config.Database.Retry),
+		hasher:       newHasher(s.config),
+		metrics:      s.metrics,
+		pollInterval: s.config.PollInterval,
+	})
+
+	s.ready.SignalReady()
+	defer s.ready.Reset()
```
liran-funaro: There is no need for a ready flag.
Ready flag is used to ensure the GRPC server does not start serving before the service's resources are ready to answer requests. But this service doesn't have requests to answer, so we don't need this flag.

service/snapshothasher/scheduler.go
```diff
@@ -0,0 +1,216 @@
...
+// snapshot transaction commits, so a committed record always names a clone that
+// exists; neither a missing name nor a missing database can arise from anything this
+// system does, which leaves external interference or storage corruption. No retry
+// repairs either, so the service stops instead of logging the same impossible state
+// once per interval forever.
+var ErrCorruptSnapshotState = errors.New("corrupt durable snapshot state")
+
+// schedulerConfig carries the scheduler's collaborators, which exceed the
+// four-argument limit as positional parameters.
+type schedulerConfig struct {
```
liran-funaro: nit: We use config for user input (YAML) and parameters for programmatic input.
```suggestion
type schedulerParameters struct {
```

utils/statedb/dbinit.go
```diff
@@ -70,8 +78,26 @@ func NewPool(ctx context.Context, config *Config) (*pgxpool.Pool, error) {
...
+		return p, terminalIfMissingDatabase(config.Database,
+			errors.Wrap(p.Ping(ctx), "failed to create a connection pool"))
+	}, retry.ErrNonRetryable)
+}
+
+// ErrDatabaseNotFound reports that a pool was opened against a database that does
+// not exist (SQLSTATE 3D000, invalid_catalog_name). It is exposed so a caller whose
+// database is guaranteed to exist can tell this apart from a connection failure and
+// treat it as the invariant violation it is, rather than retrying it.
+var ErrDatabaseNotFound = errors.New("database does not exist")
```
liran-funaro: nit: Should be at the top of the file.

utils/statedb/dbinit.go
```diff
@@ -70,8 +78,26 @@ func NewPool(ctx context.Context, config *Config) (*pgxpool.Pool, error) {
...
+// terminalIfMissingDatabase marks SQLSTATE 3D000 (invalid_catalog_name)
+// non-retryable: the named database is absent, and no number of attempts brings it
+// into existence.
+func terminalIfMissingDatabase(database string, err error) error {
+	var pgErr *pgconn.PgError
+	if errors.As(err, &pgErr) && pgErr.Code == pgerrcode.InvalidCatalogName {
+		return errors.Wrapf(errors.Join(retry.ErrNonRetryable, ErrDatabaseNotFound, err), "database %s", database)
+	}
+	return err
 }
```
liran-funaro: nit: single-use tiny methods are discouraged.
minor: We don't need to wrap the original error. `ErrDatabaseNotFound` is sufficiently descriptive.

utils/statedb/dbinit.go
```diff
@@ -70,8 +78,26 @@ func NewPool(ctx context.Context, config *Config) (*pgxpool.Pool, error) {
 		}
 		// NewWithConfig creates the pool lazily without connecting, so we ping to
 		// verify connectivity eagerly and let the retry loop handle transient failures.
-		return p, errors.Wrap(p.Ping(ctx), "failed to create a connection pool")
-	})
+		return p, terminalIfMissingDatabase(config.Database,
+			errors.Wrap(p.Ping(ctx), "failed to create a connection pool"))
```
liran-funaro: nit: We don't need to wrap the ping error here. We can wrap it after the `terminalIfMissingDatabase` check.

service/snapshothasher/main_test.go
```diff
@@ -0,0 +1,85 @@
...
+// validator-committer commits into, which is the whole deployment this service
+// assumes: the VC makes records durable, and this process discovers them.
+type testEnv struct {
+	dbEnv     *vc.DatabaseTestEnv
+	config    *Config
+	metrics   *perfMetrics
+	hasher    *hasher
+	state     *snapshotstate.StateManager
+	scheduler *scheduler
+}
```
liran-funaro: nit: We usually don't put test helpers in `main_test.go`. We put them in the relevant test file.

utils/snapshotstate/state.go
```diff
@@ -0,0 +1,211 @@
...
+		return nil, errors.Wrap(err, "failed to marshal _snapshot record")
+	}
+	return raw, nil
+}
+
+// TableName returns the `_snapshot` namespace table name, so callers do not have to
+// rebuild it from the namespace ID.
+func TableName() string {
+	return statedb.TableName(committerpb.SnapshotNamespaceID)
+}
```
liran-funaro: nit: single-use tiny methods are discouraged.

utils/snapshotstate/state.go
```diff
@@ -0,0 +1,211 @@
...
+// a record safely.
+//
+// It is shared rather than owned by one service because two components read and
+// write the same record from different processes: the validator-committer accepts
+// or rejects a new snapshot from it (and will mark it CHECKPOINTED once
+// checkpointing lands), while the snapshot service drives it from PENDING to
+// COMPLETED as it hashes the clone. Keeping the pointer key, the encoding, and the
+// locking discipline in one place is what keeps those two processes agreeing on the
+// record.
+package snapshotstate
```
liran-funaro: minor: Do we need a new package for this? Can we put this in the `statedb` package?

cendhu: Moved to statedb package.

cmd/config/create_config_file.go
```diff
@@ -72,6 +72,7 @@ type (
 		VCMinTransactionBatchSize           int           // vc
 		VCTimeoutForMinTransactionBatchSize time.Duration // vc
 
+		// Snapshot hasher configuration (for testing).
```
liran-funaro: leftover comment

docs/validator-committer.md
```diff
@@ -89,8 +89,7 @@ world state for each namespace, and internal metadata. The tables are categorize
 | `value`     | `BYTEA`   | The system configuration transaction.              |
 | `version`   | `BIGINT`  | Not Applicable                                     |
 
-* **Metadata Table (`metadata`)**: This table is a simple key-value store for other internal system metadata. For now, we
-only store the last committed block number.
+* **Metadata Table (`metadata`)**: This table is a simple key-value store for internal system metadata. It stores the last committed block number and the key of the latest accepted snapshot record, which lets a reader find that record with a single key lookup instead of scanning `ns__snapshot`.
```
liran-funaro: I would add here that it is not part of the world-state. But helper maintenance data.

service/snapshothasher/config.go
```diff
@@ -0,0 +1,37 @@
...
+type Config struct {
+	Database *statedb.Config `mapstructure:"database" validate:"required"`
+
+	// PollInterval is how often the service re-reads the latest `_snapshot` record
+	// looking for work. Hashing therefore begins within one interval of the snapshot
+	// committing; this is the price of a scheduler that is driven purely by durable
+	// state, with no notification from the validator-committer or the coordinator.
+	PollInterval time.Duration `mapstructure:"poll-interval" default:"1m" validate:"gt=0"`
+
+	ResourceLimits *ResourceLimitsConfig `mapstructure:"resource-limits" validate:"required"`
```
liran-funaro: nit: this can be non pointer

docs/snapshot-hasher.md
```diff
@@ -0,0 +1,185 @@
...
+
+- **A committed record whose clone is not there to hash.** The clone is created before
+  its snapshot transaction commits, so a committed record always names a clone that
+  exists. Neither shape of absence — no `clone_database` recorded, or a recorded name
+  whose database is gone — can be produced by this system, which leaves external
+  interference or storage corruption. No retry repairs either, and recording a failed
+  attempt on the record would invite a later tick to treat it as retryable work, so the
+  service stops and leaves the record as evidence.
+- **A record with no `TxRef`** is a hard error, because a record that cannot be named
+  cannot be driven; treating it as "nothing to do" would stall hashing silently.
```
liran-funaro, line 89: Major: Corruption can result from a deterministic bug. In such a case, all of the organization's committers will not be able to process new checkpoints because the previous one is incomplete.
If this is a deterministic bug, it will cause issues when fixing it in the future, preventing future re-ingestion.
We should also address how to fix this when a clone is lost forever, but we have a checkpoint in past blocks without the proper clone or checkpoint.
In such a case, we have no way to produce the checkpoint in the present, but re-ingestion might be able to---producing different results.

Adding an `ABORTED` state will allow a new checkpoint if this state is consistent across all organizations, without having to manually patch the database later on to overcome a stale `FAILED` entry. 
To solve the re-ingestion issue, the admins will also have to submit an abort checkpoint TX that will commit the aborted state for good. This ensures reproducible results even in a new committer version that will not have this bug.

docs/snapshot-hasher.md
```diff
@@ -0,0 +1,185 @@
...
+## 5. Hash Computation
+
+The digest is a SHA-256 over the clone's committed content, computed so that identical
+clone content always yields an identical digest:
+
+- The hashed set is derived from `ns__meta`, the authoritative namespace registry:
+  every user namespace's `ns_<id>` table, plus `ns__meta`, `ns__config`, and
+  `tx_status`. `metadata`, `ns__snapshot`, and `ns__checkpoint` are excluded — the
+  first two are exactly the tables this service and the VC write while a snapshot is in
+  flight, so including them would make the digest depend on hashing progress.
```
liran-funaro: This touches on my earlier comment.
If `metadata`, `ns__snapshot`, and `ns__checkpoint` are excluded, then re-ingestion is not an issue.
But this means that the checkpointing process is part of the ledger, but not the world state.
So why do we need this to be part of the ledger at all?

docs/snapshot-hasher.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Observability
+
+`snapshothasher_poll_errors_total` is the metric to alert on: it counts ticks that
+could not determine whether there is work at all (state database unreachable, a
+pointer that names no row, a record that does not decode). The hash-job counters
+cannot express those — such a tick completes no job and fails none — so without it a
+service whose database is down looks the same as an idle one.
+
+`snapshothasher_hash_started_timestamp_seconds` covers the opposite blind spot.
```
liran-funaro: The timestamp is recorded as part of the observation.
So we can count the total number of hash start processes, which will also give us the point in time.
A divergence between `snapshothasher_hash_started_count` and `snapshothasher_hash_duration_seconds_count` is an indicator of an error.

service/snapshothasher/hasher.go
```diff
@@ -0,0 +1,233 @@
...
+	"github.com/yugabyte/pgx/v5"
+	"github.com/yugabyte/pgx/v5/pgxpool"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/hyperledger/fabric-x-committer/utils/retry"
+	"github.com/hyperledger/fabric-x-committer/utils/statedb"
+)
+
+// txStatusPageSQL pages tx_status in primary-key order for hashing, encoding the
+// hashed value in SQL as int32BE(status)||height so every table yields the same
```
liran-funaro:
```suggestion
// hashed value in SQL as int4send(status)||height so every table yields the same
```

service/snapshothasher/hasher.go
```diff
@@ -0,0 +1,233 @@
...
+	"WHERE tx_id > $1 ORDER BY tx_id LIMIT $2"
+
+// hasher computes the deterministic content hash of a snapshot clone database.
+type hasher struct {
+	config *Config
+}
+
+func newHasher(config *Config) *hasher {
+	return &hasher{config: config}
+}
```
liran-funaro: nit: redundent

service/snapshothasher/hasher.go
```diff
@@ -0,0 +1,233 @@
...
+
+// hashSnapshotDatabase opens a short-lived pool on the clone database, hashes
```
liran-funaro: The comment contradicts itself.
```suggestion
// hashSnapshotDatabase is given a short-lived pool on the clone database, hashes
```

service/snapshothasher/hasher.go
```diff
@@ -0,0 +1,233 @@
...
+// listHashedTables returns the sorted list of table names to hash on the clone.
+// It reads the namespace registry from ns__meta (one key per user namespace),
+// then appends the fixed system tables ns__meta, ns__config, and tx_status.
+// ns__snapshot and ns__checkpoint are never in ns__meta, so they are naturally
+// excluded.
+func (h *hasher) listHashedTables(ctx context.Context, pool *pgxpool.Pool) ([]string, error) {
+	// metaTable is a sanitized fixed identifier, not user input.
+	metaTable := pgx.Identifier{statedb.TableName(committerpb.MetaNamespaceID)}.Sanitize()
+	metaRows, err := retry.ExecuteWithResult(ctx, h.config.Database.Retry, func() ([]struct{ Key []byte }, error) {
+		rows, queryErr := pool.Query(ctx, fmt.Sprintf("SELECT key FROM %s", metaTable))
```
liran-funaro: Adding `ORDER BY` to the query will allow removing the sorting below.
This will also allow a more natural order: config, meta, status, then the rest of the table in sorted order.

service/snapshothasher/hasher.go
```diff
@@ -0,0 +1,233 @@
...
+		statedb.TxStatusTableName,
+	)
+	slices.Sort(tables)
+	return tables, nil
+}
+
+// hashTable scans one table in primary-key order in bounded pages (keyset
+// pagination) and folds rows into a per-table SHA-256 using length-prefixed
+// encoding len(key)||key||len(value)||value. tx_status is encoded as key=tx_id,
+// value=int32BE(status)||height (see txStatusPageSQL). Paging bounds worker
```
liran-funaro:
```suggestion
// value=int4send(status)||height (see txStatusPageSQL). Paging bounds worker
```

service/snapshothasher/hasher.go
```diff
@@ -0,0 +1,233 @@
...
+// NOTE (future work): fetching and hashing are sequential -- each page waits for
+// the previous hash fold and vice versa. Pipelining them (fetch page N+1 while
+// hashing page N) is deliberately not done, to avoid extra concurrent read load
+// on a cluster that is also serving live transactions.
+func (h *hasher) hashTable(ctx context.Context, pool *pgxpool.Pool, table string) ([]byte, error) {
+	// table is a sanitized identifier built from ns__meta keys, not user input.
+	sanitizedTable := pgx.Identifier{table}.Sanitize()
+	query := txStatusPageSQL
+	if table != statedb.TxStatusTableName {
+		query = fmt.Sprintf("SELECT key, value FROM %s WHERE key > $1 ORDER BY key LIMIT $2", sanitizedTable)
```
liran-funaro: nit: Why `txStatusPageSQL` is a const but this is not?

service/snapshothasher/hasher_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger/fabric-x-committer/service/vc"
+)
+
+func TestSnapshotHashDeterministic(t *testing.T) {
+	t.Parallel()
+	env := newTestEnv(t)
+	ctx, cancel := createContext(t)
+	defer cancel()
```
liran-funaro: minor: redundant - already called at cleanup

service/snapshothasher/hasher_test.go
```diff
@@ -0,0 +1,155 @@
...
+// TestSnapshotHashExcludesUnregisteredNamespaces proves that rows in the
+// `_snapshot` and `_checkpoint` system namespaces are EXCLUDED from the digest:
+// those tables are not registered in ns__meta, so listHashedTables never hashes
+// them. Excluding them is what keeps the digest stable while the hash job itself
+// writes progress into `_snapshot`.
+func TestSnapshotHashExcludesUnregisteredNamespaces(t *testing.T) {
+	t.Parallel()
+	env := newTestEnv(t)
+	ctx, cancel := createContext(t)
+	defer cancel()
```
liran-funaro: minor: redundant - already called at cleanup

service/snapshothasher/hasher_test.go
```diff
@@ -0,0 +1,155 @@
...
+// Reserving a spare connection would hide that ordering; a pool sized below the worker
+// count would deadlock here rather than merely serialize.
+//
+// The digest must also be identical to the parallel one, since worker count must not
+// affect the result.
+func TestSnapshotHashWithSingleWorker(t *testing.T) {
+	t.Parallel()
+	env := newTestEnv(t)
+	ctx, cancel := createContext(t)
+	defer cancel()
```
liran-funaro: minor: redundant - already called at cleanup

service/snapshothasher/hasher_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	ref := &committerpb.TxRef{BlockNum: 720000, TxNum: 0, TxId: "snap-hash-single-worker"}
+	parallelHash := env.createAndHashClone(ctx, t, ref)
+	require.NotEmpty(t, parallelHash)
+
+	// Bounded well below the suite's timeout: an undersized pool starves the only
+	// worker instead of returning an error, so without its own deadline this test
+	// would report the regression as a hung package rather than a failure.
+	serialCtx, serialCancel := context.WithTimeout(ctx, 90*time.Second)
+	defer serialCancel()
```
liran-funaro:
```suggestion
	t.Cleanup(serialCancel)
```

service/snapshothasher/hasher_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	env.config.ResourceLimits.MaxWorkersForHash = 1
+	serialHasher := newHasher(env.config)
+	pool, err := serialHasher.openClonePool(serialCtx, vc.SnapshotDatabaseName(ref))
+	require.NoError(t, err)
+	defer pool.Close()
```
liran-funaro:
```suggestion
	t.Cleanup(pool.Close)
```

service/snapshothasher/scheduler.go
```diff
@@ -0,0 +1,226 @@
...
+// budget.
+func (s *scheduler) hashSnapshot(
+	ctx context.Context, state *committerpb.SnapshotState, pool *pgxpool.Pool,
+) error {
+	ref := state.TxRef
+	clone := state.CloneDatabase
+	if state.Status != committerpb.SnapshotState_IN_PROGRESS {
+		if err := s.state.Update(ctx, ref, statedb.SnapshotUpdate{
+			Status: committerpb.SnapshotState_IN_PROGRESS,
+		}); err != nil {
```
liran-funaro, line 194: This is a blind write. We can have two cases where this needs to be checked before an update.
1. Bad deployment: if there are two hashers, they can override each other's work.
2. Checkpoint: if this org's committer is slow, the checkpoint TX might arrive before it finishes. In such a case, the state can already move to CHECKPOINT.

I suggest adding "expected status" to the `state.Update()` call that verifies the current status before updating.

service/snapshothasher/scheduler_test.go
```diff
@@ -0,0 +1,376 @@
...
+// runs. Every other metric here is after-the-fact -- the duration histogram is
+// observed once a hash returns -- so without it a full clone scan, which can take
+// many minutes, is indistinguishable from an idle service. It must be set before
+// hashing and cleared afterwards, so a scrape never sees a stale start time once the
+// job is done.
+func TestHashStartedTimestampGauge(t *testing.T) {
+	t.Parallel()
+	env := newTestEnv(t)
+	ctx, cancel := createContext(t)
+	defer cancel()
```
liran-funaro: minor: redundant - already called at cleanup

service/snapshothasher/scheduler_test.go
```diff
@@ -0,0 +1,376 @@
...
+	// The gauge is observed while the hash is still running, which is the whole point of
+	// having it; a hash of seeded state is short, so a completed job is an accepted
+	// outcome of this wait rather than a failure.
+	require.Eventually(t, func() bool {
+		return startedAt() > 0 || len(done) == 1
+	}, 30*time.Second, 10*time.Millisecond)
+	if v := startedAt(); v > 0 {
+		require.GreaterOrEqual(t, v, float64(beforeSeconds),
+			"a running hash must publish when it started")
+	}
```
liran-funaro: This does not indicate whether we actually test anything or whether the hash always finishes before the condition ends.

service/snapshothasher/scheduler_test.go
```diff
@@ -0,0 +1,376 @@
...
+		{name: "IN_PROGRESS", status: committerpb.SnapshotState_IN_PROGRESS, blockNum: 800301, wantHashed: true},
+		{name: "FAILED", status: committerpb.SnapshotState_FAILED, blockNum: 800302, wantHashed: true},
+		{name: "COMPLETED", status: committerpb.SnapshotState_COMPLETED, blockNum: 800303, wantHashed: false},
+		{name: "CHECKPOINTED", status: committerpb.SnapshotState_CHECKPOINTED, blockNum: 800304, wantHashed: false},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			env := newTestEnv(t)
+			ctx, cancel := createContext(t)
+			defer cancel()
```
liran-funaro: minor: redundant - already called at cleanup

utils/statedb/snapshot_state.go
```diff
@@ -0,0 +1,212 @@
...
+// SnapshotUpdate bundles the fields SnapshotStateManager.Update can change. Status is required. A
+// nil Digest keeps the record's existing Hash, because a status move that publishes
+// no new digest (IN_PROGRESS) must not wipe one.
+//
+// ErrMsg, by contrast, is written unconditionally: the diagnostic describes the
+// status it arrives with, so a record that reaches COMPLETED after a failed attempt
+// must not keep the failed attempt's error text, which would otherwise read as a
+// completed snapshot that also failed.
+type SnapshotUpdate struct {
+	Status committerpb.SnapshotState_Status
```
liran-funaro, line 71: As mentioned above, here I suggest adding the expected status before the update.

docs/snapshot-hasher.md
cendhu: - metadata — excluded, and cannot be included. last committed block number is written by each sidecar on its own last-committed-block-set-interval. It advances on local timing, so two orgs holding byte-identical committed state will still hold different values at the moment a clone is taken. Including it would make the digest diverge for a reason that has nothing to do with state.
 - ns__snapshot — It can be included.
 - ns__checkpoint — It can be included.

When bootstrapping a node with a snapshot, ns__snapshot and ns__checkpoint will not be used.

docs/snapshot-hasher.md
cendhu: Good point. I will add the aborted status and snapshot abort tx in the subsequent PR. This requires adding a new status enum to fabric-x-common and making changes here. Would be good to get this PR merged sooner than later.

docs/snapshot-hasher.md
cendhu: ns__snapshot cannot be included as we also track the deletion of clone db using that. It also have error fields, etc...

liran-funaro: I completely understand why excluding these makes sense. There is no argument there.
The question was why we need the snapshot/checkpoint to be part of the ledger.
I think we need the snapshot, as it is convenient to have the Orderer put an exact snapshot point.
And the checkpoint for accountability.
Are there other justifications?
Anyway, my comment suggests documenting the reasoning on why this is part of the ledger, rather than just asserting it.

liran-funaro (approved):

Issue #666, opened 2026-07-15
I-16: [coordinator] Hold/halt and sidecar halt flow


cendhu:
# PR-16 — committer: coordinator hold/halt feedback + sidecar halt wiring (US4b)

**Repo**: `fabric-x-committer` (this repo)
**Tasks**: T039, T044, T045, T046 (see `the feature task breakdown`)
**Depends on**: PR-15 (checkpoint verify outcome drives the hold/halt signal) and — transitively via PR-05 — the hold/halt feedback proto from PR-04.
**Blocks**: nothing hard.

> Cross-service pipeline wiring (coordinator ↔ sidecar ↔ VC). Read `docs/core-concurrency-pattern.md` and `docs/coordinator.md` first.

## Scope

Feedback loop so a pending/failed checkpoint verification pauses or halts processing. The wire transport (`CheckpointFeedback` message / halt notification) already exists from PR-04; this PR consumes it.

- **T044** — hold/halt handling in `service/coordinator/coordinator.go`: `hold` → pause processing of the checkpoint until the VC signals the local hash is ready+verified, then resume (bounded by hash-compute time); `halt` → stop processing, per `the applicable API contract` §2.
- **T045** — sidecar-side handling of the coordinator halt notification (stop submitting new TXs) via the coordinator sidecar-feedback path (`service/coordinator/coordinator.go`) + the sidecar receiver (`service/sidecar/relay.go`).
- **T046** — non-recoverable-hash-failure halt in `service/vc/snapshot.go`: transient failure → status FAILED (4) + error, retry on restart; permanent/non-recoverable failure → signal the coordinator to halt (no local status transition), mirroring the checkpoint-mismatch halt, per `the applicable API contract` retry semantics and FR-016.

> **Graphify**: run `graphify_explain` on the coordinator status-forwarding path and the sidecar receiver before wiring. After changes, run `graphify update . --force`.

## TDD (test lands in this PR)

- **T039** [P] — `service/coordinator/coordinator_test.go`: a VC `hold` pauses checkpoint processing until release; a `halt` stops processing and triggers a sidecar stop-submitting notification. `require.Eventually`, never `time.Sleep`. `make test-no-db` (coordinator tests are DB-free; escalate to `make test-all-db` only if a case needs the VC DB path).

## Verification

```bash
make test-no-db   # coordinator; escalate to test-all-db if a case touches the VC DB
make lint
```

## Done criteria

[...]

PR #804, opened 2026-09-02
[vc, coordinator, sidecar] Verify a checkpoint against the local snapshot hash before committing it

### LINKED ISSUE #665: I-15: [vc/query] Checkpoint verification and GetCheckpoint
# PR-15 — committer: VC checkpoint verify-before-commit (US4a)

**Tasks**: T038, T041, T042, T043 (see `the feature task breakdown`; T040/T047 removed — see below)
**Depends on**: PR-12 (a COMPLETED local hash to verify against) **and** PR-06 (checkpoint policy gating).
**Blocks**: PR-16 (coordinator hold/halt reacts to the checkpoint verification outcome).

> Read `docs/core-concurrency-pattern.md` — the hold path is async (wait for a background hash), so tests must use `require.Eventually`, never `time.Sleep`.

## Scope

VC-only. On `_checkpoint` commit, verify the submitted hash against the locally-computed snapshot hash; commit into the `_checkpoint` namespace on match. Emit the hold/halt signal (transport defined in PR-04); the coordinator-side *handling* is PR-16.

**`GetCheckpoint` RPC removed from scope**: checkpoint reads now go through the existing `GetRows` RPC on the `_checkpoint` namespace (no new query-service code needed) — T047 is dropped. See PR-14 (superseded/obsolete) for the query-side note.

- **T041** — create `service/vc/checkpoint.go`: on `_checkpoint` commit, extract the hash from the versioned-write value. **Validation**: decode the key via `servicepb.NewHeightFromBytes`; if the key does NOT decode as a valid TxHeight, invalidate the transaction with a dedicated MALFORMED status (mirrors the sidecar's `_meta`/`_config` form-check pattern, e.g. `checkMetaNamespace`) rather than proceeding — this check belongs at the sidecar form-validation layer (see PR-10), so by the time this VC-side code runs the key is already known-valid. Look up the local `_snapshot` namespace row for that `TxHeight` by matching `TxRef` (since `_snapshot` is keyed by tx_id, not TxHeight — maintaining an index or scanning to resolve TxHeight -> tx_id is VC's concern; document the lookup path explicitly in code comments). Verify-before-commit:
  - hash present + matches → commit (step T042);
  - hash present + mismatches → return a **halt** signal (no commit);
  - hash empty (PENDING/IN_PROGRESS) → return a **hold** signal (do not treat as mismatch, do not skip verification).
- **T042** — on a verified match, commit the versioned write into the `_checkpoint` namespace via the normal namespace-write commit path (key=`TxHeight` bytes, value=hash raw bytes, version resolves to 0), then update the `_snapshot` row to status=CHECKPOINTED via a versioned namespace write (no `checkpointed_at` — there are no timestamp fields on `SnapshotState`) (unblocking the next snapshot's hash, per PR-12's gate).
- **T043** — wire `_checkpoint` into `service/vc/validator_committer_service.go` commit path (call into `checkpoint.go` when `committerpb.CheckpointNamespaceID` is committed), returning hold/halt feedback to the coordinator rather than committing on mismatch/pending.

> **Graphify**: run `graphify_explain` on the VC commit path before wiring `checkpoint.go`. After changes, run `graphify update . --force`.

- **T038** [P] — `service/vc/checkpoint_test.go`, per `the applicable API contract` §4 and `the feature data model` §6:
  1. hash-present-match → commits into `_checkpoint` namespace + `_snapshot` row status=CHECKPOINTED (versioned namespace write, no timestamp field);
  2. hash-present-mismatch → halt signal, no commit;
  3. **hash-not-computed → hold signal, then verify once the hash lands** (use `require.Eventually`);
  4. **duplicate checkpoint (same `TxHeight`) ignored idempotently**;
  5. **read-write key does not decode as a valid TxHeight → transaction invalidated with a dedicated MALFORMED status, never reaches commit**.

  Encode/decode keys via `servicepb.Height.ToBytes`/`NewHeightFromBytes`. `make test-all-db`.

All async assertions use `require.Eventually`, never `time.Sleep`.

## Verification

```bash
make test-all-db
make lint
```

## Done criteria

- Checkpoint commits only on verified hash match; `_snapshot` row → CHECKPOINTED; mismatch → halt signal; hash-pending → hold signal then verify; duplicate ignored; invalid TxHeight key → invalidated, never committed.
- No `GetCheckpoint` RPC added — checkpoint reads verified separately via `GetRows` (PR-14).
- Test green under `make test-all-db` (no `time.Sleep`).

---

cendhu:
#### Type of change

- New feature
- Test update
- Documentation update

#### Description

- Add `service/vc/checkpoint.go`: before a `_checkpoint` write commits, the VC compares the attested hash against the hash it computed for the referenced snapshot. On a match the attestation commits and the `_snapshot` record advances to CHECKPOINTED; otherwise the write is dropped and either the TX is rejected (a snapshot this committer is not awaiting) or a `CheckpointFeedback` gates block intake — HOLD while the hash is still computing, HALT when local state contradicts it.
- Reject rather than halt on bad input. `_checkpoint` is client-submittable, so halting on a checkpoint for a block that was never snapshotted would let any authorized submitter stop the committer. Both rejections reuse MALFORMED_CHECKPOINT_INVALID_KEY with a TODO: a precise REJECTED_CHECKPOINT_NO_SUCH_SNAPSHOT has to land in fabric-x-common first.
- Advance the record inside the commit transaction (`statedb.MarkSnapshotCheckpointedInTx`). A second transaction would leave a crash window where a durable attestation reads to the admission gate as "still awaiting a checkpoint", rejecting every later snapshot. Its invariant violations are non-retryable, so the commit retry now treats ErrNonRetryable as terminal.
- Verify in the validator (`rejectCheckpointIfNotVerified`), which already holds the database handle; the committer only forwards the verdict. Report no per-TX status for a held checkpoint: `insertTxStatus` would make the sidecar's re-submission a duplicate forever, so the feedback carries the full `TxRef` instead.
- Forward the status-less batch in the coordinator and release the checkpoint's node explicitly, or the re-delivered checkpoint deadlocks behind its own stale node. Decrement `numTxsInProgress` for the feedback via a shared `txCount` helper, or a feedback-only batch leaks it and `NoPendingTransactionProcessing` never reports idle.
- Gate the sidecar in `processCheckpointFeedback`: HOLD pauses intake for the new `checkpoint-hold-retry-interval` (default 1m) then ends the session retryably, HALT non-retryably. Ending the session re-delivers the checkpoint block, so the relay unwinds no per-block bookkeeping by hand. Add `checkpoint_feedback_state`, `checkpoint_feedback_total{signal}`, `checkpoint_holds_total`, and `block_pull_paused_seconds_total`.
- Key a checkpoint by block number alone: a block holds at most one `_snapshot`. Add `servicepb.CheckpointKey` / `BlockNumFromCheckpointKey` and switch the sidecar's form check off `NewHeightFromBytes`, which required a block+tx key and so rejected every conforming checkpoint.
- Raise the `service/snapshothasher` test context to 7m: its budget is dominated by per-test YugabyteDB setup (~35s of yb-admin each) plus parallel queueing, and the envs added here pushed it past the old 5m.

- The comparison is against the latest `_snapshot` record rather than a lookup by block number: a new snapshot is admitted only once the previous is CHECKPOINTED, so the one awaiting a checkpoint is always the latest.
- `CheckpointFeedback` carries `ref` and `snapshot_block_number`, replacing `tx_id` and `block_number`; field 2 is reserved. The two block numbers were conflated: `snapshot_block_number` is the snapshot attested to, `ref.block_num` locates the checkpoint TX the coordinator looks its node up by.
- Deviates from the PR-15 design: no PROCEED signal (an absent feedback already means the per-TX status is authoritative, so the value is removed and HOLD/HALT renumbered to 1/2), and no held-checkpoint stage in the VC, because a hold is resolved by the sidecar re-delivering the block.

#### Related issues

- resolves #665
- part of #666


cmd/config/samples/sidecar.yaml
```diff
@@ -150,6 +150,12 @@ waiting-txs-limit: 20_000_000
 # by reducing channel contention but increase memory usage.
 # Default: 100
 channel-buffer-size: 100
+# How long the sidecar pauses block intake before re-fetching a checkpoint block that the
+# validator-committer held because the snapshot hash it attests to was still being
+# computed. The hash is published by the snapshot hasher on its own poll interval, so a
+# shorter wait only re-reads the same unfinished state.
```
liran-funaro: minor: Convoluted sentence. The comment is unclear.

docs/performance-tuning.md
```diff
@@ -87,6 +87,12 @@ How often the Sidecar sends the latest committed block number to the Coordinator
 
 Minimal effect on steady-state throughput. Shorter intervals improve recovery speed after failures. Default: 5s.
 
+### `checkpoint-hold-retry-interval`
+
+How long the Sidecar pauses before re-fetching a `_checkpoint` block that the Validator-Committer held because the snapshot hash the checkpoint attests to was still being computed.
```
liran-funaro: minor: Convoluted sentence. Unclear.

docs/sidecar.md
```diff
@@ -391,6 +392,52 @@ the block is split into at most two segments — one segment carrying every othe
...
+| `sidecar_relay_checkpoint_feedback_total{signal}` | Number of feedback messages, grouped by signal. |
+| `sidecar_relay_checkpoint_holds_total` | Number of holds, including repeated holds for the same checkpoint. |
+| `sidecar_relay_block_pull_paused_seconds_total` | Total time spent in completed hold waits, in seconds. |
+
+The state gauge stays at `1` across hold retries. It returns to `0` when a status batch arrives without feedback, not
+when the wait ends. A `HALT` stops processing, so the gauge stays at `2` for the rest of the process.
+
+If the hold counter keeps increasing, check the snapshot hasher's progress and logs. The sidecar logs the snapshot
+block number at WARN for `HOLD`, and the block number and mismatch reason at ERROR for `HALT`. These values are not
+metric labels because they could create too many distinct time series.
```
liran-funaro: Is this necessary given we have the metric doc?

service/coordinator/validator_committer_manager.go
```diff
@@ -309,6 +316,26 @@ func (vc *validatorCommitter) recoverPendingTransactions(inputTxsNode channel.Wr
 	inputTxsNode.Write(pendingTxs)
 }
 
+// checkpointNodeToRelease removes the checkpoint's tracked node for the caller to release.
+// It uses the transaction reference in the feedback because there is no status.
+// If the node is no longer tracked, there is nothing to release.
+func (vc *validatorCommitter) checkpointNodeToRelease(feedback *servicepb.CheckpointFeedback) (
```
liran-funaro: The return options for this method are `nil, false` or `node, true`. So we don't need the boolean since `node` is never `nil`.

service/vc/checkpoint.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/hyperledger/fabric-x-committer/api/servicepb"
+	"github.com/hyperledger/fabric-x-committer/utils/statedb"
+)
+
+// checkpointTx holds the checkpoint write to verify.
+// ref identifies the checkpoint transaction. blockNum identifies its snapshot.
+type checkpointTx struct {
+	txID     TxID
+	ref      *committerpb.TxRef
+	blockNum uint64
```
liran-funaro: minor: We don't need `blockNum`. We have it in `ref.BlockNum`.

service/vc/checkpoint.go
```diff
@@ -0,0 +1,172 @@
...
+			cp.txID, cp.blockNum, state.Status)
+		return checkpointVerdict{feedback: cp.feedback(servicepb.CheckpointFeedback_HOLD, "")}, nil
+	case bytes.Equal(state.Hash, cp.hash):
+		return checkpointVerdict{}, nil // verified: the checkpoint commits with the batch.
+	default:
+		// The same snapshot has different local and checkpoint hashes. Stop for investigation.
+		return d.haltOnCheckpointDivergence(ctx, cp, state, fmt.Sprintf(
+			"local snapshot hash %x for block %d does not match the checkpoint hash %x",
+			state.Hash, cp.blockNum, cp.hash,
+		))
```
liran-funaro: Having this recorded in a string can be hard to parse for the operator in case of an indecent.
Maybe add a field to the state `CheckpointHash` which describes the checkpoint hash in case of a discrepancy.

service/vc/checkpoint.go
```diff
@@ -0,0 +1,172 @@
...
+		return checkpointVerdict{}, fmt.Errorf(
+			"failed to record the checkpoint divergence for block %d: %w", cp.blockNum, err,
+		)
+	}
+	logger.Errorf("Halting on checkpoint TX [%s] for block [%d]: %s", cp.txID, cp.blockNum, reason)
+	return checkpointVerdict{feedback: cp.feedback(servicepb.CheckpointFeedback_HALT, reason)}, nil
+}
+
+// feedback identifies the checkpoint and tells the sidecar whether to wait or stop.
+func (cp *checkpointTx) feedback(
```
liran-funaro: minor: single call tiny methods are discouraged. I thought ponytail discovers these.

service/vc/validator.go
```diff
@@ -126,6 +130,10 @@ func (v *transactionValidator) validate(ctx context.Context, db *database) error
 			return fmt.Errorf("failed to gate snapshot before commit: %w", err)
 		}
 
+		if err := db.rejectCheckpointIfNotVerified(ctx, vTxs); err != nil {
```
liran-funaro: We make a `checkpointTx` here (inside the call), then again in the committer each time we try to commit.
Can we make this once in the prepare phase?

api/servicepb/common.proto
```diff
@@ -29,32 +29,37 @@ message TxWithRef {
...
-        HALT = 3;
+        // The local snapshot hash is not ready. The checkpoint has no transaction status.
+        // The sidecar waits, then submits the same checkpoint again.
+        HOLD = 1;
+        // The local and checkpoint hashes differ. The sidecar stops without retrying.
+        // An operator must investigate.
+        HALT = 2;
     }
 
+    reserved 2; // Was tx_id, now carried by ref.
```
liran-funaro: Do we need to reserve? It wasn't even released yet.

api/servicepb/common.proto
```diff
@@ -29,32 +29,37 @@ message TxWithRef {
...
     Signal signal = 1;
-    // Checkpoint transaction ID used by the coordinator to correlate feedback with the in-flight checkpoint.
-    string tx_id = 2;
-    // Referenced snapshot block number used in coordinator and sidecar diagnostic logs.
-    uint64 block_number = 3;
-    // HALT divergence detail for operator diagnostics, including local and checkpoint hashes.
-    // Required when signal is HALT. Must be logged, not used as a metric label.
+    // The snapshot's block number, used in logs.
+    // The checkpoint transaction's block number is in ref.
+    uint64 snapshot_block_number = 3;
```
liran-funaro: redundant - ref contains this.

PR #810, opened 2026-09-03
[coordinator] Report the dependency graph queue sizes on scrape

### LINKED ISSUE #771: [coordinator] Report queue sizes on scrape instead of sampling them
Part of #730, split out of #733 so each service can be reviewed on its own. Scoped to replacing the
dependency graph's sampler; the queues the coordinator does not observe at all are tracked
separately.

### The monitor

`(*Manager).monitorQueues` (`service/coordinator/dependencygraph/manager.go:78`) wakes every
`queue-monitor-sampling-time` (`100ms`) to publish two `len(ch)` values that are available on demand:
the local and global dependency graph input queues. Converting them retires the sampler, its ticker,
and the config key.

### The two incrementally-maintained gauges

`gdgWaitingTxQueueSize` and `dependentTransactionsQueueSize` (`dependencygraph/metrics.go`) look like
queue sizes but are not channel-backed — they count transactions held in the graph's maps under the
manager's lock, maintained with `promutil.AddToGauge` / `SubFromGauge`. There is no channel to take a
length of on demand, so they keep being maintained incrementally. Their names are the problem, not
their mechanism: rename them, and rename the exported
`coordinator_dependency_graph_dependent_transactions_queue_size`, so neither claims to be a queue.

### Also in scope

Remove the `queue-monitor-sampling-time` config key (`service/coordinator/config.go:23`): the
dependency graph's sampler is its only reader. Deployed YAML that still sets it keeps loading, as
`v.Unmarshal` ignores unknown keys. The key also appears in `cmd/config/samples/coordinator.yaml`,
`cmd/config/templates/coordinator.yaml.tmpl`, `cmd/config/app_config_test.go` and
`loadgen/client_test.go`.

### Constraints

- The queues must exist before the metrics that observe them are registered, so the constructor has
  to build them first and pass them in.

### Metric-surface impact

The two converted gauges become exact at scrape time rather than up to 100ms stale. One rename:
`coordinator_dependency_graph_dependent_transactions_queue_size` becomes
`coordinator_dependency_graph_dependent_transactions`.

liran-funaro:
#### Type of change

#### Description

- Replace the dependency graph manager's ticker-based queue monitor with gauges read on scrape, and
  delete the sampler goroutine.
- Gather the two input queues into a receive-only `managerQueues` struct in `manager.go`, built
  before the metrics that observe them.
- Rename `gdgWaitingTxQueueSize` and `dependentTransactionsQueueSize` to `gdgWaitingTxCount` and
  `dependentTxCount`, since neither is a queue: both count transactions held in the graph's maps and
  stay maintained incrementally.
- Rename the exported `coordinator_dependency_graph_dependent_transactions_queue_size` to
  `coordinator_dependency_graph_dependent_transactions`, for the same reason.
- Reword `coordinator_global_dependency_graph_size`'s help to lead with what it counts.
- Remove the `queue-monitor-sampling-time` config key, whose only reader was the sampler.

Deployed YAML that still sets `queue-monitor-sampling-time` keeps loading, as `v.Unmarshal` ignores
unknown keys.

Dashboard-breaking: `coordinator_dependency_graph_dependent_transactions_queue_size` becomes
`coordinator_dependency_graph_dependent_transactions`. `coordinator_global_dependency_graph_size`
keeps its name, as it never claimed to be a queue.

One of the per-service PRs split out of #736.

#### Related issues

- resolves #771
- related to #730

### DIFF
--- a/service/coordinator/dependencygraph/metrics.go
+++ b/service/coordinator/dependencygraph/metrics.go
@@ -27,10 +27,14 @@ var bucket = []float64{.0001, .001, .002, .003, .004, .005, .01, .03, .05, .1, .
 type perfMetrics struct {
 	provider *monitoring.Provider
 
-	// queue sizes
-	ldgInputTxBatchQueueSize prometheus.Gauge
-	gdgInputTxBatchQueueSize prometheus.Gauge
-	gdgWaitingTxQueueSize    prometheus.Gauge
+	// Input queue sizes.
+	ldgInputTxBatchQueueSize prometheus.GaugeFunc
+	gdgInputTxBatchQueueSize prometheus.GaugeFunc
+
+	// gdgWaitingTxCount is not a queue: it counts transactions held in the graph's maps under the
+	// manager's lock, so there is no channel to take a length of on demand. It stays maintained
+	// incrementally by the code that adds and frees those transactions.
+	gdgWaitingTxCount prometheus.Gauge
 
 	// processed transactions by each manager
 	ldgTxProcessedTotal          prometheus.Counter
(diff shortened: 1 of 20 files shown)


service/coordinator/dependencygraph/metrics.go
```diff
@@ -24,13 +24,26 @@ const (
...
 
+// managerQueues are the queues a manager reports the size of on scrape. They are only ever
+// measured here and never sent to, so they are receive-only, as in the coordinator's struct of the
+// same name. A nil queue reports zero, which SimpleManager relies on: its pre-processing queue
+// holds a different element type and is not reported.
+type managerQueues struct {
+	ldgInput <-chan *TransactionBatch
+	gdgInput <-chan *transactionNodeBatch
+}
+
```
cendhu: In other PRs, we have not placed this queues struct in the metrics.go. It should belong to the manager.go as these structs are core for the flow and not defined for observability alone.

cendhu (approved):
LGTM.

PR #814, opened 2026-09-07
[sidecar] Replace the relay's sync maps with single-owner tracking

### LINKED ISSUE #772: [sidecar] The relay tracks in-flight blocks and TX IDs in sync maps on its per-TX path
The relay keeps two `utils.SyncMap`s and touches them four times per transaction, across its
two hot goroutines:

- `preProcessBlock` -> `mapBlock`: `txIDToHeight.LoadOrStore` for every accepted TX
  (`service/sidecar/mapping.go:322`), storing a `servicepb.Height` that is dereferenced from
  a freshly allocated one.
- `processStatusBatch`: `txIDToHeight.Load` to resolve the TX's height
  (`service/sidecar/relay.go:325`), `blkNumToBlkWithStatus.Load` to find its block
  (`relay.go:350`), and `txIDToHeight.Delete` once the status is applied (`relay.go:360`).

Both maps have 100% key churn: every TX ID and every block number is inserted and then
deleted once, which is the opposite of the read-mostly, stable-key-set workload `sync.Map` is
built for.

Neither map needs to be a `sync.Map`:

- The blocks awaiting statuses are registered in increasing block-number order and retired
  strictly in order, so the tracked numbers always form a contiguous window. That is a FIFO
  ring buffer indexed by `blockNum - base`, not a map — and the status path probes it once
  per transaction, so the difference is on the hottest path there is here.
- The TX ID set only needs to be shared because the status goroutine removes an ID as its
  status arrives. If instead the mapping goroutine evicts a whole block's IDs once it
  observes the block has committed, the set has a single owner and needs no synchronization
  at all. The trade-off is that an ID is then held until its block commits rather than until
  its own status arrives, which only makes the duplicate check more conservative.
- The status routing does not need a global TX ID lookup either. `blockWithStatus.txs` is
  already indexed by TX number and holds every TX's ref, so a status can be matched against
  the tracked block by the (block number, TX number, TX ID) triple of its own ref. That
  covers both reconnect cases the current check handles, and additionally validates the TX
  number, which the current block-number-only comparison does not.

### What this is and isn't

Measured, this is a per-operation and allocation problem, not a lock-contention one. Since
Go 1.24 `sync.Map` is backed by `internal/sync.HashTrieMap`, which is lock-free on read and
locks only a single trie node on insert, so profiling shows the map calls at only 1-2% of
samples with the relay's two writers. The cost is the per-op overhead plus the interface
boxing and `Height` allocation, which is GC pressure the relay pays on every transaction. An
isolated benchmark of the two structures under the relay's exact access pattern: ~320 ns and
3 allocs per TX today, ~123 ns and 0 allocs with the structures above.

### Note: this is not the relay's bottleneck

Worth recording so nobody expects a step change from this. `BenchmarkMapBlockSize` (pure
`mapBlock`, one goroutine) and an end-to-end relay benchmark come out at nearly the same
ns/tx, which means relay throughput is bounded by the single `preProcessBlock` goroutine's
envelope unwrapping, proto unmarshalling and form validation. The status goroutine's work is
hidden behind it. Parallelising `mapBlock` across workers is the larger lever, and is
separate from this issue.

Also noticed while reading: `activeBlocksCount` (`service/sidecar/relay.go:38`) is written on
every block and never read anywhere.

liran-funaro:
#### Type of change

#### Description

- Replace `blkNumToBlkWithStatus` with `inFlightBlocks`: the submitted blocks awaiting
  statuses always form a contiguous block-number window, so they are held in a ring buffer
  and looked up by index instead of by map probe. It also owns the next-block-to-commit
  counter, which used to be a separate relay field, so that invariant lives in one place.
- Replace `txIDToHeight` with `txIDDedup`: a plain map owned solely by `preProcessBlock`,
  which evicts a committed block's TX IDs by watching that counter rather than having the
  status goroutine remove each ID as its status arrives. Nothing else touches it, so it
  needs no synchronization, and it holds only IDs now that no caller reads the height back.
- Route a status by the (block number, TX number, TX ID) triple of its ref, against the
  tracked block's own `txs`, instead of by a global TX ID lookup. This subsumes both
  reconnect cases the previous check covered and additionally validates the TX number,
  which the block-number-only check did not.
- Remove `activeBlocksCount`, which was written on every block and never read.

The dedup window widens as a result: a TX ID is now held until its block commits rather
than until its own status arrives, so a duplicate submitted in that gap is rejected by the
relay, with a status that is not stored in the state DB and so is not notified, rather than
by the VC. That window already existed for a TX whose status had not yet arrived.

Per transaction, the status and mapping paths together drop from four `sync.Map` operations
to one uncontended mutex, and the relay's own allocations fall from 26 to 23 per TX. The
gain grows with the size of the in-flight set, since that is what made the map probes
expensive:

    BenchmarkMapBlockSize/blockSize=100     433.4k -> 448.6k tx/s
    BenchmarkMapBlockSize/blockSize=10000   331.9k -> 393.3k tx/s   (+18%)

To reproduce, run `make bench-sidecar` on this branch and on its merge base, over several
`-count` runs with benchstat rather than from a single result.

Two benchmarks written for this change are not in it. An end-to-end relay benchmark had to
duplicate `run`'s wiring and, with four goroutines in the measurement, came out at around
+/-10% run to run -- wider than the difference it was meant to show. A benchmark of the
window on its own only timed an uncontended mutex and some index arithmetic, against no
baseline, since the structure it replaced is gone. What that one did establish is asserted
instead by `TestInFlightBlocksSteadyStateDoesNotAllocate`: once the ring is large enough, a
block's whole round trip through it allocates nothing, which is the reason it is a ring
rather than a slice resliced at its head.

#### Related issues

- resolves #772

### DIFF
--- a/service/sidecar/mapping.go
+++ b/service/sidecar/mapping.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/hyperledger/fabric-x-committer/api/servicepb"
 	"github.com/hyperledger/fabric-x-committer/service/verifier/policy"
-	"github.com/hyperledger/fabric-x-committer/utils"
 	"github.com/hyperledger/fabric-x-committer/utils/retry"
 	"github.com/hyperledger/fabric-x-committer/utils/serialization"
 )
@@ -47,7 +46,10 @@ type (
 	blockMapper struct {
 		*blockMappingResult
 
-		txIDToHeight *utils.SyncMap[string, servicepb.Height]
+		// dedup is the relay's in-flight TX ID set, and txIDs collects the IDs this block added
+		// to it, which mapBlock hands to dedup as the block's eviction unit.
+		dedup *txIDDedup
+		txIDs []string
 	}
 
 	blockWithStatus struct {
@@ -66,7 +68,10 @@ const (
 	statusIdx             = int(common.BlockMetadataIndex_TRANSACTIONS_FILTER)
 )
 
-func mapBlock(block *common.Block, txIDToHeight *utils.SyncMap[string, servicepb.Height]) (*blockMappingResult, error) {
+// mapBlock maps an orderer block into the batch the relay submits to the coordinator. It records
+// every accepted TX ID in dedup, rejecting a TX whose ID is already in flight, and hands dedup the
+// block's IDs so they are released once the block is committed.
+func mapBlock(block *common.Block, dedup *txIDDedup) (*blockMappingResult, error) {
 	// Prepare block's metadata.
 	if block.Metadata == nil {
 		block.Metadata = &common.BlockMetadata{}
@@ -105,7 +110,8 @@ func mapBlock(block *common.Block, txIDToHeight *utils.SyncMap[string, servicepb
 				blockNumber: blockNumber,
 			},
 		},
-		txIDToHeight: txIDToHeight,
+		dedup: dedup,
+		txIDs: make([]string, 0, txCount),
 	}
 	mapper.withStatus.pendingCount.Store(int32(txCount)) //nolint:gosec // int -> int32
 
@@ -118,6 +124,8 @@ func mapBlock(block *common.Block, txIDToHeight *utils.SyncMap[string, servicepb
 			return nil, err
 		}
 	}
+
+	dedup.trackBlock(blockNumber, mapper.txIDs)
 	return mapper.blockMappingResult, nil
 }
 
@@ -325,11 +333,19 @@ func (m *blockMapper) rejectNonDBStatusTx(
 func (m *blockMapper) addTxIDMapping(ref *committerpb.TxRef) (
 	idAlreadyExists bool, err error,
 ) {
-	_, idAlreadyExists = m.txIDToHeight.LoadOrStore(ref.TxId, *servicepb.NewHeightFromTxRef(ref))
-	if idAlreadyExists {
-		err = m.rejectNonDBStatusTx(ref, committerpb.Status_REJECTED_DUPLICATE_TX_ID, "duplicate tx")
+	if m.dedup.add(ref.TxId) {
+		m.txIDs = append(m.txIDs, ref.TxId)
+		return false, nil
 	}
-	return idAlreadyExists, err
+	return true, m.rejectNonDBStatusTx(ref, committerpb.Status_REJECTED_DUPLICATE_TX_ID, "duplicate tx")
+}
+
+// holds reports whether ref refers to a transaction of this block: the block must carry that TX ID
+// at that position. mapBlock fills txs for every position of the block, including the transactions
+// it rejects itself, so a ref that does not match belongs to a submission the relay no longer
+// tracks — see processStatusBatch.
+func (b *blockWithStatus) holds(ref *committerpb.TxRef) bool {
+	return int(ref.TxNum) < len(b.txs) && b.txs[ref.TxNum].Ref.TxId == ref.TxId
 }
 
 func (b *blockWithStatus) setFinalStatus(txNum uint32, status committerpb.Status) error {
--- a/service/sidecar/mapping.go
+++ b/service/sidecar/mapping.go
@@ -46,10 +46,10 @@ type (
 	blockMapper struct {
 		*blockMappingResult
 
-		// dedup is the relay's in-flight TX ID set, and txIDs collects the IDs this block added
-		// to it, which mapBlock hands to dedup as the block's eviction unit.
-		dedup *txIDDedup
-		txIDs []string
+		// txIDDedup is the relay's in-flight TX ID set, and txIDs collects the IDs this block
+		// added to it, which mapBlock hands to the set as the block's eviction unit.
+		txIDDedup *txIDDedup
+		txIDs     []string
 	}
 
 	blockWithStatus struct {
@@ -110,8 +110,8 @@ func mapBlock(block *common.Block, dedup *txIDDedup) (*blockMappingResult, error
 				blockNumber: blockNumber,
 			},
 		},
-		dedup: dedup,
-		txIDs: make([]string, 0, txCount),
+		txIDDedup: dedup,
+		txIDs:     make([]string, 0, txCount),
 	}
 	mapper.withStatus.pendingCount.Store(int32(txCount)) //nolint:gosec // int -> int32
 
@@ -333,7 +333,7 @@ func (m *blockMapper) rejectNonDBStatusTx(
 func (m *blockMapper) addTxIDMapping(ref *committerpb.TxRef) (
 	idAlreadyExists bool, err error,
 ) {
-	if m.dedup.add(ref.TxId) {
+	if m.txIDDedup.add(ref.TxId) {
 		m.txIDs = append(m.txIDs, ref.TxId)
 		return false, nil
 	}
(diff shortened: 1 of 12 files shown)


service/sidecar/relay_tracking.go
```diff
@@ -0,0 +1,212 @@
...
+	// It is owned exclusively by the relay's preProcessBlock goroutine and therefore needs no
+	// synchronization at all: rather than have the status goroutine remove an ID once its status
+	// arrives, preProcessBlock evicts a whole block's IDs once it observes that the block has been
+	// committed (see evictCommitted). An ID is consequently held for slightly longer than it is
+	// strictly in flight: until the last TX of its block is committed rather than until its own
+	// status arrives. That only widens the window in which a resubmission of the ID is rejected
+	// here, with a status that is not stored in the state DB and so is not notified, rather than by
+	// the VC — a window that already exists for a TX whose status has not yet arrived.
+	txIDDedup struct {
+		ids    map[string]struct{}
```
cendhu: `any` instead of `struct{}` and we can set a `nil` value.

service/sidecar/relay_tracking.go
```diff
@@ -0,0 +1,212 @@
...
+		// relay will emit as committed. It is atomic so that setLastCommittedBlockNumber and the
+		// TX ID eviction can read the relay's progress without taking mu.
+		nextBlockNum atomic.Uint64
+		mu           sync.Mutex
+		// buf holds the tracked blocks, and its length is always a power of two so that a slot is
+		// a mask rather than a division. The block numbered nextBlockNum+i is at slot(head+i).
+		buf   []*blockWithStatus
+		head  int
+		count int
+	}
```
cendhu: this is more of a ring buffer with head denoting the start of the ring and count denotes the number of active items if I am not wrong. Is there a better naming for this struct?

service/sidecar/relay_tracking.go
```diff
@@ -0,0 +1,212 @@
...
+		if blk.blockNumber >= nextBlockNumber {
+			break
+		}
+		for _, txID := range blk.txIDs {
+			delete(d.ids, txID)
+		}
+		committed++
+	}
+	d.blocks = d.blocks[committed:]
+}
```
cendhu: Don;t we need a lock between evict and add?

service/sidecar/mapping.go
```diff
@@ -47,7 +46,10 @@ type (
 	blockMapper struct {
 		*blockMappingResult
 
-		txIDToHeight *utils.SyncMap[string, servicepb.Height]
+		// dedup is the relay's in-flight TX ID set, and txIDs collects the IDs this block added
+		// to it, which mapBlock hands to dedup as the block's eviction unit.
+		dedup *txIDDedup
```
cendhu: txIDDedup -- txID prefix or suffix is needed.

service/sidecar/relay.go
```diff
@@ -141,8 +137,12 @@ func (r *relay) preProcessBlock(
 		// The delivery client guarantees a block with a header and in the correct order.
 		logger.Debugf("Block %d arrived in the relay", block.Header.Number)
 
+		// Releasing the TX IDs of the blocks committed since the previous block was mapped keeps
+		// the dedup set owned by this goroutine alone; see txIDDedup.
+		r.txIDDedup.evictCommitted(r.inFlightBlocks.nextBlockNumber())
```
cendhu: evictCommitted to evictCommittedBelow -- I was wondering why we are evicting the next block number and have to see the evictCommitted implementation to understand the logic.

service/sidecar/relay.go
```diff
@@ -76,14 +73,13 @@ func newRelay(
...
-	r.nextBlockNumberToBeCommitted.Store(config.nextExpectedBlockByCoordinator)
 	r.incomingBlockToBeCommitted = config.incomingBlockToBeCommitted
 	r.outgoingCommittedBlock = config.outgoingCommittedBlock
 	r.outgoingStatusUpdates = config.outgoingStatusUpdates
 	r.outgoingConfigBlocks = config.outgoingConfigBlocks
 	r.outgoingCommittedBlockWithTxs = config.outgoingCommittedBlockWithTxs
-	r.blkNumToBlkWithStatus.Clear()
-	r.txIDToHeight.Clear()
+	r.inFlightBlocks.reset(config.nextExpectedBlockByCoordinator)
+	r.txIDDedup.reset()
```
cendhu: txIDDedup is used only by the preProcessBlock. This is why we do not have a lock between add and evict. We can remove the txIDDedup from the relay object and here and instead create them in the preProcessBlock itself so that the ownership is very clear.

service/sidecar/relay.go
cendhu: Further, inFlightBlocks.nextBlockNumber() reads ambiguous as it returns the nextToCommit or nextBlockNumToCommit. Hence, we can rename nextBlockNumber to inFlightBlocks.nextToCommit or .nextBlockNumberToCommit.

service/sidecar/relay.go
```diff
@@ -387,20 +380,17 @@ func (r *relay) processCommittedBlocksInOrder(
 	defer r.committedBlockMu.Unlock()
 
 	for ctx.Err() == nil {
-		nextBlockNumberToBeCommitted := r.nextBlockNumberToBeCommitted.Load()
-		blkWithStatus, exists := r.blkNumToBlkWithStatus.Load(nextBlockNumberToBeCommitted)
-		if !exists {
-			logger.Debugf("Next block [%d] to be committed is not in progress", nextBlockNumberToBeCommitted)
+		blkWithStatus := r.inFlightBlocks.first()
```
cendhu: is it not inFlightBlocks.nextBlockToCommit()?

service/sidecar/relay.go
```diff
@@ -387,20 +380,17 @@ func (r *relay) processCommittedBlocksInOrder(
...
 		if blkWithStatus.pendingCount.Load() > 0 {
 			return
 		}
-		logger.Debugf("Next block [%d] has been committed", nextBlockNumberToBeCommitted)
+		logger.Debugf("Next block [%d] has been committed", blkWithStatus.blockNumber)
 
-		r.blkNumToBlkWithStatus.Delete(nextBlockNumberToBeCommitted)
-		r.nextBlockNumberToBeCommitted.Add(1)
-		r.activeBlocksCount.Add(-1)
+		r.inFlightBlocks.dropFirst()
```
cendhu: first is internal to inFlightBlocks. Here, we should be explicit. inFlightBlocks.dropLastCommittedBlock() or something similar.

service/sidecar/relay_tracking.go
```diff
@@ -0,0 +1,212 @@
...
+		ids    map[string]struct{}
+		blocks []dedupBlock
+	}
+
+	// dedupBlock holds the TX IDs one block contributed to txIDDedup.ids, so they can be evicted
+	// together. The blocks form a FIFO ordered by block number.
+	dedupBlock struct {
+		blockNumber uint64
+		txIDs       []string
+	}
```
cendhu: we can either move this to mapping.go or relay.go or a new file.

cendhu (commented):
Idea looks good. A few naming and placement related comments.

service/sidecar/relay_tracking.go
liran-funaro: I was advocating using `any` before, but it actually makes more sense to use `struct{}` as it clearly state that there is no value (0 size) rather than `any` which suggest some value might be stored.

cendhu (approved):

PR #821, opened 2026-09-10
[sidecar] Stop allocating per transaction in key validation and TX references

### LINKED ISSUE #787: [sidecar] Key validation and TX references allocate per transaction
Two allocation sites on the mapping path, which is what decides how fast the sidecar allocates — the
collector was 57% of its CPU under load:

- `verifyTxForm` allocates a map and the slice its keys are copied into, per namespace, to check for
  duplicate and empty keys. For the key counts that occur in practice a pairwise comparison needs
  neither.
- `TxRef` and `TxWithRef` are allocated twice per transaction, and can be backed by one slice per block.

Together 61 → 56 allocations per transaction, worth 12% at the default `GOGC`.

liran-funaro:
#### Type of change

- Test update

#### Description

- `checkKeys` now takes the namespace and walks its three read/write sets through `nsKey`,
  instead of the caller copying every key into one slice for it. It finds duplicates by
  comparing pairs, falling back to a set above `maxKeysForPairwiseCheck` keys where the
  quadratic comparison would cost more than the map does.
- `mapBlock` allocates the block's `TxRef` and `TxWithRef` values as two slabs, one entry per
  message, instead of one of each per transaction. Only their addresses are handed on; nothing
  copies an element out, which `go vet`'s copylocks check enforces.
- The mapping benchmarks now generate their transactions with an EDDSA policy
  (`benchTxProfile`) instead of `workload.DefaultProfile`'s `NoScheme`. With no scheme every
  namespace carries a nil endorsement, so mapping rejected all of them with
  `MALFORMED_MISSING_SIGNATURE` before validating a single key or building a `TxWithRef`:
  the benchmarks measured the rejection path, not the path this change touches.
- Adds `TestKeyFormValidation`, covering empty and duplicate keys on both sides of the
  pairwise/set threshold and duplicates that span two of a namespace's read/write sets. Neither
  `MALFORMED_EMPTY_KEY` nor `MALFORMED_DUPLICATE_KEY_IN_READ_WRITE_SET` had any coverage before.

On the fixed workload described above, `BenchmarkMapBlockSize` drops from 24 to 19 allocations
and 1,359 to 1,246 bytes per transaction:

    go test ./service/sidecar/ -run '^$' -bench '^BenchmarkMapBlockSize$' -benchmem

Of that, `checkKeys` accounted for three allocations for the two-key namespace this workload
generates — the map, the slice the keys were copied into, and the string conversion per key —
and now allocates none at any key count. The remaining two are the per-transaction `TxRef` and
`TxWithRef`.

#### Related issues

- resolves #787

### DIFF
--- a/service/sidecar/mapping.go
+++ b/service/sidecar/mapping.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package sidecar
 
 import (
+	"bytes"
 	"fmt"
 	"sync/atomic"
 	"unicode/utf8"
@@ -50,6 +51,13 @@ type (
 		// added to it, which mapBlock hands to the set as the block's eviction unit.
 		txIDDedup *txIDDedup
 		txIDs     []string
+		// refs and txWithRefs back every TxRef and TxWithRef the block needs, one entry per
+		// message, allocated once for the block instead of once per transaction. Mapping is on the
+		// path that decides how fast the sidecar allocates, and these were two of its allocations
+		// per transaction. Nothing may copy an element out of them — they are proto messages, and
+		// only their addresses are ever handed on — which go vet's copylocks check enforces.
+		refs       []committerpb.TxRef
+		txWithRefs []servicepb.TxWithRef
 	}
 
 	blockWithStatus struct {
@@ -66,6 +74,11 @@ type (
 const (
 	statusNotYetValidated = committerpb.Status_STATUS_UNSPECIFIED
 	statusIdx             = int(common.BlockMetadataIndex_TRANSACTIONS_FILTER)
+
+	// maxKeysForPairwiseCheck is the largest number of keys in one namespace that checkKeys
+	// compares pairwise before it switches to building a set. At this count the comparison is a few
+	// hundred short byte-slice compares, well under the cost of the map it replaces.
+	maxKeysForPairwiseCheck = 24
 )
 
 // mapBlock maps an orderer block into the batch the relay submits to the coordinator. It records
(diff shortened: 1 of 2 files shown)


service/sidecar/mapping.go
```diff
@@ -66,6 +74,11 @@ type (
 const (
 	statusNotYetValidated = committerpb.Status_STATUS_UNSPECIFIED
 	statusIdx             = int(common.BlockMetadataIndex_TRANSACTIONS_FILTER)
+
+	// maxKeysForPairwiseCheck is the largest number of keys in one namespace that checkKeys
+	// compares pairwise before it switches to building a set. At this count the comparison is a few
+	// hundred short byte-slice compares, well under the cost of the map it replaces.
+	maxKeysForPairwiseCheck = 24
```
cendhu, line 81: I think this cover the max number of keys used in the evm based tx too.

service/sidecar/mapping.go
```diff
@@ -559,17 +570,53 @@ func checkMetaNamespace(txNs *applicationpb.TxNamespace) committerpb.Status {
...
+	for i := range keyCount {
+		if len(nsKey(ns, i)) == 0 {
 			return committerpb.Status_MALFORMED_EMPTY_KEY
 		}
-		uniqueKeys[string(k)] = nil
 	}
-	if len(uniqueKeys) != len(keys) {
-		return committerpb.Status_MALFORMED_DUPLICATE_KEY_IN_READ_WRITE_SET
+
+	if keyCount > maxKeysForPairwiseCheck {
```
cendhu, line 589: nit. we could reverse the condition as the majority of the txs are expected to have lesser than maxKeys....

cendhu (approved):
LGTM.

PR #808, opened 2026-09-03
[BREAKING] [sidecar] Unify sidecar RPCs into SidecarService, keep Fabric Deliver

cendhu:
#### Type of change

- Improvement (improvement to code, performance, etc)
- Bug fix
- Breaking change

#### Description

- Register a single `committerpb.SidecarService` in `Service.RegisterService`,
  replacing the separate `BlockQueryService` and `Notifier` registrations.
- Embed `*notifier` in `Service` and `UnimplementedSidecarServiceServer` in
  `notifier`, so selector depth promotes the block-query methods, the stream
  methods, and the not-yet-implemented fallbacks onto one object without
  forwarder methods.
- Keep Fabric block delivery on the Fabric `protos.Deliver` service, a separate
  gRPC service, so Fabric clients dial the sidecar unchanged.
- Move clients to `NewSidecarServiceClient`, and consume the renamed upstream
  block stream API (`StreamBlocks`, `StreamBlocksRequest`, `BlockEvent`),
  including `block_hash` and `prev_block_hash`.
- `StreamBlocks` now emits every committed block, including blocks whose
  filters match no transactions.
- `DeleteDBCloneForSnapshot` is reachable and answers `UNIMPLEMENTED` until its
  pipeline lands.
- Guard the delivery session's config blocks behind `configState.configBlock()`: the
  embedded `*ConfigBlockMaterial` is nil until the first config block is processed, so
  `ToQueue` panicked when a session ended before that (also fixes the same nil deref in
  the genesis-block comparison).

#### Additional details (Optional)

Requires the fabric-x-common release containing the unified `SidecarService`
(hyperledger/fabric-x-common#179), pinned here as
`v0.2.9-0.20260910095823-4b946650f960`.

`UnimplementedSidecarServiceServer` is embedded on `notifier` rather than on
`Service` on purpose: embedding it on `Service` would put it at the same
selector depth as the notifier's stream methods, making those selectors
ambiguous so `Service` would silently stop satisfying `SidecarServiceServer`.

#### Related issues

- resolves #659

PR #828, opened 2026-09-15
[deliverorderer] Guard the session config against a nil config block

### LINKED ISSUE #827: [deliverorderer] Nil-pointer panic in ToQueue when it exits before the first config block
### Description

ToQueue() unconditionally dereferences dataStream.ConfigBlockMaterial, which is nil when a fresh committer starts without a verification config.
If delivery exits before receiving a config block—through retry exhaustion or cancellation—building the returned SessionInfo panics.
Existing tests miss this because their shared harness always initializes NextBlockVerificationConfig.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xf18566]

goroutine 133 [running]:
github.com/hyperledger/fabric-x-committer/utils/deliverorderer.ToQueue(...)
	utils/deliverorderer/orderer.go:117
github.com/hyperledger/fabric-x-committer/service/sidecar.(*Service).startDelivery(...)
	service/sidecar/sidecar.go:353
github.com/hyperledger/fabric-x-committer/service/sidecar.(*Service).sendBlocksAndReceiveStatus.func2()
	service/sidecar/sidecar.go:220
```

### Steps to reproduce

_No response_

dean-amar:
#### Type of change

- Bug fix
- Test update

#### Description

- Use a nil-safe accessor for both SessionInfo config blocks to prevent crashes when delivery ends before a config is available, preserve the existing nil-config contract, and add a regression test.

#### Related issues

- resolves #827


liran-funaro:
This was addressed by PR #808

